PR #4, opened 2025-07-01
warscholar Vizier fixes

benj8560:
## About The Pull Request
Removed bag fixes since im to lazy to learn proper conflict resolving. I'll make the bag fix again if it becomes more of a problem...seems rare anyways.

regression (their healing spell) now always applies the standard healing buff ontop of targeted limb auto healing. This is very similar to the way normal miracles function. I don't believe this was intended originally

## Testing Evidence

Booted up a server and confirmed that even with heavy damage and bleeding on a limb the normal healing buff was applied.

## Why It's Good For The Game

Fixes bug. Bug is bad.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/mercenaries/classes/warscholar.dm
+++ b/code/modules/jobs/job_types/roguetown/mercenaries/classes/warscholar.dm
@@ -159,7 +159,6 @@
 			mask = /obj/item/clothing/mask/rogue/lordmask/tarnished
 			belt = /obj/item/storage/belt/rogue/leather
 			beltr = /obj/item/storage/belt/rogue/pouch/coins/poor
-			beltl = /obj/item/storage/belt/rogue/surgery_bag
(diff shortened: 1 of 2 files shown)


Bobinater:
Wasn't this chud supposed to be made into a wretch?

benj8560:
> Wasn't this chud supposed to be made into a wretch?

idk. This is a bug fix PR not a rebalance and rebrand an entire role PR. Tho I'd be pretty disappointed to see that they seem like a rather fun support role.

benj8560:
Tho. It may turn into a bigger PR. I have just noticed a issue on the old github with their healing. AP has done a ton of work on this class and my gut tells me it likely fixes the issue ontop of all the content added. Could you confirm for me if the plan is to rework them into a Wretch? If so I don't really care to put the work in. I dont play Wretch.

PoltavskaPraca:
vizier is now playable, chat is this real?

benj8560:
still says conflict but i gotta assume it does not automatically remove the label?

PR #5, opened 2025-07-01
Update scomm.dm

Bobinater:


KiteAE:
Noticed that this still has "Grand Duke" in the code, we have recently changed Duke to Baron so would need updated.

Also if this is related to a previous PR you made about town watch please put it all in the same PR so its not half implemented and a bit easier to understand overall, thank you.

Bobinater:
> Noticed that this still has "Grand Duke" in the code, we have recently changed Duke to Baron so would need updated.
> 
> Also if this is related to a previous PR you made about town watch please put it all in the same PR so its not half implemented and a bit easier to understand overall, thank you.

Unrelated to my other PR. Also updated.

PR #7, opened 2025-07-01
CHUNKY

Bobinater:
## About The Pull Request

Makes those full plate fat fucks slower.

## Why It's Good For The Game

Those fat fucking bastards should all have -1 speed by default, this PR makes it so.


Fatherless1Xx:
Heavy armor provides only a marginal advantage in most situations. In practice, light armor paired with Dodge Expert often proves more effective overall. To give you an idea:

Studded Leather armor:
blunt 80, slash 80, stab 60, piercing 40
Integrity 300

Brigandine:
blunt 60, slash 100, stab 80, piercing 70
Integrity: 350

This isn't a good change.

PR #12, opened 2025-07-01
Sacred flame does double firestacks against woe'd during the day

Fatherless1Xx:
Excommunication scary


Bobinater:
First off, follow the format.

Ten fire stacks is absolutely fucking overpowered. Absurdly, doubly so for a spell known to be spamable.

PR #16, opened 2025-07-01
Spellbreakers

Fatherless1Xx:
This class is a direct upgrade of the Spellblade adventurer, just like the Brute is a tougher version of the Barbarian.
image
image


Cre77:
more spellpoints for oldboys? 🥺 
get me "an old man in a profession where men usually die young."

ddsaw:
Very cool class, actually but I think you should probably make them more warrior-y.
This isn't a holy warrior with support magic like the text implies,  this is a mage with swords 3 and adv melee stats. It's also missing a lot of flavor skills.

My suggestion:
Compare them to adv spellsword and make sure that they are a step above them, as all classes with an actual job and responsibility usually are.
Give them expert swords, +1int but possibly bump arcyne down to t2, t3 if they are old.
Bump athletics and wrestling to 3.
Give them climbing and sneaking because they are inquisition.
Consider knives 2-3, alchemy and medicine etc. all of those flavor skills.

Also rename the .dm file to spellbreaker.dm.

ddsaw:
They also probably should have a preblessed silver sword instead of rapier.
Probably Patronlock them to Noc or make the Amulet dynamic, then again confessors do get the Xylix mask..

PR #18, opened 2025-07-01
Recall spell

Fatherless1Xx:
reuploaded from blackmoor:
https://github.com/BlackmoorHold/Blackmoor-hold/pull/211


KiteAE:
As i said on Blackmoor, cost needs to be 4, cast time needs to be not instant, invocation needs changed to be less of a reference, needs to be able to interrupt invisibility.

PR #53, opened 2025-07-03
looc shows true name even when masked

Fatherless1Xx:
port of this: https://github.com/BlackmoorHold/Blackmoor-hold/pull/138


Fatherless1Xx:
nevermind, I can see changes have been made on scarlet so a copy paste has fucked things up
will resubmit later

PR #54, opened 2025-07-03
No more heretic keep roles

Mannybrado:
## About The Pull Request
title

## Testing Evidence
works

## Why It's Good For The Game
It keeps being abused. You no longer deserve nice things.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/courtier/butler.dm
+++ b/code/modules/jobs/job_types/roguetown/courtier/butler.dm
@@ -7,6 +7,7 @@
 	spawn_positions = 1
 
 	allowed_races = RACES_ALL_KINDS
+	allowed_patrons = ALL_DIVINE_PATRONS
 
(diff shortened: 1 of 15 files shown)


Fatherless1Xx:
image
Please merge the thing that they said wouldn't happen keeps happening: https://github.com/BlackmoorHold/Blackmoor-hold/pull/168

Fatherless1Xx:
Oh actually I kind of think this is a bit much, it's just the combat roles (garrison) that should be forced to be ten worshippers imho. Heretical nobles makes for good conflict because they cannot just scream PRAISE ZIZO to fragbait.

Inveldy:
This is entirely unreasonable and does not benefit anyone by adding it. What good does this actually do?

KiteAE:
I personally don't agree with this and i do not wish to punish the large majority of players based on the bad actions of maybe 2 or 3.

PR #57, opened 2025-07-03
Brutes now get a choice of weapon like barbs

Fatherless1Xx:
So they're more like a direct upgrade
It's good for the game because we need more people playing inquisition to enforce the lore of the game


Fatherless1Xx:
> Format not followed at all, also this isn't just giving them a choice of weapons, this is a a direct buff to considering you've given them level 3 skills nearly across the board. A better approach would be to give them the skill in weapon via their weapon choice.

Good point, I've made those changes now.

PR #59, opened 2025-07-03
Update arcyne_bolt.dm

Bobinater:
## About The Pull Request

Fixes some leftover slop from AP bad coding.

## Testing Evidence

Launched on local no issue.

## Why It's Good For The Game

So I want people to bear in mind that attacks that do 40 damage or more (or so I'm told) do what I'll refer to as 'micro stuns' which will slow and completely stop someone's action when they occur, the issue that arises is that you can legitimately juggle arcyne bolt fast enough to the point where someone nearly can't move to dodge the next hit at all because of the aforementioned 'micro stun'. Now this is all made worse because some 'genius' on AP decided it'd be funny to make staffs that can reduce the cooldown time on spells.

### DIFF
--- a/code/modules/spells/spell_types/wizard/projectiles_single/arcyne_bolt.dm
+++ b/code/modules/spells/spell_types/wizard/projectiles_single/arcyne_bolt.dm
@@ -28,7 +28,7 @@
 /obj/projectile/energy/rogue3
 	name = "Arcyne Bolt"
 	icon_state = "arcane_barrage"
-	damage = 40
+	damage = 39
 	woundclass = BCLASS_BLUNT
 	nodamage = FALSE
 	npc_damage_mult = 1.5 // Makes it more effective against NPCs.


Maromar15910:
I'm fine with the bolt being nerfed as long as it's not to the point where casting it without a staff becomes a blatantly losing option. It's also good practice to call balance changes a rebalance, buff, or nerf in the PR's title instead of an update, it reduces accusations of bad faith.

BadAtThisGame302:
The "update" is a nerf FYI. Cope PRs yipeee!

Redin01:
Cope PR aside, you do realize that staves only reduce the charge time on spells, right? The mechanic you mentioned doesn't have any effect on arcyne bolt because it's an instant cast. Secondly, arcyne bolts cause slowdowns, but do you know what else causes slowdowns? Other weapons. The same thing can happen if you're shot by arrows or if a speed polearm build runs up to you and repeatedly stabs you, preventing you from escaping. Why not just tweak the hit slowdown mechanic instead?

Like mentioned before, armor blocks arcyne bolt damage and if the distance is closed, it's likely joever for the mage.

Gehenna-Host:
grab intent > aim mouth

PR #62, opened 2025-07-03
Ports Emberwine

BadAtThisGame302:
## About The Pull Request

Mixes and matches a few Emberwine ports to bring Emberwine back to Scarlet Reach!

## Why It's Good For The Game

Good addition to have, among other things.


deadmomo69:
I am aware this is a port, but it'd be lovely if alchemists could craft it with ingredients so they have something to sell to the brothel.

PR #69, opened 2025-07-04
***Ready for TM*** Ten worshippers cannot use lux to revive the excommunicated 

Fatherless1Xx:
Credit to https: https://github.com/BlackmoorHold/Blackmoor-hold/pull/195/files 
I literally just copy pasted their code and rewrote it since we do not have necra's curse here

### DIFF
--- a/code/modules/surgery/surgeries_hearth/revival.dm
+++ b/code/modules/surgery/surgeries_hearth/revival.dm
@@ -51,7 +51,14 @@
 		return FALSE
 	if(istype(user, /mob/living)) {
 		var/mob/living/LU = user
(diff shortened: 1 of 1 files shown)


Fatherless1Xx:
This has been tested and works:
image
image

Wesreidau:
An excellent reason for the Duke to keep a heretical subject in their employ; life insurance.

coffeecoffeecoffee:
my only one complaint with this is that the way it's written, it might not be entirely clear to players who don't keep up with the github "WHY ISN'T LUX WORKINGNGG??!?!?"
and cool RP like "heretic in the retinue" can't happen if people don't understand this pressure or know it exists
i mechanically really quite like this change!! it's kino, i just think maybe it should be clearer in text for players who just play on occasion

PR #73, opened 2025-07-04
By order of the watch patch to fix cabinets and one door

Sniperjolly:
## About The Pull Request

As @Quick mentioned in #bug-reports the cabinets and one door in the watchtower still used the Warden keytype instead of the new Watch one

## Testing Evidence

it's a minor mapping fix you can read the lines of code it changes yourself

## Why It's Good For The Game

bug fix


BadAtThisGame302:
I don't think you intended for this many changes?

PR #77, opened 2025-07-04
Staff of the shepard has cast time reduction and applies firestacks 

Fatherless1Xx:
What if we made the staff good:
image


Gehenna-Host:
Can you remove the firestacks from this so it's more like other staffs? Priests already get a nice miracle for their fire which they can use faster with this now.

PR #79, opened 2025-07-04
Dodge Experts can now be hit by other Speedsters

ddsaw:
## About The Pull Request

Currently, because of how the dodge calculation is coded, two 15spd dodge experts will always have 90% dodge against each other.
Also, as soon as somebody has dodgeexpert, even if they have a very modest spd, they will always reach 90% dodge against you..
This changes the calculation to make the enemies spd matter more for your own dodging - atleast now other spdbuilds can hit you.
## Testing Evidence
Two identical rogues can hit each other 50% of the time. Previously, this was 10%.
iltbpRt

## Why It's Good For The Game
Spd is incredibly dominant right now.
Dodge expert comes from a time when light armor was actually bad.
This atleast makes it less comical.

### DIFF
--- a/code/modules/mob/living/roguetownprocs.dm
+++ b/code/modules/mob/living/roguetownprocs.dm
@@ -495,13 +495,11 @@
 	var/prob2defend = U.defprob
 	if(L.rogfat >= L.maxrogfat)
 		return FALSE
-	if(L)
-		if(H?.check_dodge_skill())
-			prob2defend = prob2defend + (L.STASPD * 15)
-		else
-			prob2defend = prob2defend + (L.STASPD * 10)
-	if(U)
-		prob2defend = prob2defend - (U.STASPD * 10)
+	if(L && U)
+		var/dodgechance = ((L.STASPD-U.STASPD)*10)
+		if(dodgechance>0 && H?.check_dodge_skill()) //This does not penalize you if you are losing
+			dodgechance = dodgechance*1.5 //dodgeexpert multiplier = 1.5
+		prob2defend += dodgechance
 	if(I)
 		if(I.wbalance > 0 && U.STASPD > L.STASPD) //nme weapon is quick, so they get a bonus based on spddiff
 			prob2defend = prob2defend - ( I.wbalance * ((U.STASPD - L.STASPD) * 10) )


Bobinater:
This change in practice was super, super bad. To the point where parrying was a better option for light armor users.

If this isn't unmerged I'll probably never touch dodge again because it'd be suicidal.

Honestly, the way I've looked at fighting dodge users is you better fight smart or use anything but left click spam. Use feints, baits, defends. Back it up with some tossblades, a bow or a crossbow. Con checking as well has been very useful for countering.

Red-Honk:
Use a fast weapon, or grab them if you get the chance, shoved them to have a easier time doing the two, problem solved. (mostly)

ddsaw:
> Use a fast weapon, or grab them if you get the chance, shoved them to have a easier time doing the two, problem solved. (mostly)

1. mostly irrelevant, unless you are even faster yourself(lol?)
2. 90% dodge against your grab

anything86:
Misleading title, misleading PR. To get this out of the way: this is a major nerf to dodge in all scenarios. This reeks of not wanting to do maths and just flinging a change at the wall. If you make the dodge expert bonus apply only to the difference between speeds, it makes it so dodging is borderline useless against anyone using a weapon you are not proficient in, due to how the skill bloc works when dodging. Let me explain with a practical example:

John McKnight is a human knight footman with no statpack. He has 10 speed and is an expert at polearms.
Jeanne McRogue is a human man at arms skirmisher with no statpack. She has 13 speed and no skill at polearms.

Because this change makes the dodge expert bonus apply only to the speed differential, Jeanne McRogue's base dodge change against McKnight would normally be 50% (base) + 15*3 (speed differential), so 85%, right? Wrong!
Due to how the dodge skill bloc works, each level of polearms McKnight has over McRogue applies a -10% penalty to McRogue's dodge chance. So in this case McRogue's dodge chance actually goes down to a pitiful 45%, well below what McKnight would have even if he were blocking with a shield against someone two skill levels above him.

Tldr this is a lazy change that butchers dodge classes and the PR author is trying to pass it off as fixing an edge-case scenario.

anything86:
>
> And without this change, the spdclass has 90% dodge against the same knight. The SPD difference mechanic? Ignored. The weapon skill affecting dodges? Ignored. Any sort of way to bring the chance of a dodgeexpert below 90%? Practically unattainable - you will have to be using some weapon they are not proficient with, hope they did not invest in extra spd(lol), and have as much spd as the knight. I'm sorry to say that there is now an actual active choice for dodgeclasses to use their parry button based on the enemy weapon and their own skill, not that 50% dodge is bad, or that you aren't likely to spick a +spd statpack unlike all of the polearm users.
> 
> I'm gonna change the PR to better explain what this does, however.

I don't think you will bother reading this, but here's an idea: Change the numerical value of the dodge expert factor, rather than making a random change with no basis in balance that nerfs a well-established and balanced upon form of defense into the ground. 13 instead of 15 would be a major nerf already, a flat -30% across the board on 'optimized' dodge builds. This would help bridge the gap in scenarios where dodge **is** supposed to not be 90%, without making the option of dodging dogshit in 90% of combat scenarios.

> I think this PR is much needed. 90% is REALLY really fucking difficult and unfun for anyone to combat.
A kite shield is this but with halved stamina cost per hit nullified and much higher durability (because dodge clipping is now a thing), and you get to dump your speed stat and instead go all in on strength.

Gehenna-Host:
This was TM'd and it was really hard on speed builds to the point of making them unviable. Thanks for the effort though.

PR #84, opened 2025-07-05
Anastasis excommunication block fix

Fatherless1Xx:
Rewrote so it's not ambigious and doesn't cause problems with the compiler
image


Fatherless1Xx:
this PR fixes the problems my code caused by there's some left over by new merges:
image

PR #86, opened 2025-07-05
Confessor Tier1 Miracles (Like Templar)

ddsaw:
## About The Pull Request
Gives T1 Miracles to Confessors, just like the Inquisitor received. Now all Inquisition roles should have either magic or miracles.
## Testing Evidence
An edgelord in his dumb coat setting himself on fire. 
eRPPStz
## Why It's Good For The Game
They are zealous clergy, and their lack of miracles was justified because of them being psydonites. This makes it clear they are not. It also differentiates them from the guards, and lets them express a little bit of individuality through god choice.


roarbark:
it'd be cool if they had unique miracles, but this is thematically fitting for now

PR #87, opened 2025-07-05
Single letter accent removal

EmilyFontaine:
## About The Pull Request
"i" to "ich" and "a" to "un" in the german and french autotranslations would lead to people saying "ICH" and "UN" in all capital letters. There are certainly other ways to deal with this, but it's not really worth it, when it's something that people can very easily write themselves if they like those particular accent quirks enough.

## Testing Evidence
trust me

## Why It's Good For The Game
These errors come up a lot, and since I'm the one who did the "ich" thing to begin with, it bothers me a lot.

### DIFF
--- a/strings/french_replacement.json
+++ b/strings/french_replacement.json
@@ -7,7 +7,6 @@
 		"not": "nut",
 		"I'm": "j'", 
 		"am": "ahm", 
-		"a": "un",
         "and": "eend", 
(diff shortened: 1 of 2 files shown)


EmilyFontaine:
Also let me know if there are any other examples of this in other accents. I glanced at a few of them, but the dwarf accent made me not wanna dive too deep into this.

PR #90, opened 2025-07-05
Fixes Smashing/Biting missing defines

ddsaw:
## About The Pull Request
Port of https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/2779
Fixes missing damagetype defines on smashing and bites. This was resulting in those being way way stronger than intended.
## Testing Evidence
What's important is the crit message of my bite, the sort you'd get from a stab.
wjmQZRq
## Why It's Good For The Game
Otherwise bites will penetrate through basically all heavy armor in the game because of how blunt works, also bites just... shouldn't deal blunt damage?
The same with smash, all blunt has -100ap and it had 0, making it ignore heavy armors in particular.

### DIFF
--- a/code/modules/mob/living/grabbing.dm
+++ b/code/modules/mob/living/grabbing.dm
@@ -493,6 +493,7 @@
 /obj/item/grabbing/bite
 	name = "bite"
 	icon_state = "bite"
+	d_type = "stab"
 	slot_flags = ITEM_SLOT_MOUTH
 	bleed_suppressing = 1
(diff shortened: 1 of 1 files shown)


Red-Honk:
Penfactor -100, basically makes it impossible to deal damage towards someone with any sort of armor if they're hit on a armored part, which I find a tad silly, making it only focus on purely damaging the armor itself, which is kinda bad.

PR #91, opened 2025-07-05
Tetris removal PR

Fatherless1Xx:
Supposedly this is a big cause of lag
Can we TM this to see if it makes any difference?
No problems related to this PR:
image
Only problems another PR addresses


Fatherless1Xx:
returns to the old inventory system:
image

PR #103, opened 2025-07-06
Disables werewolf

Gehenna-Host:
## About The Pull Request

Temporarily disables werewolves because they seem to make every round revolve around them

## Testing Evidence

Screenshot_25

## Why It's Good For The Game

Should break up the monotony of every round being a werewolf round while antags are adjusted

### DIFF
--- a/code/modules/events/antagonist/solo/werewolf.dm
+++ b/code/modules/events/antagonist/solo/werewolf.dm
@@ -14,7 +14,7 @@
 	base_antags = 1
 	maximum_antags = 2
 
-	weight = 10
+	weight = 0
 
 	earliest_start = 0 SECONDS
 


ddsaw:
I love my werewolf massacres.. They are genuinely the only rounds in which I feel something happens. This is just a good example of how overtuned antag code on a community like AP which restricts them by rules goes completely crazy once unfettered by the same rules.
Nerf wolves a bit instead.

ddsaw:
> I never thought they were that overpowered, it's more the fact that the werewolf usually just infects everyone and it becomes WW horde vs Keep every round

they cant infect people if they just get killed before the snowball starts, it's a classic snowball conversion antag. This is what it does when it "wins".

PR #105, opened 2025-07-06
No heretic garrison

Mannybrado:
## About The Pull Request
Prevents heretics from playing garrison roles

## Testing Evidence
it works

## Why It's Good For The Game
It's been a problem and staff want to prevent people from using it as a justification to disregard the basic function of their role.


Bobinater:
Yet another cope PR wanted by like only four people but those four people are the loudest yappers. This will unironically change little to nothing.

pyrzal:
^ he plays heretic garrison

DelineFortune:
But heretics in garrison is your way to be antag against the keep and get away with it
Because average good boy noble ravoxan going to bash your head until purple text for any minor crime 

Heretics usually going to close their eyes for drugs/organs/money/cut

Gremillion:
i think it's better to enforce the rules and just job ban people who don't do it properly. otherwise this seems unnecessary, i think the only thing that makes sense is removing heretics from knight. MAA etc there's little IC sense. i think it's fun:)

pyrzal:
garrison exists to enforce the status quo, they are not going to enforce the status quo as a heretic. allowing them to spawn as a heretic and then banning them for roleplaying as a heretic is entrapment

Bobinater:
>It's been a problem and staff want to prevent people from using it as a justification to disregard the basic function of their role.

Can you provide some examples of this perhaps?

ddsaw:
I think this is only an issue if there is too many of them too regularly (simple flavor of the month issue). Let's hope this is just a fad. If not, I would prefer to solve it by changing player expectations about your ic punishments regarding this (the inquisition should become fraggier).

Mannybrado:
> I would say this limits roleplay and forces characters to be different than how the player intended them to be.

Yeah that's the point! If you want to be an evil guy that helps out the antags, you can do it on aaaaaany other role than the one designed explicitly to combat them. Sorry!

Not like it matters, though. Apparently the rules are going to change or something, so I'm closing this and let's cross our fingers that happens.

PR #107, opened 2025-07-06
Werewolf dynamic stats (werewolf nerf)

ddsaw:
## About The Pull Request

WW stats were hardcoded to be 10 for everything, except str,end and con, which were hardcoded to be 20.
Now WW stats are ontop of your previous stats: +5 to str,end,con -3 to int, and the same you had as nonww for the other ones.
This generally results in weaker WWs, especially in the most important fragger stats, aswell as some variation between them.

## Testing Evidence

pre transformation, intentionally super weak build
oifGEQD
after transformation, weak asf werewolf
sxbUYPV

## Why It's Good For The Game
So we can have ww ingame without them fragging everybody and "making the round revolve around them".

### DIFF
--- a/code/modules/antagonists/roguetown/villain/werewolf/werewolf_transformation.dm
+++ b/code/modules/antagonists/roguetown/villain/werewolf/werewolf_transformation.dm
@@ -118,10 +118,10 @@
 	W.emote("rage")
 
 
-	W.mind.adjust_skillrank(/datum/skill/combat/wrestling, 5, TRUE)
-	W.mind.adjust_skillrank(/datum/skill/combat/unarmed, 5, TRUE)
-	W.mind.adjust_skillrank(/datum/skill/misc/climbing, 6, TRUE)
- 	W.mind.adjust_skillrank(/datum/skill/misc/swimming, 5, TRUE)
+	W.adjust_skillrank(/datum/skill/combat/wrestling, 5, TRUE)
+	W.adjust_skillrank(/datum/skill/combat/unarmed, 5, TRUE)
+	W.adjust_skillrank(/datum/skill/misc/climbing, 6, TRUE)
+	W.adjust_skillrank(/datum/skill/misc/swimming, 5, TRUE)
 	
 	W.STASTR = src.STASTR +5
 	W.STACON = src.STACON +5


roarbark:
since this is more how ratwood handles it, from personal experience playing WW there, the stats they got there that i liked the most were PER, SPD and END, that was for being able to do hit and runs, being able to target weakspots and being able to have a stay in stamina attrition

this means that ranger characters and speed builds might end up making the best werevolves, with this change.

outside of this curiosity, i think this is overall a good change, balance wise

roarbark:
grabs are resisted by end, con and str, which you will trump most people in, as well as trumping them with wrestling skill. i dont think this is going to be an issue.

ddsaw:
> This change makes no sense, because with that "dynamic stats" they are turning from mediocre threat, who you need to gather 2-3 adventurers for - to a wuss with low armor, huge hitbox, and questionable fighting ability. They have a good ability to wrestle, because duh - 20 str, level 5 wrestling. But even then they didn't have a complete invulnerability to grabs. Now, when they are FIVE STRENGTH AND CONSTITUTION POINTS WEAKER, i can see them simply getting tackled in a lot of cases. Yes, even if you have 15 speed, you are still very much vulnerable to someone landing a lucky grab. Yes, werewolves can use speed. No, it's not making them good after such a drastic strength nerf.

they are not five strength and con points weaker. They are your original strength +5. If you were a strclass, and had 13 str(ordinary), your ww form has 18 str.

Mannybrado:
> grabs are resisted by end, con and str, which you will trump most people in, as well as trumping them with wrestling skill. i dont think this is going to be an issue.

to add onto this, you also have infinite stamina, so really this won't change a whole lot in that regard

roarbark:
i seriously doubt this, as i said, give them some SPD and maybe a little PER and testmerge it and see what happens.

Shap04ka:
if they are going to get speed, and perception - i will agree that it is at least a sensible change. I'm not talking about whether this change is justified or not, since it's subjective.

ddsaw:
> > they are not five strength and con points weaker. They are your original strength +5. If you were a strclass, and had 13 str(ordinary), your ww form has 18 str.
> 
> Not everyone who gets turned into a werewolf is a combat role. Should we just forget about non-combat towners? And yes, of course , if I was a barbarian with muscular stat pack I would be 20 STR wolf like before. But that comes at a cost of having way less speed, because your nerf does not give them any speed
> 
> > > grabs are resisted by end, con and str, which you will trump most people in, as well as trumping them with wrestling skill. i dont think this is going to be an issue.
> 
> > to add onto this, you also have infinite stamina, so really this won't change a whole lot in that regard
> 
> Yeah well that is very nice and all but wolves don't get neither a good dodge, or a good parry option. Their claws have only 2 defence, and they also don't get good speed unless they were a dodge class before. And yes, if they were a good dodge class before they have even less strength, which makes them even more prone to grab-fu adepts.

Yes. This is a nerf. This is supposed to be a nerf. Werewolfs are hard-disabled because they snowballed too hard (due to being too strong).
I really doubt you are wrestling a master-wrestling ww, and I don't really care if a 15 str monk somehow manages to do it(even he still has at best a 40% chance, and this is assuming you were a weak asf 10str ww and he maxxed str as hard as he could).

Foxentail:
This is bad. I tried a round with it and you're just so weak that any two garrison members can probably take you out lol. I was lunacy embracer too and I had like very sub-par stats.

Werewolves havent been played much recently and most of the time they are taken out fairly easily. This will make it werewolves all but extinct I feel. Especially since werewolf genitals havent been added yet so you can't even goon as them. Smh.

PR #116, opened 2025-07-07
Any age character can be Duke/Duchess

coffeecoffeecoffee:
## About The Pull Request
Title. 

## Testing Evidence
Yes, I even tested this one-line change. If this has knock-on effects though, I wouldn't know.

## Why It's Good For The Game
We're seeing the Duke role go completely unfilled for whole rounds particularly on lowpop fairly often. With our current demographics, it seems somewhat unfeasable, and it's limiting for no good reason; the idea was that the roleplay would be better with older dukes, but I don't think we've seen any meaningful difference in the amount of slop just by restricting it to older characters. It makes enough sense that the heirs may be stepchildren, nephews and nieces lore-wise, which is what people are sometimes roleplaying due to racial differences already anyways.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/nobility/lady.dm
+++ b/code/modules/jobs/job_types/roguetown/nobility/lady.dm
@@ -8,7 +8,6 @@
 	allowed_sexes = list(MALE, FEMALE)
 	allowed_races = RACES_TOLERATED_UP
 	allowed_patrons = NON_PSYDON_PATRONS
-	allowed_ages = list(AGE_MIDDLEAGED, AGE_OLD)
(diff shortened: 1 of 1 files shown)


Gehenna-Host:
It's like that because there aren't any children in the game and the duke has to be older than their adult children because the duke is their biological parent

coffeecoffeecoffee:
> It's like that because there aren't any children in the game and the duke has to be older than their adult children because the duke is their biological parent

People already are capable of regularly making reasonable assumptions about why an elven pair of parents might have a pair of wildkin kids - like that they're stepchildren, nieces, nephews, adopted or just inheritors. As well, "adult" can reasonably mean anything from 18 to maybe 40, and a 36 year old can feasably have an 18 year old child.
I think people are capable of seeing an "adult" duchess and an "adult" prince and it still being believable that the duchess has seniority here, is why I thought this might be chill.
If we had "children" or "young" characters, it'd be a very very different story.

ayasanasa:
If you're doing this you should do same for the consort.

PR #117, opened 2025-07-07
Removal of area specific traits (Conflicts Fixed)

BadAtThisGame302:
## About The Pull Request

Removes the area specific buffs of Woodsman/Guardsman/Dungeonmaster and Tavern Fighter.

## Why It's Good For The Game

As discussed in coder chat, most agree these buffs are hugely unnecessary and OP.


roarbark:
guards and ESPECIALLY wardens are balanced around having these buffs, should they not be buffed after this? this nerf makes them as weak as adventurers (7 weighted statpoints) and weaker than mercenaries (10 ish weighted statpoints). antagonists, especially wretches and bandits profit from up to 15 statpoints, with the latter often outnumbering or going even against the keep during skirmishes. this change seems heavy-handed to me and is going to rock the boat extremely balancewise, if dropped into the game with no further changes...

or in short, you are removing 6 to 7 statpoints from guard roles that struggle against organized bandits already. and to be honest i liked the buffs. they're fun.

anything86:
God yes please. The +1 speed and +2/+3 perception are particularly egregious and made it so dodge guards were straight up as fast as classes that were paying -1 strength just to have +3 speed. Not to mention the 17/18 perception archer guards/wardens.

chemistrymain3:
Keep tavern fighter because it’s funny.

ddsaw:
Agreed. Do this and buff guard roles with +2 weighted statpoints. I think it's silly (tavern superbonus..) and a bit too convoluted. It also leads to people making false assumptions about the power of guards.
The entire retinue chasing people into the forest is not a thing I have ever seen in all of my scarlet reach rounds.

ddsaw:
> People already go mercenary to play MAA+, and roundstart sprint to the keep. They're better equipped, have stuff like spells and generally aren't restricted like MAA. Not really something this PR could counter, but figure I should toss that out there. Grudgebearers and Naledi make MAAs look like a joke.
> 
> The buffs, while not great for balance, were great incentive to force players to remain in appropriate areas for their roles. They're also balanced around such. Without this, they're going against the Mathlete and his goon squad who are always at full strength.
> 
> Compare Skirmisher MAA, the strongest MAA subclass, to Knave. Knave gets TEN stat points. Skirmisher gets FIVE. Half, because they're entirely 'balanced'(as poor as it is) around remaining in the town itself. Without this PR, they'd still be receiving their ten by comparison.
> 
> Mind, brigands can also, at least until very recently, juice their stats by just roundstart buying the stat boosting kit. And even without that, they still have the hoardmaster.

Bandits and wretches should simply also be brought in line and have their stat inflation undone. There is no reason for them to have 10-20 statpoints and master skills, it makes little sense.
Mercenaries should probably also feel a bit of the nerf - they really are just adventurer+.
MAA are extremely numerous and shouldn't be this strong whatsoever, knights should be a bit stronger(why are they basically maa stat and skillwise..?)

PR #118, opened 2025-07-07
improves tiefling accent

Fatherless1Xx:
the current spanish accent is lacking
image


Fatherless1Xx:
more improvements:
image

Mannybrado:
looks cool but you should change friend to a full work replacement so that "friendly" doesn't turn into "amigoly"

also the s -> ss thing sucks and doesn't fit

DorfGuy1:
Hey, can you make it to where the snow elf accent is back to like the softer russian accent instead of that horrible phaux cyrillic shit?

PR #124, opened 2025-07-07
Lessens the racial locks of Consort

BadAtThisGame302:
## About The Pull Request

Lets every race but Tieflings, Half-Orcs, Dark Elves and Goblins to be a Consort.

## Why It's Good For The Game

Considering the implications of political marriage in the Consort description text over a form of love, one could rightfully assume that political marriage can happen between a variety of races, all due to the political/military/economical power of the person that provides the Consort, be it a Human, a Lupian, a Tabaxi or an Elf. All races would have some form of nobility in their respective cities/nations/areas of the world and that doesn't stop them from marrying outside their race to expand their influence or power.


KiteAE:
No I do not wish to see this.

PR #150, opened 2025-07-09
Softcap removal

Fatherless1Xx:
Not a fan of content removal PRs. Never have been, never will be. Frankly, people who make a habit of them usually aren't doing the game any favors. That said, I think we should get rid of softcaps.

This softcap ends up gutting a lot of buff spells. This also hits power roles the hardest - like templars who have insane statlines by design. If balance is the goal, there are better ways to do it than strangling high-end potential.


Mannybrado:
👍 

Another azurism. They powercrept the fuck out of the statspreads, so instead of toning it back they just nerf every single stat so giving people more of a stat is meaningless. Genius. Really fucking genius.

chemistrymain3:
Idiotic. Keep the soft cap.

FreeStylaLT:
image

Ophanimdev:
Going to make one of these for int spell cooldown too.

PR #153, opened 2025-07-09
Lets half-elves have humen ears as an option, lets wild-kin and half-kin have tusks

coffeecoffeecoffee:
## About The Pull Request
Simply allows half-elves to better pass as humen if they so wish by giving them the option to unselect ear sprites.
Also does it for tusks on wildkin and halfkin lol whoops almost forgot that part

## Testing Evidence
F2yo3lifGZ
FzXCzKnZ3c

## Why It's Good For The Game
I was making a half-elf, realized they only get the elf ears and thought the size of them was kind of obnoxious. I thought it'd be a nice way to differentiate them very slightly and allow people to portray different aspects and fractions of elvish blood in their characters while still taking the race, instead of going "ooo, my mother was a half-elf!" '>Corbus is a towering Humen male...'
If preferred I could just sprite a smaller set of elf ears instead so you have to play with SOME kind of visibly point knife-ears instead.
Somebody asked for the other thing and it seemed completely reasonable.

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/furry/anthromorph.dm
+++ b/code/modules/mob/living/carbon/human/species_types/furry/anthromorph.dm
@@ -80,6 +80,7 @@
 		/datum/customizer/organ/penis/anthro,
 		/datum/customizer/organ/breasts/animal,
 		/datum/customizer/organ/vagina/anthro,
+		/datum/customizer/organ/horns/tusks,
 		)
 	body_marking_sets = list(
 		/datum/body_marking_set/none,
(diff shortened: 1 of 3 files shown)


coffeecoffeecoffee:
> > If preferred I could just sprite a smaller set of elf ears instead so you have to play with SOME kind of visibly point knife-ears instead.
> 
> Feel free, more ears is good. I play a half elf with ear tips that have been sliced off so would be great to have more options for character customization purposes.

I was avoiding doing it because spriting is effort
also somebody mentioned wildkin and halfkin can't have orc tusks so lol lmao here we go

PR #162, opened 2025-07-09
squire race expansion ?

CandleJaxx:
## About The Pull Request

basically makes squires races like azure

## Testing Evidence

removed two lines and it let me select nonrespected races as squire in game :) see code

## Why It's Good For The Game

character freedom & more squires who are weird. knight hasn't been touched so you can just say that non-respected race squires just aren't allowed to become knights

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/squire.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/squire.dm
@@ -5,8 +5,6 @@
 	faction = "Station"
 	total_positions = 4
 	spawn_positions = 4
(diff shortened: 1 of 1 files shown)


coffeecoffeecoffee:
I would like to see some kind of lore reflecting this or something because yeah, my first thought was "I'd like to see this but the implication is they could become knights and I don't want that"
Would be nice if we could come up with a description change or something to reflect this possibility of just being a knight's ~~slave for life~~

coffeecoffeecoffee:
there it is
Peavk

PR #164, opened 2025-07-10
Allows higher saturation for color customizer

Rin053:
## About The Pull Request

There was a PR that lowered max saturation to 80%. Which is too low. I set it back to 100% and left everything in-tact with the rest of the PR. Though under the previous the color below is essentially considered sparkledog with a saturation of 98.3 yet just looks like actual light blonde hair in game.

image

image

## Testing Evidence

image

## Why It's Good For The Game

Removes arbitrary limits to patterns of 16,777,216 color combinations.

### DIFF
--- a/code/_globalvars/customization/organ_customization.dm
+++ b/code/_globalvars/customization/organ_customization.dm
@@ -42,7 +42,7 @@ GLOBAL_LIST_INIT(customizers, build_customizers())
 		.[type] = new type()
 	return .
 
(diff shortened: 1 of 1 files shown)


coffeecoffeecoffee:
> crazy to me that we allow people with horse cocks and dwarf sized tits but draw the line at bright hair

one is obnoxious and annoying (and much more public-facing) and the others are fun (and only happen with clothes off)
that is literally it

Ophanimdev:
Do this. I have an idea how to fix certain problems with bright hair.

PR #168, opened 2025-07-10
Volf Form Riding/Adjustments, Adds Saiga Form Faction

ophaq:
## About The Pull Request
Adds the following
- Saiga form now has 'saiga' faction.
- Volf form's STR 7 ->10.
- Ridable volf form.

## Testing Evidence
volfpr

## Why It's Good For The Game
- Saiga form should have had saiga faction originally. It was forgotten and often you'd get mauled by saigas while exploring Mt. Decap in it.
- Volf form currently can't fireman carry people if they tried, and the threshold to do so is about 10 STR... so I bumped the STR up.
- I gave volfs the ability to be ridden because when the druid PR first came out we had this feature and I thought it was cool. Although it looks a tiny bit jank. Volf riders (such as goblins on their back) and such are a big thing in fantasy and I miss seeing that.

I know some people might feel iffy on having ridable volfs when we have saiga form that does that too but hey, we also have hog riders.

Let me know if there is something you don't like and I'll adjust things.

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/wildshape/volf.dm
+++ b/code/modules/mob/living/carbon/human/species_types/wildshape/volf.dm
@@ -7,7 +7,7 @@
 	// Someone else balance this, I am here for code, not numbers
 
 //BUCKLING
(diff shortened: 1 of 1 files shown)


ophaq:
> There should be a define for riding offsets. Here's for example the saiga one. image

I looked into it and apparently there is none that exists for volfs so... I think I would end up having to make a new define for volfs? Normally you can't ride them. Not sure how to do this, otherwise.

PR #170, opened 2025-07-10
deazurefies Ravox

Ophanimdev:
I worship the idea of living a life in battle and dying gloriously but I would NEVER EVER cleave someone's limbs from bone or sing a saga about a mighty warrior.

### DIFF
--- a/code/datums/gods/patrons/divine_pantheon.dm
+++ b/code/datums/gods/patrons/divine_pantheon.dm
@@ -92,9 +92,9 @@
 
 /datum/patron/divine/ravox
 	name = "Ravox"
-	domain = "God of Justice, Glory, Battle"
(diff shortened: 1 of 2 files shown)


roarbark:
this entirely changes ravox - god of justice hasn't been an AP thing, wasn't he like that everywhere? hence his usage in lawmaking and bounty stuff. 

also the text is written a little awkwardly and should get a lookover IMO

Red-Honk:
God of War, justice and combat, I believe, but if we take justice out of it, we can actually have a more diverse set of characters.

zeratino:
this is badass

im sick of people decapitating every bad guy they meet, hiding the head and then calling it 'ravox's justice'

coffeecoffeecoffee:
> this is badass
> 
> im sick of people decapitating every bad guy they meet, hiding the head and then calling it 'ravox's justice'

and we are to do this by... removing morality or justice from ravox and making him even more like graggar?

PR #171, opened 2025-07-10
Delitzes Critical Hits

Ophanimdev:
This PR does very little save for removing certain hardcoded crit-nullifiers in two files.
In essence what this means is that you no longer have to fully damage a limb before you can land a critical wound on it.
Nor will there be any babyrails preventing any possibly fatal crits from occuring if you're standing up.

This should help with the overall high TTK and elucidate people as to the true lethality of their weapons (and the general sorry state of force value inflation on most.)

### DIFF
--- a/code/modules/surgery/bodyparts/bodypart_wounds.dm
+++ b/code/modules/surgery/bodyparts/bodypart_wounds.dm
@@ -124,8 +124,6 @@
 		var/mob/living/carbon/human/human_owner = owner
 		if(human_owner.checkcritarmor(zone_precise, bclass))
 			return FALSE
-		if(owner.mind && (get_damage() <= (max_damage * 0.9))) //No crits unless the damage is maxed out.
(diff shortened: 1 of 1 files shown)


chemistrymain3:
Abysmally fucking bad PR. Do not do this. It was changed for a reason.
You do not want standing decapitations back, you do not want standing instakills back, and the TTK is fine. This solves a nonexistent issue to a nonexistent problem.

The entire POINT of the existing crit system is to slowly move away from everyone being super armored out of necessity and requiring absurd defend chances etc to survive.

pyrzal:
It seems like this pushes us even closer to a "everyone needs armor, or else combat is death in two clicks" meta, which is the opposite of how we want things to go. Realistic or not, cementing the armor meta even further is going to turn fights into even more of a slog, which is bad for gameplay reasons. I don't think we want this change.

PR #173, opened 2025-07-10
Intelligence Spellcaps Unlocked

Ophanimdev:
Previously, intelligences' benefits when spellcasting were capped out at 15 int and its benefit towards spell cooldowns was marginal at best.

This raises the cap to 99 and increases the spell cooldown reduction to 20%. 

That's all it does, all I did was change a few defines.


roarbark:
mage buff. do they really need it?

DorfGuy1:
iF mages just use fireball and ranged stun, it's pretty crazy. 
And boring

PR #178, opened 2025-07-11
Humble Druid PR: brother may I have more slots

Murderfoxen:
**## About The Pull Request**

- Increases the slots from 2 to 5.
- 
## Testing Evidence

It is two values numerically, it shall work.

**## Why It's Good For The Game**

Druids are not fully treated as members of the church, though they are dendorites. Having a lock on them that is similiar to church acolytes helps them to stand out and roleplay around their focus.

Druids are also very isolationist and solitary, meaning they tend to spread out very far across the map or otherwise huddle together.

I believe that 5 slots will serve scarlet reach's current population boost as a good place for a while.


KiteAE:
4 and i will accept it.

PR #195, opened 2025-07-12
Sawbones gets Astrata's healing miracles

Fatherless1Xx:
From an IC perspective this makes sense since Matthios loves stealing Astrata's shit From an OOC perspective it'll make it easier for bandits to capture people alive since they often overshoot and kill people on accident Think of the fragging
image


benj8560:
I don't imagine they need revive personally. Little bit much if bandits are playing in such a way they are getting killed often enough to warrant a revive spell they are just playing the role wrong.

PR #201, opened 2025-07-13
Deadite trait

Fatherless1Xx:
-Unique examine text
-Stops deadites from being able to jump
image


CloverIsLucky:
Did you just die by a deadite player using jump intent to salt-pr about it?

The kick intent in my K buttons: 
a7c7de671c93bd0d53209e2c26e5a6586078d457_hq

PR #203, opened 2025-07-13
Zizo poison

Fatherless1Xx:
Turns the drinker into a deadite
image
image


ddsaw:
So how many u does this need to essentially instakill somebody?

PR #208, opened 2025-07-14
WIP - LOOC always shows true name and charges you triumphs

Fatherless1Xx:
.

### DIFF
--- a/modular_hearthstone/code/interface/LOOC.dm
+++ b/modular_hearthstone/code/interface/LOOC.dm
@@ -57,12 +57,22 @@
 		to_chat(src, span_danger("You have OOC muted."))
 		return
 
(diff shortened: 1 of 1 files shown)


Smuggytheone:
W PR as always pookie

Fatherless1Xx:
I've heard your concerns about new players and have addressed them here:
https://github.com/Scarlet-Reach/Scarlet-Reach/pull/220

BadAtThisGame302:
The usage of LOOC will finally be cut down, I think this is a better solution than removing LOOC.

coffeecoffeecoffee:
> The usage of LOOC will finally be cut down, I think this is a better solution than removing LOOC.
how about not changing or removing LOOC because it's not a real problem

PR #212, opened 2025-07-14
Someone wearing a cursed mask dying triggers a bad omen goblin invasion

Fatherless1Xx:
Either accept the half bounty or actually give them the chance to survive the sentncing

image

### DIFF
--- a/code/modules/clothing/rogueclothes/mask.dm
+++ b/code/modules/clothing/rogueclothes/mask.dm
@@ -237,6 +237,16 @@
 		say("YOUR PENANCE WILL BE COMPLETE IN [timer_minutes] MINUTES.")
 	return
 
+/obj/item/clothing/mask/rogue/facemask/prisoner/on_mob_death(mob/living/L, gibbed)
+	. = ..()
+	if(istype(L, /mob/living/carbon/human))
+		// Trigger Goblin Portals (Gob Invasion) event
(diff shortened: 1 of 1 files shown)


Fatherless1Xx:
> Won't this just be used by salty bounties getting caught killing themselves and punishing everyone else on the round for it? Why such a global punishment?? Why not, I don't know, a fine (Bounty if they can't pay the fine) to the person that claimed the bounty??? (This would _also_ be a bad idea, due to what I already mentioned, salty players just killing themselves to punish their captors. But I feel is better than just goblin raiding the place.) Or, another option, people locking up masked people, to prevent them from dying or killing themselves, also preventing them from interacting with much of the round for the time being.
> 
> Also why would ravox use _goblins_?
> 
> And, before you even mention "Admins should just ban the ones that kill themselves, duh", dumping more things for admins to moderate doesn't help, _**specially**_ when this adds nothing valuable to the game.

goblin invasions are fun it's not really a punishment on an OOC level
maybe goblins are just taking advantage of Ravox's outrage to attack while he's unlikely to intervene

Fatherless1Xx:
> > goblin invasions are fun it's not really a punishment on an OOC level
> 
> Then... what's the purpose of this PR if not punish people from killing claimed bounties?

IC and OOC are two different things
A punishment for a character doesn't have to be unfun for the player

PR #224, opened 2025-07-15
Adds zizo worshippers to the undead faction

Fatherless1Xx:
This means NPC undead will not aggro on them
image


benj8560:
> won't this mean they can just waltz through dungeons and slurp up all the loot Liches and necromancers and the like should absolutely be undead faction if they aren't already

exactly this and this is 100% how the powergamers will use it. This would need to be a more....elaborate fix assuming its for the new spell. unfortunately.

coffeecoffeecoffee:
> > won't this mean they can just waltz through dungeons and slurp up all the loot Liches and necromancers and the like should absolutely be undead faction if they aren't already
> 
> Boohoo you can already do this as a Necra missionary.

"because there's a problem, we should make it worse!"

KiteAE:
I don't think this is a good idea for multiple reasons. If it didn't include hostile dungeon mobs I would accept but in its current state this is a bit too much.

PR #232, opened 2025-07-16
Sire, the peasants have learned how to fold clothes!

CloverIsLucky:
## About The Pull Request

"Adds" a bit of common sense to the inventory management:

* Basically most clothes(non armors) get their inventory grid usage reduced, using instead 1x2 grids, loincloth uses 1 grid. 
* All pants less plated pants use 1x2, 
* Most light armors and some medium armors like hauberks and half-plate corslet(the bikini half-plate) use 2x2 grids. 
* Chainmail shirts, despite of being medium armors, they became smaller sized than their hauberk sister to 1x2. 
* Other medium armors and heavy ones are left untouched.

## Testing Evidence

I'm leaving it on draft so I haven't tested it yet and I think I'm missing more armor pieces to edit the space they use. But feel free to take a look and leave a comment about it.

Edit WED,16JUL 2025: I tested it and it works.
imagen

Loincloth uses 1 grid as intended because is made of 1 cloth.

imagen

imagen

Chainmaille, hauberk, chain chauses, and arming jacket: 
imagen

## Why It's Good For The Game

Since I've experienced the inventory grid system, it always was annoying to see clothes and armors like the non-armor ones for example, being so cumbersome, and the chainmail shirts using as much room as a fullplate as if those never get folded to be more compact. I know this was intended from the beginning for game balance, but what do you think?

v4-460px-Fold-Clothes-Step-1

### DIFF
--- a/code/modules/clothing/rogueclothes/armor.dm
+++ b/code/modules/clothing/rogueclothes/armor.dm
@@ -717,7 +717,7 @@
 	anvilrepair = /datum/skill/craft/armorsmithing
 	smeltresult = /obj/item/ingot/steel
 	armor_class = ARMOR_CLASS_MEDIUM
-	grid_height = 64
+	grid_height = 32 // Chainmail shirts aren't so rigid, are they?
 	grid_width = 64
 
 /obj/item/clothing/suit/roguetown/armor/chainmail/Initialize(mapload)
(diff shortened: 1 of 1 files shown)


Kepteyn:
Do they have folded sprites?

CloverIsLucky:
> Do they have folded sprites?



> Foldable shirts is cool, I like it anyhow.

Unfortunally no sprites, I just modified their inventory grid size, but I like the idea. Though even an unfolded shirt shouldn't leave you no room to store more things in your satchel or backpack.



> there isn't any icon file associated, so I assume not. I still think we should make the item size for non-armor clothes 1x2 since DEAR GOD is it painful to try and carry around a shirt, tights, and coat as a knight.

Should I edit the chainmail shirts to use the same grid slots are the clothes too? 🤔

PR #237, opened 2025-07-16
Blacksmithing now is the only smithing

Heheniy:
## About The Pull Request
Removes armorsmithing and weaponsmithing entirely, blacksmithing is now the only smithing skill
Reduces blacksmithing exp gain

## Testing Evidence
image

## Why It's Good For The Game
Separating smithing into 3 different branches makes little to no sense, expecially when all 'smithing' roles already have all 3 skill at same level


KiteAE:
This would cause too many headaches in the long run so sadly going to have to decline this.

PR #246, opened 2025-07-16
Aasimar is made into a tolerated race once again.

Wildwestmars:
This was undone by accident, from what I understand.
image
- Redin, person making the lore as I write this.


Kepteyn:
Drakian should totally be tolerated too. Aasimar definitely deserve tolerance though.

PR #254, opened 2025-07-17
Fixes and improves marriage code

Fatherless1Xx:
The biggest difference is it now actually works
However, this no longer randomly assigns you a surname. The priest assigns one upon completing the ceremony:
It also blocks excommunicated from being married
image
image


Fatherless1Xx:
image

PR #257, opened 2025-07-18
Minor Racial Job Restriction Rework Port

BadAtThisGame302:
## About The Pull Request

Ports this PR, https://github.com/BlackmoorHold/Blackmoor-hold/pull/130, with my own changes and excluding some from said PR.

Effective Changes:

**Kobolds and Verminfolk fall under the despised category.**

Druid: Not playable by Constructs. (should be self-explanatory why)

## Why It's Good For The Game

I think it's about time this PR got ported and Kobolds/Verminfolk got put in the same category as goblins, it never made any sense they would be above them, or Dark Elves or Tiefs etc. They should all be bog 0 due to their nature and the racial description text emboldening such.


BadAtThisGame302:
The person that was going to make the PR originally made it, so.

PR #259, opened 2025-07-18
Bugfixes Ignition and changes Sacred Flame

ShadowDeath6:
## About The Pull Request
Fixes ignition, it now works

Sacred Flame now ignites people instead of electrocuting them, non-undead will take less damage and get fewer burn stacks
also changes the SFX for sacred flame to be bless.ogg instead of the lightning bolt sfx to further differentiate the spells

Changed the torch code in flashlight.dm to return true when it lights because otherwise ignition won't function properly on them if they never actually tell the spell that they got ignited

## Testing Evidence

image

Top- ignition working on different items after selecting the spell and using middle mouse (the failure cases are on the items after they're lit, so you don't waste devotion trying to light a lantern/torch already on fire)
Middle- sacred flame on a human
Bottom- sacred flame on a skeleton, I made the text different to show that the fire is stronger

## Why It's Good For The Game

ignition's problem was a literal one-word change in its /invoked proc which was why I originally decided to make this PR in the first place
I changed sacred flame because it didn't really make sense that holy fire was electrocuting people but I'm open to balancejaks since I don't know how heavily this will effect its power in combat

### DIFF
--- a/code/game/objects/items/devices/flashlight.dm
+++ b/code/game/objects/items/devices/flashlight.dm
@@ -238,7 +238,8 @@
 				var/mob/M = loc
 				M.update_inv_hands()
 			START_PROCESSING(SSobj, src)
-	..()
+			return TRUE
+	return ..()
 
 /obj/item/flashlight/flare/torch/afterattack(atom/movable/A, mob/user, proximity)
 	. = ..()
--- a/code/modules/spells/roguetown/acolyte/astrata.dm
+++ b/code/modules/spells/roguetown/acolyte/astrata.dm
@@ -1,6 +1,8 @@
+
 /obj/effect/proc_holder/spell/invoked/projectile/lightningbolt/sacred_flame_rogue
 	name = "Sacred Flame"
 	overlay_state = "sacredflame"
+	sound = 'sound/magic/bless.ogg'
 	req_items = list(/obj/item/clothing/neck/roguetown/psicross)
 	invocation = null
 	invocation_type = "shout"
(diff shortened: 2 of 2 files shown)


ddsaw:
I would personally prefer sacred flame not be a ray of fire and instead be like bladeburst, and be a somewhat spammable/low cost thing. Lets it be somewhat close to the original whilst not being as mindless.
Infact I just preferred the old sacred flame to this weird lightning bolt/fire ray. No idea why this was changed.

ShadowDeath6:
> Infact I just preferred the old sacred flame to this weird lightning bolt/fire ray. No idea why this was changed.

I assumed it was an oversight to have it electrocute like lightning bolt since it inherited that spell's projectile, which is why I included it in this bugfix PR. I'm open to tweaking the values on what I have here, of course, but I'm not sure what the old sacred flame did. 
A full rework/reversion of the spell might be better as its own PR rather than in this one, regardless.

roarbark:
the old sacred flame let you put 5 firestacks on anyone you clicked on, which required you to pat yourself down twice. this was a 5 second cooldown.

ShadowDeath6:
Consensus so far has been that it's really underwhelming 
SF is now 4 stacks on non-undead targets, 10 on undead

roarbark:
> > this still kind of is a LOT worse compared to what the miracle used to be...
> 
> I was under the impression that SF was previously extremely powerful because of it being spammable high fire stacks. Would this version of it be better with higher potency or lower cooldown? I assume people who use miracles in combat would prefer the latter for a spellcaster-y feel.

the thing about old sacred flame was that it self extinguished after a few seconds. it was in itself extremely weak, it's just a quirk with the atmos system made it deal way more damage after a few casts, because the victim's body temperature would rise a bunch...

be aware, the cooldown on this is still 25 seconds. you fire this at someone, they press x twice, taking minimal damage and you have to wait for half a minute before doing it again. this should probably be shorter if you want it do anything or even be as remotely strong as the other acolyte's miracles

ShadowDeath6:
> be aware, the cooldown on this is still 25 seconds. you fire this at someone, they press x twice, taking minimal damage and you have to wait for half a minute before doing it again. this should probably be shorter if you want it do anything or even be as remotely strong as the other acolyte's miracles

I dropped SF cooldown from 25s to 15s. I'm hesitant to further buff it since I'd rather not end up with something too overtuned and then the PR gets stalled forever.

PR #249, opened 2025-07-17
THE RICH GET RICHER, THE REACH GETS POORER — money rebalance for bandits, adventurers and rich character roles

coffeecoffeecoffee:
## About The Pull Request
Stops bandits from fucking claiming their own bounties for upwards of hundreds of mammon and taking the opportunity from literally anybody else in the game. It will now give you a little love bite and do a small amount of damage to your hand. People who were doing this probably should've gotten a talking-to from an admin.

Blacksteel items and the dragonscale necklace in the hoardmaster have been repriced. These items are meant to be late-game rewards for extremely successful bandits, and people are getting full sets of them in the first 10 minutes because you could buy them for **100-150 mammon** (?!?!?!) In conjunction with the fact bandits can and do complete dungeons at the start of the game, this is clearly not appropriate. You could literally get blacksteel from a bandit for cheaper than it would be to make it with ingots (which are 100 each.) Each piece is now however much it would be in blacksteel ingots (100 each, so for an armet, 200) + one hundred on top, as part of the difficulty of getting blacksteel is _supposed_ to be getting rare materials, and they're getting it for no labor at all — _Except the stealing that they should be doing._ The dragonsteel necklace has been doubled. Generally speaking, the bandits were made pretty cracked on Azure because very few people are willing to play antagonists over there...

**We do not have this problem.** In exchange, bandit role slots were increased from 5 to 7 to compensate for them having less access to god-tier equipment on par with the Duke. If this is too much, let me know.

Some sources of massive easy wealth for adventurers have been reduced. You should no longer find straight-up loose coin in dungeons per multiple requests, and prices of certain heads have been halfed - namely the minotaur and trolls. These are, frankly, fairly easy to kill and are still _at minimum_ worth twice as much as direbear heads. Also, Imps no longer give you SIX INFERNAL ASH EACH (why did this innocuous simplemob that does nothing but walk away from you give you 60 fucking mammon???) It still gives you one. _I also considered halfing the prices of gems and adjusting the prices of all the rings to account for them, but people said this also fucked over other roles._

[...]

PR #263, opened 2025-07-18
Buffs Rich and Noble virtue

Mannybrado:
## About The Pull Request
Makes both traits in title give you more money in the pouch they give you. Roughly 3x more money with a 50% chance of even more like the normal rich pouches that roles like heartfelt lord get.

ALSO makes rich pouches come with zennari instead of ziliqua.

## Testing Evidence
it compiles

## Why It's Good For The Game
Basically, I do not want to grind alone in a corner of the map to have some money to throw around for roleplay purposes.
I can see the argument that people would abuse this for gear or something, but actually, it doesn't matter if these people get their gear 10 minutes earlier than they would otherwise.


Rayforx:
I would look at "noble_income" and see if you can adjust some numbers. It rewards players for staying in a round, and gives nobles a lot more "passive" income on hand

Mannybrado:
This conflicts with #249 so. Later.

PR #264, opened 2025-07-18
Javelin confessor

Fatherless1Xx:
Built around javelins rather than bows/crossbows
image
image


ddsaw:
Just add a loadout choice to get silver javelins instead of tossblades or bow/crossbow. Throwing weapons are stupid anyways and completely ignore any and all stats/skills and such anyways.
I don't see why having a javelin would mean you are ungabunga barbarian strong(on an otherwise stealth dodger subclass..?), super fucking fast but dumb as bricks and.. imperceptive(on a thrower..?)

Fatherless1Xx:
> If you added it as another option would accept, dont want it to be a replacement.

the problem with this is throwing weapons scale with strength rather than perception so the confessor's base stats would be awful for this (church roles should be strong since they're responsible for enforcing the status quo)
if that's the case i could make it another loadout option for disicples?

DorfGuy1:
There was a trait on other servers that was 'strong thrower' that gave bonus damage to the throwing weapons.

KiteAE:
> > If you added it as another option would accept, dont want it to be a replacement.
> 
> the problem with this is throwing weapons scale with strength rather than perception so the confessor's base stats would be awful for this (church roles should be strong since they're responsible for enforcing the status quo) if that's the case i could make it another loadout option for disicples?

Fine by me, another issue is throwing weapons are completely busted ATM.

PR #274, opened 2025-07-19
Civilized barb now gives punches AP comparable to knives (50)

Fatherless1Xx:
please let me play a monk without being gutted


coffeecoffeecoffee:
buffing it is good but jesus that is a lot of fucking AP

PR #276, opened 2025-07-19
Fixes Iron Shield and Cudgel for Outlaw Marauder

roarbark:
## About The Pull Request
Fixes a little oversight.

## Testing Evidence
I couldn't test this because the current code doesn't compile properly. The change should work though.

## Why It's Good For The Game
Was a bug. It spawned you without a shield and no skill ups.


roarbark:
image

i tested this now

PR #277, opened 2025-07-19
Turns wretch regenerate into a HP potion, adds nasty easter

Fatherless1Xx:
Nasty eater to stop them throwing up when drinking blood
Regenerate isn't an aheal any more, it's a health potion


Fatherless1Xx:
image

Fatherless1Xx:
image


Silver now works against wretch vampires but ONLY when they're not disguised

john-pawson:
> image
> Silver now works against wretch vampires but ONLY when they're not disguised

Considering how silver is one of the things *used to specifically reveal vampyres*, I don't think that's the greatest of ideas. At least make it consume vitae or something to resist it.

roarbark:
drinking from people doesn't drain their vitae pool as a wretch, also you still throw up blood- take a look at grabbing.dm

Fatherless1Xx:
image

vampire wretches no longer throw up when drinking blood

Fatherless1Xx:
image
image

Victim now loses blood
Numbers need to be balance jakked but it works

Fatherless1Xx:
image
sunlight sets you on fire

Fatherless1Xx:
more reasonable blood loss
image

Fatherless1Xx:
code is optimized, it takes 60 vitae every minute rather than 1 per second

coffeecoffeecoffee:
this actually seems to my untrained non-wretch-playing eye like a really good and cool PR
i'm a fan

PR #281, opened 2025-07-20
Emberwine Cauldron Recipe [HOTFIX]

anything86:
## About The Pull Request
Changes around a few of the ingredient weights for cauldron recipes to enable making Emberwine poison at a cauldron.

## Testing Evidence
[https://i.gyazo.com/e6bc3c58a40f9e13ca39a47e23737903.png](url)

## Why It's Good For The Game
Someone asked me for this. Lets people without brew Emberwine in their backyards. The recipe already existed in the code, it just wasn't associated with any ingredient.


Extisek:
This PR makes it impossible to make big_stam_poison since you can only get 4 points (3 from Paris and 1 from mineral dust) where you only needed swampweed dust and paris before.

Wouldn't it be better to introduce new ingredients rather than changing the existing ones that messes with the recipes? Rosa for instance is not used from what I know of, you could add existing things for alchemy such as foods/plants from soilsons/etc.

PR #282, opened 2025-07-20
Parry & Dodge Nerf, Other Misc Weapon Changes

anything86:
## About The Pull Request
Debloats a lot of weapons and shields' wdefense, slightly buffs the wdefense of a few weapons that had completely unusable values. Reduces the impact dodge expert has on dodging from functionally +5% base dodge chance per SPD to +2.5% per SPD.

These are just value changes so as always, check the changed files for a detailed log, but a few pointers:

- Shields lose 3 points of wdefense across the board. This means a kite shield now offers a parry of 80% assuming equal shield and weapon skill. The point of shields is now to absorb blows without compromising your weapon's durability, and also offer protection against projectiles. Their parry is still good, but no longer '90% parry chance even if the opponent has higher skill' levels of good.
- Polearms lose 1 point of wdefense across the board. They are still king in parrying with a weapon, but going from 90% parry to 80% parry is still a significant nerf (twice the amount of blows going through).
- In particular the glaive had a 120% parry before. It goes down to 75% now, as it has overall better weapon intents than the spear.
- The naginata gets a small damage debuff but gains a niche as a swift polearm.
- Swords stay relatively the same, with minor nerfs in wdefense here and there. Arming swords are better at parrying than longswords, which are now better at parrying than greatswords and zweihanders. This means there is a tradeoff to using zweihanders and they are no longer strictly the best swords by virtue of having reach 2 and big peel.
- Maces also stay relatively the same, with minor buffs to parrying with maces due to blunt weapons being in a bad spot right  now.
- Daggers get decent parry chance, all the way up to 75% with the special Etruscan parrying dagger. Still not good parrying weapons but should allow the parrying daggers to actually perform their function now.
- No change to whips and flails.

As for the dodge changes:

- The weight of dodge expert was significantly reduced. This means having dodge expert is no longer a guarantee of 90% dodge chance. At 15 speed this nerf represents -30% dodge chance.
- In exchange, dodging costs 6.5 stamina, down from 10. This is still more than parrying, but should overall be less taxing for combat.

## Testing Evidence
76283c20ae86601534db54d3caae5c2c

## Why It's Good For The Game
Both parry and dodge chances were bloated to the point of being 80-90% in most cases. This, coupled with other issues, led to unsatisfying combats that only resolved once someone was swarmed to the point their stamina broke. This change addresses one of these issues by reducing wdefenses across the board.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/axes.dm
+++ b/code/game/objects/items/rogueweapons/melee/axes.dm
@@ -151,7 +151,7 @@
 	smelt_bar_num = 2
 	gripped_intents = list(/datum/intent/axe/cut/battle ,/datum/intent/axe/chop/battle, /datum/intent/axe/bash, /datum/intent/sword/peel)
 	minstr = 9
-	wdefense = 4
+	wdefense = 3.5 //Slightly less than a longsword, better intents.
 
 /obj/item/rogueweapon/stoneaxe/battle/getonmobprop(tag)
 	if(tag)
@@ -186,7 +186,7 @@
 	smeltresult = /obj/item/ingot/steel
 	gripped_intents = list(/datum/intent/axe/cut/battle ,/datum/intent/axe/chop/battle, /datum/intent/axe/bash)
 	minstr = 12
-	wdefense = 5
+	wdefense = 4
 
 /obj/item/rogueweapon/stoneaxe/oath/getonmobprop(tag)
 	if(tag)
@@ -245,6 +245,7 @@
 	gripped_intents = list(/datum/intent/axe/cut,/datum/intent/axe/chop, /datum/intent/mace/warhammer/pick, /datum/intent/axe/bash)
 	smeltresult = /obj/item/ingot/steel
 	wlength = WLENGTH_NORMAL
+	wdefense = 3
 	toolspeed = 2
 
 
@@ -368,7 +369,7 @@
 	max_blade_int = 400
 	smeltresult = /obj/item/ingot/silver
 	gripped_intents = null
-	wdefense = 4
+	wdefense = 5.5
 	is_silver = TRUE
 	blade_dulling = DULLING_SHAFT_METAL
 
@@ -416,7 +417,7 @@
 	smeltresult = /obj/item/ingot/iron
 	associated_skill = /datum/skill/combat/axes
 	blade_dulling = DULLING_SHAFT_WOOD
-	wdefense = 6
+	wdefense = 4
 	pickup_sound = 'modular_helmsguard/sound/sheath_sounds/draw_greatsword.ogg'
 	sheathe_sound = 'modular_helmsguard/sound/sheath_sounds/put_back_to_leather.ogg'
 
--- a/code/game/objects/items/rogueweapons/melee/blunt.dm
+++ b/code/game/objects/items/rogueweapons/melee/blunt.dm
@@ -65,7 +65,7 @@
 	parrysound = list('sound/combat/parry/parrygen.ogg')
 	swingsound = BLUNTWOOSH_MED
 	minstr = 7
-	wdefense = 2
+	wdefense = 3
 	wbalance = WBALANCE_HEAVY
 	blade_dulling = DULLING_SHAFT_METAL
 	intdamage_factor = 1.35
@@ -118,7 +118,7 @@
 	icon_state = "smace"
 	wbalance = WBALANCE_HEAVY
 	smeltresult = /obj/item/ingot/steel
-	wdefense = 3
+	wdefense = 4 //70% parry chance wielded, because blunt weapons are in a bad spot right now.
 	smelt_bar_num = 2
 
 /obj/item/rogueweapon/mace/steel/palloy
@@ -135,7 +135,7 @@
 	force = 24
 	gripped_intents = null
 	possible_item_intents = list(/datum/intent/mace/strike, /datum/intent/mace/smash, /datum/intent/effect/daze)
-	wdefense = 4
+	wdefense = 5.5
 	smeltresult = /obj/item/ingot/silver
 	smelt_bar_num = 2
 	is_silver = TRUE
@@ -197,7 +197,7 @@
 	w_class = WEIGHT_CLASS_NORMAL
 	wbalance = WBALANCE_NORMAL
 	minstr = 7
-	wdefense = 1
+	wdefense = 2
 	resistance_flags = FLAMMABLE
 	blade_dulling = DULLING_SHAFT_WOOD
 	grid_width = 32
@@ -225,7 +225,7 @@
 	blade_dulling = DULLING_SHAFT_REINFORCED
 	wbalance = WBALANCE_SWIFT
 	minstr = 7
-	wdefense = 5
+	wdefense = 4
 
 /obj/item/rogueweapon/mace/cudgel/getonmobprop(tag)
 	. = ..()
@@ -438,7 +438,7 @@
 	blade_dulling = DULLING_SHAFT_CONJURED
 
 /obj/item/rogueweapon/mace/warhammer/steel
-	force = 25
+	force = 27 //Specialized for one-handed use.
 	possible_item_intents = list(/datum/intent/mace/strike, /datum/intent/mace/warhammer/pick, /datum/intent/mace/warhammer/stab)
 	name = "steel warhammer"
 	desc = "A fine steel warhammer, makes a satisfying sound when paired with a knight's helm."
--- a/code/game/objects/items/rogueweapons/melee/knives.dm
+++ b/code/game/objects/items/rogueweapons/melee/knives.dm
@@ -259,6 +259,7 @@
 	possible_item_intents = list(/datum/intent/dagger/thrust,/datum/intent/dagger/cut, /datum/intent/dagger/thrust/pick, /datum/intent/dagger/sucker_punch)
 	force = 15
 	max_integrity = 100
+	wdefense = 4
 	name = "iron dagger"
 	desc = "This is a common dagger of iron."
 	icon_state = "idagger"
@@ -286,6 +287,7 @@
 	icon_state = "sdagger"
 	force = 20
 	max_integrity = 150
+	wdefense = 5.5
 	smeltresult = /obj/item/ingot/steel
 
 /obj/item/rogueweapon/huntingknife/idagger/steel/holysee
@@ -313,6 +315,7 @@
 	icon_state = "stiletto"
 	force = 25
 	max_integrity = 200
+	wdefense = 6
 	smeltresult = /obj/item/ingot/steel
 
 /obj/item/rogueweapon/huntingknife/idagger/steel/parrying //direct upgrade but more costly.
@@ -321,12 +324,12 @@
 	force = 15
 	throwforce = 15
 	icon_state = "spdagger"
-	wdefense = 6
+	wdefense = 7
 
 /obj/item/rogueweapon/huntingknife/idagger/steel/parrying/vaquero
 	name = "sail dagger"
 	desc = "An exceptionally protective parrying dagger popular in the Etruscan Isles, this dagger features a plain metal guard in the shape of a ship's sail."
-	wdefense = 7
+	wdefense = 7.5
 	force = 17
 	throwforce = 17
 	icon_state = "sail_dagger"
@@ -425,6 +428,7 @@
 	name = "elvish dagger"
 	desc = "This beautiful dagger is of intricate, elvish design. Sharper, too."
 	force = 22
+	wdefense = 5.5
 	icon_state = "elfdagger"
 	item_state = "elfdag"
 	last_used = 0
--- a/code/game/objects/items/rogueweapons/melee/polearms.dm
+++ b/code/game/objects/items/rogueweapons/melee/polearms.dm
@@ -205,7 +205,8 @@
 	pixel_x = -16
 	inhand_x_dimension = 64
 	inhand_y_dimension = 64
-	wdefense = 10
+	wdefense = 5.5 //This was at 10 (so 13 wielded) before for some insane reason. No, stop, wooden staffs aren't A.T. Fields. It was higher than a kite shield, get real.
+	max_integrity = 180 //It's a piece of wood. Unique staffs get higher durability.
 	bigboy = TRUE
 	gripsprite = TRUE
 	associated_skill = /datum/skill/combat/polearms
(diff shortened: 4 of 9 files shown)


anything86:
> [Rotwood-Vale/Ratwood-Keep#3007](https://github.com/Rotwood-Vale/Ratwood-Keep/pull/3007) May I suggest porting this PR if you're doing polearm wdefense changes? To avoid them being used as semi-shields while one-handed _and_ letting them keep high wdefense when they're wielded.
> 
> On another note, I should point out: Naginata having a swift balance as a polearm may go ahead and raise a problem with shoe-poking SPD builds that bypass most of your parries and just cut your leg off because you didn't put on plated boots.
> 
> On yet _another_ note... why do the regular daggers get such a high wdefense? Part of the disadvantages of daggers is being bad at parrying, (Exclusion being parrying daggers, obviously) but now a regular steel dagger is better at parrying than a shortsword? A SABRE is only 5% better at parrying than a regular dagger?? (Should also mention that this would mean a buff to dual wielding knife enjoyers, considering how knives already are the easiest to dual-wield without sacrificing two belt/back slots, and now would also be benefitting from a very good parry chance without having to sacrifice damage like with the parrying daggers)

The basic heavy leather boots still have around 30 slash defense, which mostly stops boot slashing. No high speed class gets good polearm skill, either (even champion knight gets none).

As for the dagger thing: 5.5 is really not that much. Sabres get 6.5, so that is +10% parry chance over a steel dagger. Not to mention the closer to 100%, the higher the relative increase. I am not too concerned about 55% parry chance when most swords can get at least 65%.

john-pawson:
> The basic heavy leather boots still have around 30 slash defense, which mostly stops boot slashing. No high speed class gets good polearm skill, either (even champion knight gets none).

The basic boots have not protection, and heavy leather boots are only used by keep roles. The more common normal leather boots also only have 10 slashing protection. I should also mention that with statpacks and easy speed boosts being a thing, it's not exactly hard to get high speeds to take advantage of that. On another note, here's a list of some of the classes that get journeyman or more polearms while having a speed bonus: (This is, of course, not counting anyone that might just take militiaman apprentice to get journeyman spears)
Knaves (SPD 3, Journeyman)
Roguemage (SPD 1, Access to haste, Journeyman)
Sellsword (SPD 2, Expert)
Disciples (SPD 2, Expert)
Desert Riders (SPD 2, Expert)
Hierophants (SPD 2, Access to haste, Joruneyman)

>As for the dagger thing: 5.5 is really not that much. Sabres get 6.5, so that is +10% parry chance over a steel dagger. Not to mention the closer to 100%, the higher the relative increase. I am not too concerned about 55% parry chance when most swords can get at least 65%.

Oh, yeah, MB, was reading the falchion (Which is still a bit... weird that a falchion would have lesser chance to parry than a dagger). But I honestly don't think "swords usually get at least 65% parry chance" is... anywhere near a reason to buff dagger parrying so much. You say 5.5 isn't that much yet you give all of the other one-handed swords less than it? 
Hunting sword: 2 -> 3.5
Messer: 4 -> 4.5
Shortsword: 4.5 -> 5

Why do all of these get minor .5 - 1.5 raised wdefense and daggers would get an extra 2.5 all of the sudden? Shouldn't these full-length and main weapons also get a close if not higher amount?

>Also: What do you mean exactly by using polearms as shield in one hand? The highest wdefense is 5.5 for quarterstaffs, which is still worse than something like a sabre when not wielded in two hands. Remember, wielding a weapon gives it +3 wdefense. All two-handed weapons have their wdefense balanced around that.

I meant that the PR that I suggested stops people from using polearms as shields in one hand (For example, that for some reason wooden staffs had the same defense than a tower shield, allowing you to have ridiculously high parrying chance + no bog slowdown + being easy to craft.) while also not having to nerf their wielded defense. 
So you can do things such as making wooden staffs/spears really bad at parrying while in one hand, and really good at parrying when they're wielded properly.

ddsaw:
I like this conceptually, but.

This is a huge relative buff to dodge classes, which do not need it whatsoever (already being the strictly superiour option, especially with shields now gutted). 30% dodgenerf.. Is not enough if your opponent has something like 275% dodge. They still get 90% against basically everybody.
Math this out at expert and journeyman weaponskills. The usual 14-15 spd dodgeexpert versus somebody of say - 10 spd? They're never getting hit. Ever. Probably won't even get hit if the enemy has 13 spd and they use a weapon the dodger has zero familiarity with. Peak.

Meanwhile, parry users lose the only reliable way to gain 90% parry (which mind you, is a good thing, this shouldn't be the default). Swift weapons and SPD builds are already much better than STRbuilds. The sabre is by far the best sword, for example, reliably giving you as a dodgebuild another 1.5 weaponskill level due to swift balance.

anything86:
I will lower the dagger parry but I have to say it is a very strange nitpick. People don't tend to use low parry weapons for parrying, and obviously dodge classes would rather dodge when using a dagger, so I am not sure it warrants 4 paragraphs of positing about the parrying capacity of a steel dagger on. In real life it is a good implement to parry with and the thought process is that it is a sidearm for knights, who often get journeyman or lower knife skill.

anything86:
> I like this conceptually, but.
> 
> This is a huge relative buff to dodge classes, which do not need it whatsoever (already being the strictly superiour option, especially with shields now gutted). 30% dodgenerf.. Is not enough if your opponent has something like 275% dodge. They still get 90% against basically everybody. Math this out at expert and journeyman weaponskills. The usual 14-15 spd dodgeexpert versus somebody of say - 10 spd? They're never getting hit. Ever. Probably won't even get hit if the enemy has 13 spd and they use a weapon the dodger has zero familiarity with. Peak.
> 
> Meanwhile, parry users lose the only reliable way to gain 90% parry (which mind you, is a good thing, this shouldn't be the default). Swift weapons and SPD builds are already much better than STRbuilds. The sabre is by far the best sword, for example, reliably giving you as a dodgebuild another 1.5 weaponskill level due to swift balance.

I have to ask where you are getting 275% dodge from. Can you elaborate on this? At 15 speed, if we choose to ignore the relative calculation for dodging, you have '225% dodge chance'. An opponent with a polearm at 11 speed will be reducing that by roughly 150%, leaving you with 75% dodge chance. Obviously 11 speed isn't the optimal for countering dodge, but this isn't the old 'always have 90%', and other factors such as haste can also factor in.

anything86:
I might also be revisiting Sentinel of Wits here. Before, it was somewhat mediocre because everyone had 90% parry by default anyway. Now it might actually be an issue due to pushing those 70%-75% parries into 90% territory with as little as 14 intelligence, which a lot of roles can get.

ddsaw:
> > I like this conceptually, but.
> > This is a huge relative buff to dodge classes, which do not need it whatsoever (already being the strictly superiour option, especially with shields now gutted). 30% dodgenerf.. Is not enough if your opponent has something like 275% dodge. They still get 90% against basically everybody. Math this out at expert and journeyman weaponskills. The usual 14-15 spd dodgeexpert versus somebody of say - 10 spd? They're never getting hit. Ever. Probably won't even get hit if the enemy has 13 spd and they use a weapon the dodger has zero familiarity with. Peak.
> > Meanwhile, parry users lose the only reliable way to gain 90% parry (which mind you, is a good thing, this shouldn't be the default). Swift weapons and SPD builds are already much better than STRbuilds. The sabre is by far the best sword, for example, reliably giving you as a dodgebuild another 1.5 weaponskill level due to swift balance.
> 
> I have to ask where you are getting 275% dodge from. Can you elaborate on this? At 15 speed, if we choose to ignore the relative calculation for dodging, you have '225% dodge chance'. An opponent with a polearm at 11 speed will be reducing that by roughly 150%, leaving you with 75% dodge chance. Obviously 11 speed isn't the optimal for countering dodge, but this isn't the old 'always have 90%', and other factors such as haste can also factor in.

Doesn't dodgechance start at 50%? So it would be 15*15+50.

ddsaw:
> Read the PR. The weight has been reduced to 12.5 from 15. The normal weight is 10.

I said "a 30% nerf.. is not enough when the opponent (previously) had 275%)" The 30%(it's 40% in this situation) is what accounts for the nerf from 15->12.5 and takes the chance to 12.5*15+50 = 237. 
237.5% vs spd of 11 polearm expert and they have 0 skill in polearms themselves (just about the worst possible situation you can conjure up, there isn't really any 15spd polearm experts running around and if its swords or daggers you are likely to have that skill yourself thus granting you extra dodge.) still leaves them comfortable with 87%.

anything86:
> > Read the PR. The weight has been reduced to 12.5 from 15. The normal weight is 10.
> 
> I said "a 30% nerf.. is not enough when the opponent (previously) had 275%)" The 30%(it's 40% in this situation) is what accounts for the nerf from 15->12.5 and takes the chance to 12.5*15+50 = 237. 237.5% vs spd of 11 polearm expert and they have 0 skill in polearms themselves (just about the worst possible situation you can conjure up, there isn't really any 15spd polearm experts running around and if its swords or daggers you are likely to have that skill yourself thus granting you extra dodge.) still leaves them comfortable with 87%.

A 11 speed polearm user is not 'the worst situation I could conjure up', no. There are also other ways to check someone if they are investing all their stats into speed. It is also a bit misleading to say "250% dodge chance", don't you think? All in all, I don't see how this is productive. The point is that other speedy characters can now hit through dodge, and you actually need speed investment to get to 90% dodging now (as opposed to before where 12 speed was enough, enabling things like Vampire Nudist Wretch).

Fatherless1Xx:
good change anything that makes combat more deadly is a step in the right direction

ddsaw:
> > > Read the PR. The weight has been reduced to 12.5 from 15. The normal weight is 10.
> > 
> > 
> > I said "a 30% nerf.. is not enough when the opponent (previously) had 275%)" The 30%(it's 40% in this situation) is what accounts for the nerf from 15->12.5 and takes the chance to 12.5*15+50 = 237. 237.5% vs spd of 11 polearm expert and they have 0 skill in polearms themselves (just about the worst possible situation you can conjure up, there isn't really any 15spd polearm experts running around and if its swords or daggers you are likely to have that skill yourself thus granting you extra dodge.) still leaves them comfortable with 87%.
> 
> A 11 speed polearm user is not 'the worst situation I could conjure up', no. There are also other ways to check someone if they are investing all their stats into speed. It is also a bit misleading to say "250% dodge chance", don't you think? All in all, I don't see how this is productive. The point is that other speedy characters can now hit through dodge, and you actually need speed investment to get to 90% dodging now (as opposed to before where 12 speed was enough, enabling things like Vampire Nudist Wretch).

Misleading? You know exactly how the calculation works itself. As for this enabling other speed characters to hit through dodge? I don't see how this is enough as I pointed out. Two equal speed equal skill characters always hit each other at 50%.
Productive? I just consider this, in it's current state, to severely degrade balance tipping it even further towards light armor dodgebuilds. Argumenting about it is just about as productive as you could be balancejakking.
Yes, it nerfs dodgebuilds in some situations.
Yes, it does matter.
But it doesn't matter enough to balance dodge in a world where nobody reliably has 90% parry and also dodge barely costs more than parry now.

anything86:
> Misleading? You know exactly how the calculation works itself. As for this enabling other speed characters to hit through dodge? I don't see how this is enough as I pointed out. Two equal speed equal skill characters always hit each other at 50%. Productive? I just consider this, in it's current state, to severely degrade balance tipping it even further towards light armor dodgebuilds. Argumenting about it is just about as productive as you could be balancejakking. Yes, it nerfs dodgebuilds in some situations. Yes, it does matter. But it doesn't matter enough to balance dodge in a world where nobody reliably has 90% parry and also dodge barely costs more than parry now.

Okay. Would you rather dodge remain as it was before? The parry chances have to be debloated, that is obvious. I further reduced the weight to 12. What exactly do you want here, since no alternative satisfies you? I know for a fact there is plenty of counterplay to dodge, and having to wear light armor is in fact a detriment (no light armor has leg coverage with more than 50-60 stab protection, rendering them vulnerable to leg stabbing). You seem to refuse accepting Haste or moondust or having some speed investment as acceptable counterplay. It really makes me wonder what you want out of this?

john-pawson:
> Two equal speed equal skill characters always hit each other at 50%.

I do also feel the need to point out this is false. The dodge expert trait scales up the dodging gain, but not the dodging lost. So for example:
Player 1 has 12 SPD 
Player 2 has 12 SPD
Player 1 gets 12 * 12% chance to dodge. = 144% chance + 50% base
144% chance + 50% base = 194% chance to dodge.
Player 2 has 12 SPD, so that means we subtract 120% (Since it doesn't scale with dodge expert, see picture if you'd like to check)
194% - 120% = 74% chance to dodge.
Exact same thing applies to when Player 1 attacks player two.
That's why usually two dodge-experts was just a match of "who gets tired first", since when the scaling was 15% there was a big chance both of them just had 90% chance to dodge.
image

anything86:
> > Two equal speed equal skill characters always hit each other at 50%.
> 
> I do also feel the need to point out this is false. The dodge expert trait scales up the dodging gain, but not the dodging lost. So for example: Player 1 has 12 SPD Player 2 has 12 SPD Player 1 gets 12 * 12% chance to dodge. = 144% chance + 50% base 144% chance + 50% base = 194% chance to dodge. Player 2 has 12 SPD, so that means we subtract 120% (Since it doesn't scale with dodge expert, see picture if you'd like to check) 194% - 120% = 74% chance to dodge. Exact same thing applies to when Player 1 attacks player two. That's why usually two dodge-experts was just a match of "who gets tired first", since when the scaling was 15% there was a big chance both of them just had 90% chance to dodge. image

In a situation like this you can either:
a) Have at least one speed point over them while using a swift weapon. This should tip things in your favor by 25% per speed point, bringing this to 49% chance to hit.
b) Use a weapon in which you are trained but they are not. For example, Royal Champion knight gets expert axes, while basically no other dodge class gets any. With that in mind you'd apply a -40% penalty, bringing this to 34% change to dodge at same speed.

john-pawson:
> In a situation like this you can either: a) Have at least one speed point over them while using a swift weapon. This should tip things in your favor by 25% per speed point, bringing this to 49% chance to hit. b) Use a weapon in which you are trained but they are not. For example, Royal Champion knight gets expert axes, while basically no other dodge class gets any. With that in mind you'd apply a -40% penalty, bringing this to 34% change to dodge at same speed.

...yes? I mean, I know. I just wanted to point out that the phrase "two equal speed skill characters always hit 50%" is false. It can sometimes be misleading.

anything86:
> ...yes? I mean, I know. I just wanted to point out that the phrase "two equal speed skill characters always hit 50%" is false. It can sometimes be misleading.

Thing is basically none of those alternatives I gave would have worked in the past because the formula was too skewed at 15 x speed. Now it isn't.

john-pawson:
> Thing is basically none of those alternatives I gave would have worked in the past because the formula was too skewed at 15 x speed. Now it isn't.

Fair point!

PR #288, opened 2025-07-20
Rite of the dark crystal

Fatherless1Xx:
Summons a necro crystal for zizites who don't use heavy armor
image
image


Fatherless1Xx:
image

PR #290, opened 2025-07-20
Craftable dark crystals

Fatherless1Xx:
Don't leave your lux lying around

image

### DIFF
--- a/code/modules/roguetown/roguecrafting/alchemy.dm
+++ b/code/modules/roguetown/roguecrafting/alchemy.dm
@@ -259,3 +259,13 @@
 	craftdiff = 6
 	verbage_simple = "transmute"
 
+/datum/crafting_recipe/roguetown/alchemy/necro_crystal
+	name = "Dark crystal"
+	category = "Transmutation"
+	result = list(/obj/item/necro_relics/necro_crystal = 1)
+	reqs = list(/obj/item/reagent_containers/lux = 1, /obj/item/alch/viscera = 1)
+	skillcraft = /datum/skill/craft/alchemy
+	craftdiff = 4
+	verbage_simple = "craft"
+	always_availible = TRUE
+


benj8560:
1 viscera 1 lux? This is far far FAR to spamable for what allows to you assemble a army of player controlled frag machines. 
There can easily be upwards of 5 wretches all together here! If it was about ''don't leave lux laying around'' that'd be nice but what's far more likely is someone will just use the surgeons bag at zchurch to grind expert medicine and start collecting it themselfs. 

The idea in your other PR was far more appealing due to ritual cooldown preventing it being combined tp easily with armament rites.

benj8560:
This being taken to its fullest extreme would just be barber surgeon, zizoite, Heresiarch (optional), ritualist. Boom you are  genuinely better than the average necromancer round start and can sit around the Wretch church convincing people to raise a army with you.
Hell as a barber you can likely just start asking for lux donations over SCOM and rake it in free of effort.

Cre77:
with how much of a pushover the non-lich skeleton is, i'd actually love to see more of them running around

Fatherless1Xx:
> This being taken to its fullest extreme would just be barber surgeon, zizoite, Heresiarch (optional), ritualist (also optional but armor spam). Boom you are genuinely better than the average necromancer round start and can sit around the Wretch church convincing people to raise a army with you. Hell as a barber you can likely just start asking for lux donations over SCOM and rake it in free of effort.

this should be something that's seen as suspicious and draws the attention of the inquisition

Enite:
That round was peak, it was fun having an actual threat for once. Just remove the ability to harvest Lux from goblins and other dungeon mobs like those vikings and it should be fine. Gives them an incentive to capture random people and the guards a reason to want to stop them from doing that.

benj8560:
> Self farming gives them a debuff, and they weren't sitting in the cave all round only making skeletons. They directly led them out multiple times and were vulnerable to being killed. It's not like the town was drowned in skeletons either, or even 'lost'. In the end the wretches were still outnumbered and the inevitable attrition of them not being able to revive their own meant it only got worse for them as time went on.
> 
> I think it's a good change that the northern wilds are actually dangerous to travel through now, since the wretches have an incentive to kidnap people they see and the inquisition/retinue an incentive to patrol the area if people are getting taken.

I'm not quoting exact happenings here I'm pointing out the kind of extreme behavior that could happen suppose I should more clearly differentiate the two. It is only a matter of time before this kind of shit happens. And there was absolutely _some_ degree of sitting around while skeletons go do the dirty work. This debuff is irrelevant to players who just sit around the inhuman church and even to those who not is a TINY price to pay for more numbers.

benj8560:
> i hope they do (it still won't be enough for wretches to be a creditable threat to the bored adventurer horde)

dreamseeker_01ADLPWbg8
starting to see what you mean here..but I do think darksteel spam is busted. Don't even make sense from a lore perspective as well skeletons don't have lux!

At its core at least from balance this PR is starting to grow on me in some degree. My primary issue is that this is every round now ever since its been added its the exact same fucking gimmick of spamming skeletons and going to war with the town. 

Although it's new and fun at the moment I do think this is gonna get old pretty soon...again as I have been pointing out wretch's are not **PRIMARY ANTAGS** they are _flavor_ for the round. They are currently filling the entire pot right now however.

This does regardless mostly work and Wretches are not exactly steamrolling here but its still just a LOT of firepower for a secondary antag

Enite:
Secondary antag? What's the primary antag? They're all disabled. We have Wretches and Bandits. That is IT. Those are the antags. And even now they still pose no credible threat to an organized town almost every round.

Aestellex:
Unmerging this was a massive mistake - as the person who DOES the necromancy army. 

2 crystals is not enough when skeletons aren't scaled to the same place as people are - you have to fight on average 20+ as wretch, and magic is hot garbage. We had a round with 15-20 skeletons vs the town, and killed maybe two unarmored towners. That is how weak they are currently.

KiteAE:
This created far far too many skeleton armies, good idea but bad execution.

KiteAE:
> > This created far far too many skeleton armies, good idea but bad execution.
> 
> What if we changed the materials? Hearts instead of lux, maybe?

Aslong as its harder to get im cool with it.

PR #291, opened 2025-07-21
Kobold Removal

SaklasDev:
## About The Pull Request

This removes kobolds in their entirety.

## Testing Evidence

image

## Why It's Good For The Game

Prevents ageplay like:
image
From ever happening again. Since this little thing that supposedly **never** happens keeps **happening** with this one specific race.


ThinkArbies:
This PR might be a joke. It might not be.

But Im 100% serious in saying we should go for it. I'm sick of seeing gross underage-like AI headshots.

And also any niche they fill in is just filled by goblins anyways.

pyrzal:
just remove it thanks

pyrzal:
when i see the screenshot i think "Yeah this is just how kobolds talk" but _that is why kobolds should be removed_ because (like @EmilyFontaine said) kobolds are just the thinly veiled ageplay race

cynical24:
i agree with Smiting

TamedAchilles:
So where's the ageplay at. I see a kobold talking with an accent like any other race unless you left it out of the cropped logs you posted here lol

zeratino:
> > > Just ban pedophiles and dox them, I don't see why you need to kill off a chunk of the server's pop.
> > 
> > 
> > This "chunk" of the server's pop consisting of people for whom the existence of an ageplay lizard race makes or breaks their willingness to play said server is a miniscule percentage that would not be missed by anyone
> 
> Shouldn't really be the point. If ageplay is an actual issue, remove the pedophiles and ban them permanently. People can play a lizard race and not be pedophiles and assuming that anyone that plays a kobold does that makes the Staff Team look like fools when tackling this whole ageplay/pedophilia issue.

i think it's easier to remove the few races that enable this behavior than to constantly observe every kobold for signs of pedophilia

TamedAchilles:
> > > > Just ban pedophiles and dox them, I don't see why you need to kill off a chunk of the server's pop.
> > > 
> > > 
> > > This "chunk" of the server's pop consisting of people for whom the existence of an ageplay lizard race makes or breaks their willingness to play said server is a miniscule percentage that would not be missed by anyone
> > 
> > 
> > Shouldn't really be the point. If ageplay is an actual issue, remove the pedophiles and ban them permanently. People can play a lizard race and not be pedophiles and assuming that anyone that plays a kobold does that makes the Staff Team look like fools when tackling this whole ageplay/pedophilia issue.
> 
> i think it's easier to remove the few races that enable this behavior than to constantly observe every kobold for signs of pedophilia

You'll think it's easier now,

However if and when you remove a playable race in this manner what do you think is magically going to happen with the behavior the community is presenting while playing as them? It isn't going to magically disappear because kobolds are removed. It'll go to another playable race such as the Goblins. Then someone will make a PR to push for the removal of Goblins because they think ageplay is happening with them now. Then it'll be Elves/Drow. Then Half Orcs. So on, so fourth.

Now you're 4-5 Races in and you still have pedophiles creating problems with alleged ageplay because people wanted to take the easy route as opposed to immediately putting their foot down, and just getting rid of people who show even the slightest inkling of this behavior. If the Staff Team can't find the time, or doesn't want to find the time to observe their community and protect it's members, they shouldn't be on the staff team. Period.

Quink-Fix:
Local Playerbase Says "I'm Sick Of Seeing X" And Removes An Entire Playable Race Instead Of Just Fucking Reporting X

RyszardParys:
what happened to make bro remove them lmao

PR #296, opened 2025-07-21
Piercing Armor Buffs, Broadhead AP reduction, Toxophilite capped

anything86:
## About The Pull Request
As it says on the tin. Azure had the bright idea of giving armors piercing values to protect against projectiles but I guess at some point in their process they got balancejakked all wrong, and we wound up with gems such as padded gambesons having 80 piercing protection and the toughest boots having 50. Broadheads were piercing plated boots at 10 perception, let alone at 15. This PR does away with the broadhead arrow AP, giving actual reason to use bodkins, and juices up the piercing protection of a lot of the plate and maille armors. Crossbows are mostly unaffected because they have a whopping 50 AP.

Also caps Toxophilite to Journeyman, same as all other weapon virtues. It's a dumb virtue that just straight up gives you +1 bow skill. It's too much of a no brainer on archer classes and is thus unhealthy.

## Testing Evidence
It's just variable changes.

## Why It's Good For The Game
Nerfs .50 Barrett w/ scope. Far too easy to minimize draw time, far too effective if you just aim for the feet due to armor not protecting against it, bypasses parries and dodge.

### DIFF
--- a/code/__DEFINES/roguetown.dm
+++ b/code/__DEFINES/roguetown.dm
@@ -12,40 +12,40 @@
 #define ARMOR_PADDED_GOOD list("blunt" = 80, "slash" = 50, "stab" = 50, "piercing" = 80, "fire" = 0, "acid" = 0)
 #define ARMOR_PADDED list("blunt" = 60, "slash" = 40, "stab" = 30, "piercing" = 50, "fire" = 0, "acid" = 0)
 #define ARMOR_PADDED_BAD list("blunt" = 40, "slash" = 30, "stab" = 20, "piercing" = 40, "fire" = 0, "acid" = 0)
-#define ARMOR_LIGHTCUIRASS list("blunt" = 30, "slash" = 70, "stab" = 70, "piercing" = 20, "fire" = 0, "acid" = 0)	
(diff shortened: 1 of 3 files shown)


ddsaw:
Finally. I especially hate the fact that all boots in the game except for blacksteel plate boots are very susceptible to arrows. This fixes that, too.

anconfuzedrock:
> > chain having equal pierce resist to brigandine is a bit weird, chainmail would be weak to stabbing & piercing (though it's currently super duper resistant to stabbing, which is weird... maybe that needs a bit of a nerf. that's not here, though!
> 
> How so? I think you have a wrong idea here, chainmail is in real life very resistant to both stabbing and piercing. Broadheads didn't pierce chainmail easily. You can do it, atm, with something like 15 perception, but just barely - which I think is a good balancing point. Meanwhile bodkins penetrate it much more easily.

chain is resistant to being stabbed, but still less resistant than actual plates- you at least CAN stab through chain with a really good hit. so it's odd that chainmail has equal piercing resist to actual plate armor.

DorfGuy1:
I've been playing a lot of archery lately.  Gotta say, things feel bad as of current.

PR #297, opened 2025-07-21
Unapologetic salt PR

Fatherless1Xx:
Mindlink breaks if either person transforms into a Zad or is a zad
Message is made t2 so witches cannot get it
This is because of witches using zad form in combination with these spells to play like a crime reporting drone
As the person who made mindlink this was never it's intended use
image
image

### DIFF
--- a/code/modules/spells/spell_types/wizard/utility/message.dm
+++ b/code/modules/spells/spell_types/wizard/utility/message.dm
@@ -6,7 +6,7 @@
 	releasedrain = 30
 	recharge_time = 60 SECONDS
 	warnie = "spellwarning"
-	spell_tier = 1
+	spell_tier = 2
 	associated_skill = /datum/skill/magic/arcane
 	overlay_state = "message"
 	var/identify_difficulty = 15 //the stat threshold needed to pass the identify check
--- a/code/modules/spells/spell_types/wizard/utility/mindlink.dm
+++ b/code/modules/spells/spell_types/wizard/utility/mindlink.dm
@@ -67,6 +67,12 @@
 		if(HL.real_name == second_target_name)
 			second_target = HL
 
+	// Check if either target is a zad
+	if(istype(first_target, /mob/living/simple_animal/hostile/retaliate/bat/crow) || istype(second_target, /mob/living/simple_animal/hostile/retaliate/bat/crow))
+		to_chat(user, span_warning("Zads are immune to mindlinks!"))
+		revert_cast()
+		return FALSE
+
 	user.visible_message(span_notice("[user] touches their temples and concentrates..."), span_notice("I establish a mental connection between [first_target] and [second_target]..."))
 	
 	// Create the mindlink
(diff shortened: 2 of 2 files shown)


Fatherless1Xx:
doesn't effect normal usage
image

Heheniy:
restricting all animals from mind link would make more sence so far

Fatherless1Xx:
> screwing over every single other class that has access to this for the crimes of 1 is pretty awful. Please try and find a fix that JUST effects witches. Their bird form is really powerful as is just block any and all spellcasting while in it? They could simply keep swapping between forms but at least that makes a shit ton of noise

there's no way i could just stop them from landing, transforming, sending a message, then transforming back

benj8560:
well then don't do it? I'm sorry but this is just not reasonable to be nerfing several classes for a issue only around 1 class and likely only a handful of gamers even doing this. Maybe make the sound of transforming even louder so you could hunt them down easily? Make the spell a shout invocation? Idk.  I don't see a good fix myself beyond keeping a mage with counter spell around to stop them.

Fatherless1Xx:
> well then don't do it? I'm sorry but this is just not reasonable to be nerfing several classes for a issue only around 1 class and likely only a handful of gamers even doing this. Maybe make the sound of transforming even louder so you could hunt them down easily? Make the transformation spell a shout invocation? Idk. I don't see a good fix myself beyond keeping a mage with counter spell around to stop them.

there isn't a good option only bad ones
the worst option is allowing people to play as spy drones

Mannybrado:
instead of all of this you should just prevent the use of all spells while in an animal form

coffeecoffeecoffee:
> well then don't do it? I'm sorry but this is just not reasonable to be nerfing several classes for a issue only around 1 class and likely only a handful of gamers even doing this. Maybe make the sound of transforming even louder so you could hunt them down easily? Make the transformation spell a shout invocation? Idk. I don't see a good fix myself beyond keeping a mage with counter spell around to stop them.

it's completely reasonable
this is egregious the way it sounds like it is rn

benj8560:
This also does not really fix the issue entirely. I'm willing to bet if I walked up to the keep and offered to be a rat for them they would **immediately** give me a scomstone for their own friendly UAV. Then one can simply swap in and out on rooftops/trees like you say. 

Some ideas I've been cooking in my head:

Simply nerfing the practice rather then outright preventing it is a lot more ideal I really do think that shout invocation and preventing casting of message in bird form would fix this since at least there's some HEAVY risk in trying to do this. Perhaps a longer cooldown on the transformation spell to.

Alternative options would be just reworking the transformation into something else. Maybe a cat? Black cat ideally but we'd need a sprite. preferably a cat thats much more capable then the druid version given this spell is really the best they have that or offer them some better spells  in exchange for losing borb form. Maybe start with arcane bolt and wither? Or their own custom debuff curse spell? Casting debuffs seems fitting for a witch.

benj8560:
> ...why do we want people who simply took a virtue to be able to have this spell? Am I fucking deranged or aren't virtues supposed to be _you can take a fall_ or _people feel good looking at you?_ A **virtue** can let you transform into a fucking zad at will and shower the town in gore??????

They are talking about the message spell not zad transformation that's witch exclusive

coffeecoffeecoffee:
> > ...why do we want people who simply took a virtue to be able to have this spell? Am I fucking deranged or aren't virtues supposed to be _you can take a fall_ or _people feel good looking at you?_ A **virtue** can let you transform into a fucking zad at will and shower the town in gore??????
> 
> They are talking about the message spell not zad transformation that's witch exclusive

oh yeah that makes a lot more sense now

PR #311, opened 2025-07-23
Arcane affinity reimagined

Fatherless1Xx:
A few things prompted this change. First off, the timer just seems bugged. It says 25 minutes, but even with 16 INT, it felt a lot closer to 50. Something's clearly off and I do not think it's just time dilation.

Second, the way time played scales with spells known feels wrong. So here's the adjustment: instead of learning one spell every 25 (let’s be honest it's really 50) minutes, the acolyte gets to use the miracle once, selects six total spells, and then the miracle vanishes forever.

Why six? Because rounds are three hours long, and most acolytes have high INT anyway. If you factored int in it should probably be seven.
 This can easily be balancejakked by changing 1 (one) number.

image
image

### DIFF
--- a/code/modules/spells/roguetown/acolyte/noc.dm
+++ b/code/modules/spells/roguetown/acolyte/noc.dm
@@ -1,3 +1,5 @@
+var/global/list/arcyne_affinity_spells_remaining_global = list()
+
 // Noc Spells
 // Blindness is a cancerous spells and should not be available to everyone.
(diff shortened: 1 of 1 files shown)


dawsonkeyes:
as far as i'm aware, arcane affinity doesn't give you everything right away on purpose; as it's intended to be comparable to things like zizo's Rituos which gives you one temporary spell until the next cycle
we would be more amenable to a simple rework that concretely makes it work based on sleep cycles (like Rituos)

PR #332, opened 2025-07-24
Even more wretch changes

Fatherless1Xx:
-Vampires and werewolves both get zombie immunity
-Vampires get no sleep trait so they cannot eaisly recover from damage (pay to use regen or find someone to heal you) 
-Vampire disguise now changes all organ colours when disguising and undisguising rather than just your skin tone

Shown by this disgusting sparkledog that should be KOS'd
image
image


ToothyDeer:
wretch vampires cough up blood when they try to drink it from people. is this still an issue

PR #335, opened 2025-07-25
Removal of area specific traits

BadAtThisGame302:
## About The Pull Request

Removes the area specific buffs of Woodsman/Guardsman/Dungeonmaster and Tavern Fighter.

## Why It's Good For The Game

As discussed in coder chat, most agree these buffs are hugely unnecessary and OP.


anconfuzedrock:
yeah the loss of 1 speed and endurance is what really hurts for chasing people. now your stats are a bit worse than other roles but other roles have a lot more customization. perception loss is less important but just _feels_ bad. Skirmisher gets 5 stat points while a condottiero merc gets 7. Same story if we compare the footman with the notoriously mediocre forlorn merc, forlorn gets +1 endurance and +2 con for -1 int. Even the sergeant has less stats than a merc. They still CAN fight but most of the stats they're losing, endurance, speed, perception, are the stats they need to do guard stuff. good luck chasing someone if you're slower and have less endurance. and if you can't chase people, that encourages unfair ambushes as the only way to catch them.

ThinkArbies:
It's not like these roles weren't balanced around having the buff or anything and now, with that classical lack of foresight, they're left twisting in the wind.

The area traits were balanced for what they're intended. Either compensate for a lack of them or leave them be.

roarbark:
im pretty sure this also means that wardens lose longstrider out of town, which was their one upside over bandits...

what's the intended change from this PR? personally i feel this is going to incentivize murderballs on highpop because you can't kill antags otherwise, as well as making guards a lot easier to wipe out on lowpop...

benj8560:
noticed a pretty  significant dip in these roles with this...like uh well:
6XhZJ2hIP1

This is lowpop hours mind you. But 1 in the Garrison is...._wew._ and look at all those Mercenaries! I used to only see maybe 3 at these hours.

benj8560:
what I'm mostly getting at is this removes most of the mechanical incentive to play these roles over adventure or merc. Who are often comparable or with mercs far better to the Garrison with this PR.

chemistrymain3:
Pair it with a reduction for bandits and the removal of their necklace.

PR #347, opened 2025-07-25
Eastern tats are made masc clothes only

Fatherless1Xx:
image

These do not have a female sprite and will thanos snap away your character's boobs


Inveldy:
So you removed the tattoos from female members entirely or just the sprite?

Fatherless1Xx:
> > So you removed the tattoos from female members entirely or just the sprite?
> 
> The former, if I'm looking at the code right. And this kind of sucks since the tattoos are the armor, no?

if you care more about starting armor than having boobs then just use she/her (masc clothing)

coffeecoffeecoffee:
i think we just need a spriter to do this tbh
don't remove things just because they need fixed, would be an incredibly easy thing to sprite

Fatherless1Xx:
> i think we just need a spriter to do this tbh don't remove things just because they need fixed, would be an incredibly easy thing to sprite

awesome, when you sprite it you can make a pr to remove the masc only flag

Cre77:
kind of a flat buff to women now that their armor slot is unlocked and they can actually wear the 300 integrity flak vest from the merchant
i'd rather tattoos go entirely, they are already made redundant with arm markings in chargen

Fatherless1Xx:
> kind of a flat buff to women now that their armor slot is unlocked and they can actually wear the 300 integrity flak vest from the merchant i'd rather tattoos go entirely, they are already made redundant with arm markings in chargen

then go he/him (femme clothes)

Fatherless1Xx:
image

kind of
boobed = true does stop breast flattening but it looks pretty janky, but it's still better than being completely flattened
image

Fatherless1Xx:
never mind, looks like there's already a better solution to this pushed directly to the main branch

PR #350, opened 2025-07-25
An attempt of stat deflation

Fatherless1Xx:
"When everyone's exceptional no one is"

This is an attempt to bring everyone back closer to being average without actually nerfing any roles directly. This PR does three things:

1) Racial stats (apart from golems due to their unique mechanics) are gone.
2) All stat packs apart from virtuous are gone.
3) In place of the old stat packs, are the old racial stats. This means you can get +1 to a useful stat or +2 to less useful stats at most, rather than the massive +6 (ontop of racial stats) we saw from struggler.

image

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/roguetown/golem/golem.dm
+++ b/code/modules/mob/living/carbon/human/species_types/roguetown/golem/golem.dm
@@ -13,7 +13,7 @@
 	conflicted on Golems, for their sensibilities vary wildly from one to the next. \
 	into cohabitation with races they'd deem lesser.<br> \
 	(Insomnia, No hunger, no blood.) \
-	(+1 Constitution, -2 Speed)"
+	(-1 Strength, -2 Speed)"
 
 	construct = 1
 	skin_tone_wording = "Material"
@@ -45,7 +45,7 @@
 		OFFSET_NECK_F = list(0,-1), OFFSET_MOUTH_F = list(0,-1), OFFSET_PANTS_F = list(0,0), \
 		OFFSET_SHIRT_F = list(0,0), OFFSET_ARMOR_F = list(0,0), OFFSET_UNDIES_F = list(0,-1), \
 		)
-	race_bonus = list(STAT_CONSTITUTION = 1, STAT_SPEED = -2)
+	race_bonus = list(STAT_STRENGTH = -1, STAT_SPEED = -2)
 	enflamed_icon = "widefire"
 	organs = list(
 		ORGAN_SLOT_BRAIN = /obj/item/organ/brain/golem,


Fatherless1Xx:
image

ShadowDeath6:
So what would a typical stat spread look like now compared to previously? 
I'm not a _huge_ fan of the available statpacks now- options to trade -1 in one stat for +2 in another stat rather than just being +1 for whichever would be nice (although obviously not as extreme as what we had before where you get +3 to two stats and -2 to two other stats, since the entire point of the PR is to reel back on gigantic amounts of stats).

benj8560:
this seems like the most tasteful way to go about this to me and all around sounds very good! There was definitely a lot of bad nuclear options and this is not such.

anything86:
God please no. This kills build and character concept variety. It makes combat objectively dumber. Statpacks are THE worst thing you could go after if your goal is to 'debloat stats', which isn't even some objectively good metric assuming the stats are there for a reason.

Fatherless1Xx:
> God please no. This kills build and character concept variety. It makes combat objectively dumber. Statpacks are THE worst thing you could go after if your goal is to 'debloat stats', which isn't even some objectively good metric assuming the stats are there for a reason.

there are no builds because everyone hits softcaps

roarbark:
are high stats really an issue? i really enjoyed building characters and picking out statpacks. in no fight have i been "if only our stats had been more average", i genuinely don't see this "if everyone is exceptional, nobody is" in gameplay at all

Fatherless1Xx:
> are high stats really an issue? i really enjoyed building characters and picking out statpacks. in no fight have i been "if only our stats had been more average", i genuinely don't see this "if everyone is exceptional, nobody is" in gameplay at all

they're such an issue that AP put in softcaps for basically everything including swift
We don't have most of those caps (other than strength) but even AP recognizes they're a problem

roarbark:
unless i'm mistaken, the soft caps on AP are literally just STR and swift balance, though? and swift was an issue, even on ratwood (letting any spd build deduct like 60 to 70% of parry) and will be an issue here even after this PR, should you remove softcaps.

the strength softcap... never made much sense to me. people that stat up to 16 to 17 STR, which seems to be the upper limit with builds, lose out 7 to 14% of damage, so maybe 2 to 4 damage points a hit (from the softcapping) and will heavily dump, PER, INT, SPD which makes them fun to counterplay.

DorfGuy1:
> > God please no. This kills build and character concept variety. It makes combat objectively dumber. Statpacks are THE worst thing you could go after if your goal is to 'debloat stats', which isn't even some objectively good metric assuming the stats are there for a reason.
> 
> there are no builds because everyone hits softcaps dracian + struggler is 1 shy of the strength softcap before factoring in bonus stats from roles 12/13 strength should be impressive rather than the baseline, instead we've got dodge experts hitting like barbs

wait, some people just do strength rogue and still get the dodge? How do they not get vibe checked for having low speed. Surely the balance is not so fucked that they let people get ~13s in both speed and str

RyszardParys:
the rot consumes

Fatherless1Xx:
> > > God please no. This kills build and character concept variety. It makes combat objectively dumber. Statpacks are THE worst thing you could go after if your goal is to 'debloat stats', which isn't even some objectively good metric assuming the stats are there for a reason.
> > 
> > 
> > there are no builds because everyone hits softcaps dracian + struggler is 1 shy of the strength softcap before factoring in bonus stats from roles 12/13 strength should be impressive rather than the baseline, instead we've got dodge experts hitting like barbs
> 
> wait, some people just do strength rogue and still get the dodge? How do they not get vibe checked for having low speed. Surely the balance is not so fucked that they let people get ~13s in both speed and str

ignoring the balance argument if there's only one optimal build are you actually "making builds" or are you minmaxxing

Quink-Fix:
...are we seriously going back to old racial stats? Just purge racial stats and leave statpacks. No one wants to be pigeonholed into a certain set of stats that are chosen arbitrarily for picking a particular race, especially when we have races like wildkin and verminfolk which literally _have no specific lore or physical traits._

anything86:
> there are no builds because everyone hits softcaps dracian + struggler is 1 shy of the strength softcap before factoring in bonus stats from roles 12/13 strength should be impressive rather than the baseline, instead we've got dodge experts hitting like barbs

If you bring "softcaps" up again as if it is some universal concept in the code I am shooting you and then myself. The **ONLY** stat softcap we have is for Strength, and **ONLY** on damage.

PR #356, opened 2025-07-26
Yet another proposal

Bobinater:
## About The Pull Request

From the void I bring balancing changes, woe.

## Testing Evidence

image

## Why It's Good For The Game

For a long while now, skirmisher has been the single best ranger class in the game, which is absolutely insane to start. Next, cavalry and the HORRID """stalker""" classes are cut, stalker was literally thief subclass for more or less the security role of the server. Why?

Also adds flavor by making your picked weapon increase the skill of that weapon, otherwise it's cut down on the bloat.


BadAtThisGame302:
I respect the grind.

PR #362, opened 2025-07-27
misc. clothing recipe changes + guard tabard in the loadout

PoltavskaPraca:


BadAtThisGame302:
I don't think giving people the ability to have guard tabards selectable as a loadout item, meaning they can put it on and pretend to be guards without ever seeing one themselves is a good change.

PoltavskaPraca:
> I don't think giving people the ability to have guard tabards selectable as a loadout item, meaning they can put it on and pretend to be guards without ever seeing one themselves is a good change.

You can select every other kind of tabard already, the reason why I want this added in particular - it just looks better on a thin sprite.

PR #363, opened 2025-07-27
Gives Astrata clerics their OG sacred flame back

Fatherless1Xx:
This is in addition to rather than instead of the firebolt 
This now costs more (50) devotion so only acolytes will have enough devotion to be able to use this AND firebolts offensively
image


ShadowDeath6:
tfw the only reason I originally fixed ignition to work is so I could use it as a towner devotee and now I won't even be able to use it u_u

dawsonkeyes:
we are not looking to further bolster this at this time

PR #377, opened 2025-07-28
Let heretics of same Ascendant detect each other, Matthios-style

kimesik:
## About The Pull Request
Lets heretics of same Ascendant deity detect each other, like Matthiosites do.

## Testing Evidence
There's literally nothing that could break. It's the same code that Matthiosites use, but extended to Zizoists, Graggarites and Baothans

## Why It's Good For The Game
Heretics already can find their allies in heart-RMB tab, so I don't know why they can't do the same on Examine. Plus, it makes it easier for heretics to find each other and cooperate inside of the town to actually conspire instead of gathering fragbands in the abandoned church and killing people outside of the town.

### DIFF
--- a/code/modules/mob/living/carbon/human/examine.dm
+++ b/code/modules/mob/living/carbon/human/examine.dm
@@ -870,16 +870,22 @@
 			heretic_text += "A member of Zizo's cabal."
 			if(HAS_TRAIT(examiner, TRAIT_CABAL))
 				heretic_text += " May their ambitions not interfere with mine."
+		else if(HAS_TRAIT(examiner, TRAIT_CABAL))
+			heretic_text += "Fellow scholar of Her Works!"
 	else if((HAS_TRAIT(src, TRAIT_HORDE)))
 		if(seer)
 			heretic_text += "Hardened by Graggar's Rituals."
 			if(HAS_TRAIT(examiner, TRAIT_HORDE))
 				heretic_text += " Mine were a glorious memory."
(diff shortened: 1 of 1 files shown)


kimesik:
> No. This removes any and all potential for subtlety, infiltration, deceit or sabotage when every heretic can just instantly tell each other apart. No one will be able to disguise themselves, nor will Xylixians be able to effectively fake their faith.
> 
> Heretics already have their cheap metaknowledge hideout to gather in. They don't need to be able to pick out complete strangers as instant allies, too.

With the way the game and the map are currently set up, not letting heretics spot their co-believers on Examine (and through heart-RMB tab) only further encourages gathering in the metaknowledge hideout, memorising everyone whom you see there and then making a warband and robbing/kidnapping/fragging any person who steps out of the town's borders.

And as far as I know, Xylixians faking their faith is needed more for flavour and tricking the Inquisition rather than tricking the heretics. Admittedly, I haven't seen a Xylixian pretend to be a heretic, ever; of course, others may have a different experience.

With an easier time finding your allies, you could then get rid of the metaknowledge hideout and actually encourage heretics to meet, send letters to each other, conspire and decide a meeting spot inside the town or somewhere in the wilderness instead of going to the same spot every time. Plus, this would sow some much needed distrust and factionalism amongst heretics; while they can easily find those who believe in the same Ascendant, it's not the case for those who worship a different Ascendant, further hindering creation of heretic fragbands.

PR #381, opened 2025-07-28
Sleep Level-Ups for Wretch Vampires

Cre77:
## About The Pull Request

Adds TRAIT_VAMP_DREAMS to Wretch Vampires, making them receive a moodlet upon daybreak that allows them to trigger the sleep levelup prompt.

## Testing Evidence

image

## Why It's Good For The Game

Enables vamps to do more interesting things than just run around naked and make people on discord cry.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/wretch.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/wretch.dm
@@ -66,7 +66,6 @@
 				ADD_TRAIT(H, TRAIT_NOBREATH, TRAIT_GENERIC)
 				ADD_TRAIT(H, TRAIT_NOPAIN, TRAIT_GENERIC)
 				ADD_TRAIT(H, TRAIT_INFINITE_STAMINA, TRAIT_GENERIC)
-				ADD_TRAIT(H, TRAIT_VAMP_DREAMS, TRAIT_GENERIC)
 				// Apply -1 to all stats
 				H.change_stat("strength", -1)
 				H.change_stat("perception", -1)
--- a/code/modules/jobs/job_types/roguetown/adventurer/wretch.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/wretch.dm
@@ -66,6 +66,7 @@
 				ADD_TRAIT(H, TRAIT_NOBREATH, TRAIT_GENERIC)
 				ADD_TRAIT(H, TRAIT_NOPAIN, TRAIT_GENERIC)
 				ADD_TRAIT(H, TRAIT_INFINITE_STAMINA, TRAIT_GENERIC)
+				ADD_TRAIT(H, TRAIT_VAMP_DREAMS, TRAIT_GENERIC)
 				// Apply -1 to all stats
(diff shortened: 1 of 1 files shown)


ThinkArbies:
When sleep skill-ups were first introduced on Ratwood, making sure vampires couldn't sleep to benefit from it was a _major_ part of balancing out vampires. They were individually strong on their own, it was reasoned, that they shouldn't be allowed to potentially endlessly and effortlessly grind for more skills. It was their incentive not to frag out too hard in the event that there was something they actually needed from someone. Like a blacksmith, for example.

This just enables an already powerful class to become even more powerful, which they'll abuse rather than do anything remotely interesting like the original intent.

john-pawson:
> When sleep skill-ups were first introduced on Ratwood, making sure vampires couldn't sleep to benefit from it was a _major_ part of balancing out vampires. They were individually strong on their own, it was reasoned, that they shouldn't be allowed to potentially endlessly and effortlessly grind for more skills and have been balanced around that fact.

I should point out that this is false, by the way. Ratwood's vampires have been able to sleep and gain points on since the sleeping was merged, both of them are even introduced on the same PR, and the trait wasn't removed since. 

https://github.com/Rotwood-Vale/Ratwood-Keep/pull/498 <-- The PR itself

Not only that, but the current VL and spawns also have the trait on their code, so even current vampires, which are the ones that should be converting and the such to benefit from the skills of a new spawn, can still sleep and gain points.

Alright after investigating a bit, since I was curious about this, it seems that there's two vampire datums. Vampire lord (Used by ratwood, which added the sleep PR, and the new trait. Also used for the regular antag) and Vampire (Which is unused in Ratwood, and only used here used for wretches recently) so it was likely simply never added to this datum since it was... never used. 

You can even see how it's traits haven't been touched since the initial commit, because there was no need to touch them, and how they don't use the inherent_traits list like the other more updated VL datum either, and it directly adds traits on the wretch subclass. The datum is also ignored on such things as fyritius flowers being dipped (It only checks for VL spawns or werewolves, a wretch vampyre wouldn't be affected), and they don't even interact with silver poultices. Them getting burnt by silver was only added recently on this PR: https://github.com/Scarlet-Reach/Scarlet-Reach/pull/277 and they still don't get affected by things such as picking up silver items like goblets or silver necklaces even while not disguised.

I don't know if it was used on AP or how much it was used, I don't know if there's a specific omission of it on the port of the vampyre sleeping trait that I'm not aware of, but I am sure that vampires weren't balanced around the fact that they couldn't sleep, since they have been able to sleep since sleeping was added, and this is just a case of vampire wretches using an outdated datum of vampires.

I would like to clarify, I don't mean to say that I agree with the PR, I just felt the need to clarify that no, vampires weren't designed to be unable to sleep, and this is likely an oversight (80% sure, I'd be glad to be proven wrong) (I *wish* I could find the PR that adds vamp wretches originally to see if there was any discussion over it but the port doesn't mention it and I can't find it anywhere else on the other repositories ;-;)

Gehenna-Host:
Them not leveling from sleep is intentional. Gonna close this.

PR #382, opened 2025-07-28
Pyromancer adventurer class

Fatherless1Xx:
This is a hybrid mage/cleric.
Capped at t2 and forced Astrata, so they always get ignite and sacred flame.
They get no spell points or mage armor, but they get fireball and spit fire assigned directly.
image


Fatherless1Xx:
pyromancers have a devotion limit of 499 so they can use their miracles in combat but cannot unlock revive:
image

Genevesh:
Not a big fan of this. Adventurers are already unreasonably powerful foreigners, making a hybrid like this is just.. EH..
It'd be nicer to turn this into some kind of a combat churchman role, maybe a subclass of templar?

PR #383, opened 2025-07-28
Acid and Frost bolt speedup jak

saltprjak:
## About The Pull Request

Makes Acid Splash and Frost Bolt projectiles move with the same speed as any other wizard projectile.

## Testing Evidence

it just works

## Why It's Good For The Game

Currently THE most underused offensive spells due to their abysmally slow speeds which, coupled with the comparatively generous charge time, make them easily dodged by anyone who isn't likely already leg-checking you with a zweihander. This gets even worse on highpop, to the point you're actively robbing yourself of damage and CC by trying to land it instead of spamming a rotation of BoL -> Fireball -> Spitfire -> Arcyne Bolt.

Frost Bolt would work just fine in terms of cost-to-value thanks to the debuff, but acid splash might need to either keep a (much less oppressive) slowdown or have a higher cost - it's 100 burn (spread over 20 seconds; keep in mind everyone is tanky as shit). I'd need feedback and opinions.


Genevesh:
Making acid splash a 4-cost spell would be pretty neat, honestly, since that would be the only of the three 4-costs, and one of a kind offensive spell.

Mithrun:
Yeah this is great, both of those spells are way too slow rn....

PR #384, opened 2025-07-28
elf/french accent

Fatherless1Xx:
hon hon hon
image


ayasanasa:
There is already Otavan accent which basically covers this, would be nice if normal elves got some accent instead.

PR #387, opened 2025-07-29
Impreg Rework: Add impregnation probability handling to sex actions and genital organs.

darklizard45:
## About The Pull Request

The idea is to change how the game handles the impregnation proc to be more interactive and cumulative so players can have a bit more interaction with the probabilities to trigger it.

## Testing Evidence

I tested the changes by running a private server on Dream Daemon, granted admin on myself by moving the files from config/example to config and labeling my ckey to the required rank, I instead of using the "roguetest.dm" I waited 5 minutes and ran the standard map as I was still figuring out how this works, managed to get the Server working and spawning myself as a "Male" character stripped it and went back to lobby to respawn as a "Female" both characters with the corresponding organs to properly test it, noticed I forgot to Toggle the ERP Panel Menu, toggled it ON, did the mechanical ERP until I trigger the corresponding Proc making sure it works.
And here are the server logs on a .txt of the test.
[Debug - censured.txt](https://github.com/user-attachments/files/21478864/Debug.-.censured.txt)

Edit: I changed the file contents to be more secured with my info.

## Why It's Good For The Game

It gives more risk to the action and makes players think more carefully about the chances as it scales with every attempt.

### DIFF
--- a/code/modules/surgery/organs/feature_organs/genitals.dm
+++ b/code/modules/surgery/organs/feature_organs/genitals.dm
@@ -100,14 +100,16 @@
 	var/impregnation_probability = IMPREG_PROB_DEFAULT
 
 /obj/item/organ/vagina/proc/be_impregnated(mob/living/carbon/human/father)
-	if(pregnant)
-		return
-	if(!owner)
-		return
-	if(owner.stat == DEAD)
-		return
-	to_chat(owner, span_love("I feel a surge of warmth in my belly, I’m definitely pregnant!"))
-	pregnant = TRUE
+    if(!owner)
+        return
+    if(owner.stat == DEAD)
+        return
+    if(pregnant)
+        to_chat(owner, span_love("I feel a surge of warmth in my belly again..."))
+        return
+    to_chat(owner, span_love("I feel a surge of warmth in my belly, I’m definitely pregnant!"))
+    pregnant = TRUE
+	//TODO add a way to trigger lactating when pregnancy happens
 
 /obj/item/organ/breasts
 	name = "breasts"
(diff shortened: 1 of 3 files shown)


BadAtThisGame302:
I think it'd be even better if you added some sort of examine text on a pregnant person that said something about their belly. That will give their actions definitely more risk.

darklizard45:
> > I think it'd be even better if you added some sort of examine text on a pregnant person that said something about their belly. That will give their actions definitely more risk.
> 
> I think you'd end up with an edgelord problem. Maybe put that on a timer.

Yes, for this reason alone I won't be adding any descriptions on examine and is left up to the player how they want to handle it as an IC risk. I'm considering linking it to the lactation mechanic as it increases the hunger rate, it should work as a consequence enough.
In the past it used to cause nausea but it got removed for being too annoying.

darklizard45:
The following document is an addendum of how I summarize how I'm trying to make the percentages work, so far I have encountered an issue when dictating the "Virility" values.

[Percentages.pdf](https://github.com/user-attachments/files/21497647/Percentages.pdf)

As such I'm going to split the code in two parts, the first one will be what I have done so far apart apart from a few tweaks if necessary, the second will be trying to get a clear idea of what values would be the best to use and seeing how to implement them.

code/modules/surgery/organs/feature_organs/genitals.dm
darklizard45, line 1: Tested it recently to showcase what it does:
It gives a message to the player that the condition of "pregnant" has been set to TRUE
Test-A

Any further attempts in which the proc has been triggered result in the following message:
Test-B
Test-C

Originally used for debugging but felt it could be good to give players an extra bit of flair.

DorfGuy1:
i love that you made a document but for the love of god fuck pdfs.

EvenInDeathIStillServe:
Thank you for your service.

PR #393, opened 2025-07-29
Buffs werewolf claw

Fatherless1Xx:
With infection code gone they've had to adapt as lone wolves 
Old wolves were definitely on the weaker side imho, their defense/force/AP is really low, this buffs it.
These lonewolves are not ones you should want to 1v1


Genevesh:
This is a must-have with the infections gone

PR #396, opened 2025-07-29
Big hairrport (fixed)

Fatherless1Xx:
Adds all the new hairstyles without removing any of the hairstyles we have that AP does not
image


Fatherless1Xx:
there's a conflict because my branch is outdated, durr, easier to resubmit than fix this branch

PR #400, opened 2025-07-30
LETS GO GAMBLING - Fortune+buff based slot machines

K4rlox:
## About The Pull Request

Makes the xylix's fortune calculate depending on the user's fortune AND BUFFS
Current default is 10 luck starting at 50%
increasing by 5% every point of luck
30 luck sets it to 100% which i maxcapped at 95%
anything under 10 adds a -5% penalty to win chance

## Testing Evidence

https://github.com/user-attachments/assets/6caca47f-acc4-4120-b07b-1bd164239f47

## Why It's Good For The Game

Gives the luck stat buffs an actual use, Gives people a reason to seek ways to boost their luck to gamble and also makes the slots machines more fun
However this also slightly decreases the default chance to win


DelineFortune:
it allows people to literally have unlimited money from the air

PR #402, opened 2025-07-30
Werewolves passively regenerate natural armor  and health

Fatherless1Xx:
They are too angry to die
Also sends them a message each time they heal to remind them that they are angry murder beasts and not big fluffy puppies
This happens every 3 minutes so they can run away to heal but it's not likely to help midcombat
No runtimes relating to werewolves:
image

### DIFF
--- a/code/modules/antagonists/roguetown/villain/werewolf/werewolf.dm
+++ b/code/modules/antagonists/roguetown/villain/werewolf/werewolf.dm
@@ -18,6 +18,7 @@
 	var/wwbreastsize = 3
 	var/wolfname = "Verewolf"
 	var/wretch_antag = FALSE
+	var/list/mob_timers = list()
 
 /datum/antagonist/werewolf/lesser
 	name = "Lesser Verewolf"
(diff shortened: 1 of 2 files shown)


coffeecoffeecoffee:
Wait, does it play these messages even when the WW is at full health? Just whenever it procs healing, or whenever it actually heals?

Genevesh:
A lot of people seem to forget that they're meant to be feral as WWs. I think this is a good reminder, and doing it with a 3 minute gap between messages makes sure it's not THAT bad.

I think you could lowkey track how long the night is with these too, since they're on a fixed timer - I think every night is 30 minutes? So by the time there is a 10th message, you know you're about to turn back.

Overall a GREAT buff to WWs, since combat here relies heavily around armor use, and WWs' biggest bane is being crippled. I'm really happy to see this.

dawsonkeyes:
please state why this is good for the game in your pr

dawsonkeyes:
werewolves as a wretch-available antagonist aren't something we're looking to make any stronger at this time

PR #403, opened 2025-07-30
[TM ONLY] The scom costs copper to use and is delayed

Fatherless1Xx:
image

Each scom message now costs one (1) copper.
There is a delay of 30 seconds from you sending the message to it being broadcast. 
This is so an antag can murder you if you snitch without the entire town descending on your location.

You ready a scom for use by inserting coin:
image
image

This is a very big change and there may be a few bugs I did not catch in testing, please TM.

### DIFF
--- a/code/modules/roguetown/roguemachine/scomm.dm
+++ b/code/modules/roguetown/roguemachine/scomm.dm
@@ -443,10 +443,8 @@
 		to_chat(user, span_warning("I need to insert a copper coin into the scomstone first!"))
 		return
 	
-	// Check if user has noble trait for free sending
-	if(!HAS_TRAIT(user, TRAIT_NOBLE))
-		// Remove the coin since it was used
-		ready_to_send = FALSE
+	// Remove the coin since it was used
+	ready_to_send = FALSE
 	
 	var/usedcolor = user.voice_color
 	if(user.voicecolor_override)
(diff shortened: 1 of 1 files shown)


DelineFortune:
Finally antags going to have a chance to do something before NORTH GATES BANDITS

also keep in mind you can just use scom ring, right click at it and have pre-typed words ready like

be a steward

right click at your scom

pre-type STEWARDY!! BANDITS!!

be in safety

Fatherless1Xx:
> Finally antags going to have a chance to do something before NORTH GATES BANDITS
> 
> also keep in mind you can just use scom ring, right click at it and have pre-typed words ready like
> 
> be a steward
> 
> right click at your scom ring
> 
> pre-type STEWARDY!! BANDITS!!
> 
> be in safety

I plan to do more with this later like letting the duke change the cost from the throne
But right now this is just the foundation to build from if this gets merged

awawaawawaawa:
I'm not sure why people are down-voting this PR. 
I think it's a good edition.

Inveldy:
I like this, but I think you should be able to put in a silver coin for a reduced delay, or a gold coin for no delay, as a sort of "Priority" function.

I do think there are solutions that could work a bit better but these are a decent band-aid.

coffeecoffeecoffee:
> imo it's a very rough solution to a very real problem.
> 
> I think a more elegant way to go about people instantly calling for help is making it possible to break, booby-trap and repair stationary SCOMs, while introducing delay only to portable SCOMstones. It'd encourage Wretches and Bandits to do prep-work before their raids.
> 
> Not voting on this one because it's better to have it than change nothing at all, but it's not ideal.

this is basically why I downvoted it; excellent way to put it, a really awkward solution to a real problem.

dawsonkeyes:
please state why this is good for the game in your pr (i agree with the core intent; but please do follow the format)

PR #412, opened 2025-07-31
MAA Balancejak - Guardsman trait removal + reimbursement

AbsorbentHeart5:
## About The Pull Request

Removes the Guardsman trait from MAAs & Sergeant, instead reimbursing these roles with +2/3 stats to bring them to a total of +7, whereas before it would be +5 or +10 WITH the Guard buff. Sergeant was given a total of +8, which should be fine considering it's a 1 slot role.

Other garrison roles are left untouched.

## Testing Evidence

image

## Why It's Good For The Game

I've been told that the Town Buffs weren't exactly liked, however the consensus was that people didn't want the trait entirely removed without taking into account how the stats removal would affect guards going up against bandits and wretches. This PR will hopefully be a better alternative to #335 as it brings MAA up to be similar to mercs, but still not on par with both bandits and wretches, and other special limited roles.

Knights and Knight Captain were left untouched as they already have a good stat line up, and leaving the keep should be heavily discouraged. Wardens might be touched upon in a later PR.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/manorguard.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/manorguard.dm
@@ -79,13 +79,12 @@
 	H.adjust_skillrank(/datum/skill/misc/riding, 1, TRUE)
 	H.adjust_skillrank(/datum/skill/misc/tracking, 1, TRUE)
 	ADD_TRAIT(H, TRAIT_MEDIUMARMOR, TRAIT_GENERIC)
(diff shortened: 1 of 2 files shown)


DelineFortune:
whats point, to let keep players charge into the forest/bog/behind north gates for free and antag hunt?

Teron-123:
Why is the area buff disliked? I think this is a good way to balance things, keep should be careful when venturing out of their home turf to hunt antags.

roarbark:
there isn't really a good argument to remove the area buffs. they add depth to play, making the choice of letting antags leave your buff area or risking a fight on less stats.

PR #415, opened 2025-07-31
Increases all werewolf stats to 20 and gives dwielder

Fatherless1Xx:
They were previously nerfed because of how quickly they snowballed with infections
With infection code likely to go, they should be buffed again 
This is so they're notably stronger than heretics in zizite armor who do not have daytime/silver vulnerabilities and do not lose stats (not being a werewolf/vampire is a +10 in all stats)


roarbark:
i suggest just giving them +3 spd and +3 per instead on top of what they have...

Kregle:
Absolutely not.

PR #419, opened 2025-08-01
Peel can only apply stacks if armor inegrity is below 50%

Fatherless1Xx:
I have really mixed feelings about peel. I like that it's a counter to heavy armor, that’s important to have. But it’s too effective imho. You shouldn’t be able to completely ruin high-end armor like darksteel in just a few atatcks. It makes heavy armor feel pointless. A better compromise would be requiring the armor to take a certain amount of damage first before peel can be attempted.

image
image

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -1361,6 +1361,13 @@ GLOBAL_VAR_INIT(rpg_loot_items, FALSE)
 /obj/item/proc/peel_coverage(bodypart, divisor)
 	var/coveragezone = attackzone2coveragezone(bodypart)
 	if(!(body_parts_inherent & coveragezone))
+		// Check if armor integrity is below 50% before allowing peel stacks
+		if(max_integrity != 0)
+			var/int_percent = round(((obj_integrity / max_integrity) * 100), 1)
+			if(int_percent >= 50)
+				visible_message(span_info("The armor is too intact to peel effectively!"))
+				return
+		
 		if(!last_peeled_limb || coveragezone == last_peeled_limb)
 			if(divisor >= peel_threshold)
 				peel_count += divisor ? (peel_threshold / divisor ) : 1


Fatherless1Xx:
> unironically a skill issue

I know some will dismiss this as "just use bait idiot" but that’s not a realistic counter when mask code and group fights come into play. Good luck trying to figure out which of the six unknowns is currently trying to peel your chassues off. Even without masks, keeping track of the scroll is a challenge on its own.

This would be fine if every fight were just 1v1 and you could see who you were fighting, but you can't.

benj8560:
peel is BUSTED as hell I debated pushing to just have it removed. I have just seen people so effortlessly melted by it. This is a better non-nuclear option.

pyrzal:
> > unironically a skill issue
> 
> I know some will dismiss this as "just use bait idiot" but that’s not a realistic counter when mask code and group fights come into play. Good luck trying to figure out which of the six unknowns is currently trying to peel your chassues off. Even without masks, keeping track of the scroll is a challenge on its own.
> 
> This would be fine if every fight were just 1v1 and you could see who you were fighting, but that isn't the case.

just use defend idiot

Fatherless1Xx:
> > > unironically a skill issue
> > 
> > 
> > I know some will dismiss this as "just use bait idiot" but that’s not a realistic counter when mask code and group fights come into play. Good luck trying to figure out which of the six unknowns is currently trying to peel your chassues off. Even without masks, keeping track of the scroll is a challenge on its own.
> > This would be fine if every fight were just 1v1 and you could see who you were fighting, but that isn't the case.
> 
> just use defend idiot

This isn't a counter to peeling, they're still destroying your 700 integrity Armor in 3 (three) hits. 
There isn't enough visual feedback for peel to realistically be counterable for any fight that isn't a 1v1.

LostLittleMeowMeow:
peel is already terrible against actual good players. defending and baiting doesn't just counter it and fuck over the person peeling, but even **_regenerates peel stacks_**. peel has so many counters that it only works against dumbasses (which most platebeasts fall under.) fucking with this without addressing blunt weapons or other stuff being weak as hell has terrible knight buff implications. if you want heretic armor buffed specifically against gank-groups (ignoring the issue of 6v1 gank pvp being impossible to balance to begin with), increase the peel threshold on those pieces of armor instead of gutting the entire system. some armors already take more hits to peel than others.

but then you've got the issue of those dudes being unpeelable immortal gods in non-group fights.

zeratino:
actual skill issue

ChillyLobster:
Hooooooly saltPR

anconfuzedrock:
I think peel is fine, but I'd rather peel be removed than this. You have no way to tell what level of damage their armor is at, so if you get it to 50% damage there's no reason not to just keep swinging with regular cuts. This leaves peel as nothing more than a trap for new players and an intent you have to avoid accidentally picking.

dawsonkeyes:
this is unintuitive and not something we're looking to add to the game

if something like peel is as much of a problem as it's stated, something that's been suggested above (like removing peel entirely and buffing certain blunt options, which plate is meant to be weak to) is more likely to get merged

PR #421, opened 2025-08-01
Socialite gives you every language

Fatherless1Xx:
image
image


PoltavskaPraca:
Wherefore?

dawsonkeyes:
socialite is already just beautiful but with the extra benefit of empath; and intellectual only gets a pick of 3 languages
there is no reason for this trait to get made better; there is no change to this pr that would get it merged

PR #426, opened 2025-08-02
a fix

Fatherless1Xx:
.
image


benj8560:
oh you noticed this bug yeah. Please..others have pointed it out. Explain your PRs just a TINY bit better god.

For context you could spam jump while grabbed to roll resisting the grab like 50 times back to back. with 0 cooldown.  pretty much guarantee an escape unless your exhausted regardless of stats.

EvenInDeathIStillServe (approved):

PR #427, opened 2025-08-02
Throwing coins creates an admin alert

Fatherless1Xx:
image

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -246,6 +246,13 @@
 
 
 	if(thrown_thing)
+		// Admin alert for coin throws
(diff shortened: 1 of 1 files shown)


coffeecoffeecoffee:
...why?

sippykot:
i assume so admemes can punish people doing it to lag the server? 

but imo the better solution would be to just... not have coin stacks spill when you throw them
for gameplay reasons

coffeecoffeecoffee:
> i assume so admemes can punish people doing it to lag the server?
> 
> but imo the better solution would be to just... not have coin stacks spill when you throw them for gameplay reasons

nah i think them spilling is pretty soul
i just didn't realize it was a source of lag when people do it over and over
just wanted to know; good change

PR #428, opened 2025-08-02
VerminFolk get same roles as WildKin

RiceGateau:
## About The Pull Request

Seen someone make some fair points so I pushed the PR, tested it so it shouldn't blow up the server. I edited the roguetown.dm file in _DEFINES to allow for VerminFolk to receive the same roles as their larger kin.

## Testing Evidence

Test 1
Test 2
Test 3

## Why It's Good For The Game

"Wildkin existing already makes the fact that vermin aren't tolerated in those roles very silly. We already have rat people in the garrison, we have rat people as priests, hell some of the zardmen have been even more ridiculous than some of the rats. As long as wildkin exists, they will be better rats than normal rats if you want to be multiple jobs. Also, verminvolk doesn't even necessarily mean rat people. It can be gnomes, bunnies like karmen, and other non-rat races. We had them as councilors and such before with little issue"

### DIFF
--- a/code/__DEFINES/roguetown.dm
+++ b/code/__DEFINES/roguetown.dm
@@ -146,6 +146,7 @@
     /datum/species/lupian,\
     /datum/species/moth,\
     /datum/species/anthromorph,\
+    /datum/species/anthromorphsmall,\
     /datum/species/tabaxi,\
     /datum/species/lizardfolk,\
     /datum/species/dracon,\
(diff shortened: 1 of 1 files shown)


coffeecoffeecoffee:
vermin are clearly intended and characterized as being more untrustworthy, scheming and uncivilized than regular wildkin, which is clearly borne out ingame by the players

no

play vermin because you want to play vermin, not because you want "the most access to roles" like it's some kind of racial powergaming simulator

SFox63:
This is progress. We need this.

Fatherless1Xx:
Wildkin but short

PR #440, opened 2025-08-03
Buffs deserter orders

Fatherless1Xx:
## About The Pull Request

-Can now be self cast
-Recruitment as a mechanic is gone, they can cast use their orders on anybody without needing to recruit them first

## Testing Evidence

image

## Why It's Good For The Game

This is good for the game because they're already worse clerics/heretics so they don't need all these additional weaknesses
Recruitment is cool on paper but in practice it's why we don't ever see anybody play this class


dawsonkeyes:
deserter is worse cast-wise than heretic because they get expert in every weapon skill and halfplate roundstart; their orders are intended to make them a team player and we don't want them to be able to just buff themselves freely

PR #447, opened 2025-08-04
Empowers Baotha rituals but gives them all shout invokes

Fatherless1Xx:
.

## About The Pull Request

Baotha’s miracles have long offered little more than flavor. This proposed revision serves two purposes: First, to make the miracle set meaningful, given that Baothan clerics are valids unlike Ten clerics. Second, to prevent them from hiding too easily.

The changes significantly enhance the impact of Baotha’s miracles, but balance this power by requiring them to be invoked aloud, as shouts. 

-Baotha's blessing now heals fatigue and nourishes you. Nothing too crazy.
-Enrapturing powder applies emberwine and drunk walking for a few seconds.
-Painkiller is where it gets wacky. No fatigue, no death and no pain is applied for a minute. This means the only way you can die is with a beheading. To be subdued you have to be crit into submission.

## Testing Evidence

image
image
image
image

## Why It's Good For The Game

Baotha's a cool concept. It’d be awesome to see more solid representation for her, something that actually holds up mechanically that isn't just flavor.

### DIFF
--- a/code/datums/status_effects/rogue/roguebuff.dm
+++ b/code/datums/status_effects/rogue/roguebuff.dm
@@ -208,6 +208,37 @@
 	owner.cmode_music = originalcmode
 	. = ..()
 
+/datum/status_effect/buff/divine_protection
+	id = "divine_protection"
+	alert_type = /atom/movable/screen/alert/status_effect/buff/divine_protection
+	duration = 1 MINUTES
+
+/datum/status_effect/buff/divine_protection/on_apply()
+	. = ..()
+	ADD_TRAIT(owner, TRAIT_BLOODLOSS_IMMUNE, TRAIT_MIRACLE)
+	ADD_TRAIT(owner, TRAIT_STUNIMMUNE, TRAIT_MIRACLE)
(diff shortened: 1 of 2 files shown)


benj8560:
I am strongly against Emberwine for this. The OD is permanent. It is a _sex_ drug so in order to properly follow rules someone would need to check non-con prefs with EVERY person they cast on and I assure you no one will. Don't mix combat and ERP mechanically. 

Enrapturing powder already could kill people in fullplate with only 3 hits via OD. It was pretty good.

Fatherless1Xx:
> Baothans have two things going for them, mechanically, because their miracles suck. Firstly, the healing miracle is outright the best in the game. It's Necra's heal but objectively better, at 50% health (for the bonus) as opposed to 25% health. Secondly, the ability to pop as many drugs as they want to keep a buff indefinitely, which makes Baothan bandits an actual chore to deal with.
> 
> While the miracles could do with changes, this ain't it. It's a wild swing in the opposite direction and isn't great. In my opinion, at least.

The difference is Baothans are antagonists that you are justified in killing upon discovery, they should be better objectively better than Ten clerics.

PR #449, opened 2025-08-04
Passive cinnabar generation into stockpile

DelineFortune:
## About The Pull Request

Adds slowly passive cinnabar regeneration into stockpile

## Testing Evidence

image

## Why It's Good For The Game

Its low chance to get one from mining but very vital thing for mages


Clocks-scp:
Really? What is the chance to get cinnabar supposed to be right now? Because I haven't been getting any even with fortune 20.

PR #450, opened 2025-08-04
Mercenary balacjak

Dorodomki:
## About The Pull Request

rebalances warrior mercs to fit the same armor/cover of the zybantu mercenary (probably the strongest one alongside atgervi footman)

gives the mercenaries a minimun of armor for the vital parts of a combat role (leather coif to prevent a critical kill aka 1 tap)

gives the mercenaries the perk of Steel heart since they are combat roles with very experienced backgrounds shouldn't get nervous of blood and dismembering on a fight reducing the stress they get from fight to fight

updates the stats of the mercenaries & some of their subclasses to be a flat 7 instead of be lacking, doesn't touch the 10 statpoint mercs since they already include -stats to compensate for those, somewhat leaving them on 7 statpoints

## Testing Evidence

image

here we can see a unarmored merc with the basics for neck protection, same as a brigand npc, it prevents a critical hit on classes that should be wearing that armor since they exist for that, weaker gear than court roles & comparable to adventurers/wretches

dreamseeker_56bhdIoKsK

here we got a redesigned kozak who used to spawn with two whips, no head armor and no "skirmish gear" like his title says, now he has a coif for head/neck protection, flail for a variety on weapons and javelins for a proper skirmish gear 

## Why It's Good For The Game

a player on scarlet reach placed a bounty on he rebalance of the mercenaries stats & skills to be fair, the pr intends to make it real 

makes the weakest mercs slighly more armored (in leather) and MAINLY covered, it means they will still take damage yet they won't be one shot by anyone with 15 str + spear roundstart and die since they don't spawn with coif/helmet while adventurer & keep "counterparts" do

### DIFF
--- a/code/modules/jobs/job_types/roguetown/mercenaries/classes/warscholar.dm
+++ b/code/modules/jobs/job_types/roguetown/mercenaries/classes/warscholar.dm
@@ -98,7 +98,7 @@
 			H.adjust_skillrank(/datum/skill/misc/stealing, 3, TRUE)
 			H.adjust_skillrank(/datum/skill/misc/lockpicking, 3, TRUE)
 			H.change_stat("strength", 3)
-			//H.change_stat("constitution", -1) //doesn't seem clear the idea of reduce some cqc focused role constitution compared to the atgervi
+			H.change_stat("constitution", 2) //doesn't seem clear the idea of reduce some cqc focused role constitution compared to the atgervi
 			H.change_stat("endurance", 1)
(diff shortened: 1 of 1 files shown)


roarbark:
im pretty sure the idea is that mercs spawn with vital gear missing so they take jobs in town early so they can get proper armour and cant just go out fragging from day 1. i'd rather see the strongest mercs having huge gaps in their starting armour as a small tradeoff for their statlines, and something they want to set out and fix

PR #454, opened 2025-08-05
removes "good -> bon" in otavan accent

Mannybrado:
## About The Pull Request
title

## Testing Evidence
compiles

## Why It's Good For The Game
its annoying and the only reason why i know bon means good in french is because of this game, its a bad word replacement

### DIFF
--- a/strings/french_replacement.json
+++ b/strings/french_replacement.json
@@ -51,7 +51,6 @@
 		"enter": "entrer",
 		"for": "pour",
 		"shit": "merde",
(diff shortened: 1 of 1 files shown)


RiceGateau:
not bon

carbous:
so here's the main issue with "bon" in french there's two 'good' bon and bonne, one is masculine and one is feminine.
So "good luck" in french is "bonne chance" or "bonne luck". It's the kind of word you can't automatically have translated into english like that.

PR #459, opened 2025-08-05
RETURN OF THE VOLF PIT but not really unless you build it - adds volfs, bears and horses as expensive import items

coffeecoffeecoffee:
## About The Pull Request
Adds horses, direbears, volfs, and something special to the imports list.
Horses are just reskinned saigas without the slight delay on sprinting. Pretty much just a cute prestige mount.
All of the prices have been designed to be fairly high to prevent mass ordering them. If these are too low by any means, let me know. They're meant to be luxury things for fun.
Volfs - 300
Horse - 500
Direbear - 1000
By the way, there's some kind of glitch where injured mounts don't return to their previous speed once fully healed. It's not just horses. No, I don't know how to fix it.

## Why It's Good For The Game
i asked santa for a pony one time and never got over not getting one
Also dukes should be able to make cool volf pits if they want them. And have an attack bear. And have it go terribly wrong.
These aren't all that dangerous mobs and if you have 1000s and 1000s of mammon, you should be enabled to spend it irresponsibly.

### DIFF
--- a/code/modules/roguetown/roguestock/import.dm
+++ b/code/modules/roguetown/roguestock/import.dm
@@ -170,6 +170,64 @@
 	. = ..()
 	new /mob/living/simple_animal/hostile/retaliate/rogue/saiga/saigabuck/tame/saddled(src)
 
+/datum/roguestock/import/volfcrate
+	name = "Volf Crate"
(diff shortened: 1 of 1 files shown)


dawsonkeyes:
these ponies look awfully just like two saigas stitched together

PR #467, opened 2025-08-05
Hedge mage arcyne dagger

Gehenna-Host:
## About The Pull Request

Makes it so that hedge mages spawn with cinnabar and a silver dagger to play around with the mage loop.

## Testing Evidence

Screenshot_3

## Why It's Good For The Game

Hedge mages will be able to make runes and summon evil creatures without being forced to visit the town for cinnabar and get jumped by the keep.


dawsonkeyes:
hedge mages are already in a very good spot and getting the items to do this should be part of the difficulty with playing a wretch

PR #474, opened 2025-08-06
Grants grab immunity for a few seconds after breaking out of a grab

roarbark:
## About The Pull Request

HUGE thanks to azarak for most of the code

-> successfully resisting a grab will grant a baseline immunity window of 5 seconds where any grab made on you will be voided as if parried
-> the ATTACKER'S wrestling skill lowers this cooldown by 0.4 seconds per level

## Testing Evidence

!!MODIFIED VALUES IN THE VIDEO!!

here, the -0.4 second deduction on the immunity window was raised to 2 seconds to show the difference

the ranger has wrestle 1, the monk has wrestle 4. the monk grabbing the ranger grants no immunity on modified values and the monk breaking the ranger's grab gets a 3 second immunity as expected. i tested this with the actual values and it felt right, as well.

i also added the grabbreak.ogg noise on slips for player feedback

https://github.com/user-attachments/assets/62a7dd1f-3286-4bbf-9db4-93b9719dcbb8

## Why It's Good For The Game
-> wrestling is disproportionally stronger in 1vX, while this is not a bad thing in itself, getting chaingrabbed and not being able to do anything because you failed to metabuddy up sucks and isn't much fun. while getting grabbed in a 1v2 situation is still bad for you, you can still struggle and defend yourself within your immunity windows

-> wrestling in 1v1 can add more depth to a fight, but is bothersome and unfun when someone just spams it and does nothing else... now you can still have wrestle attempts, but you'll have to fight "normally" inbetween them

### DIFF
--- a/code/__DEFINES/cooldowns.dm
+++ b/code/__DEFINES/cooldowns.dm
@@ -79,3 +79,12 @@
 #define COOLDOWN_STARTED(cd_source, cd_index) (cd_source.cd_index != 0)
 
 #define COOLDOWN_TIMELEFT(cd_source, cd_index) (max(0, cd_source.cd_index - world.time))
+
+//simpler world.time + timespan method
+#define ANY_MOB_TIMER(mob, timer) (mob.mob_timers[timer])
+
+#define MOB_TIMER_CHECK(mob, timer) (mob.mob_timers[timer] && mob.mob_timers[timer] > world.time)
+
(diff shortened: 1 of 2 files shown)


dawsonkeyes:
if you are in a 1vX situation you should be losing to being dogpile-grabbed; this is not the way to go about addressing issues with not having comrades to help you in a fight
1v1 wrestling is in a fine spot right now as far as i'm concerned

PR #476, opened 2025-08-06
Kobolds Eating Gems/Stones

ProbablyCarl:
## About The Pull Request
As it says for the title. Kobolds can now eat rocks and gems. Someone had posed an interesting question ICly to a character of mine, and it led to some fun RP, so I figured 'why not'. Right? I 'unno.

Regardless, this is how it works:
 - Stones provide (very little) nutritional value, alongside wound healing and blood regeneration. Those three are based on stone level. A ten second buff is applied when you eat a stone to give the healing and blood regeneration.
 - Gems provide a full spectrum heal, excluding blood, based on sell price. The better the gem, the better the heal. This includes riddles. A four second burst heal is applied upon eating any gem. Riddles give a serious stress event, while also giving a text alert to nearby players, too.
 - Riddles are somewhat different in the eating process, in that there's an inherent 50/50 chance on all riddles to horribly harm the eater and provide no benefit, at the loss of the riddle. As per the comment below: "Staff can modify the chance of a negative event via the 'det_chance' var on the riddle."

For an idea of how powerful the gem healing is, Riddles give a flat 90 for heals, while dorpels give 32.7. This is over the duration and spread through those four seconds, not all at once per tick. It cannot replace health potions, nor does it come close to miracles given the rarity of most decent gems unless you have superb luck.

This entire idea was more theme based, as opposed to one that could mechanically carry. Something to set them apart as a unique SPECIES, as opposed to generic tiny lizard race. If I had my way, I'd do this for every single species to set them apart, but I don't know how far I can push this. Also, Half-Kin exist which throws a wrench in that. This PR itself had host approval for the idea, not implementation.

## Testing Evidence
image
image

image
image

image

## Why It's Good For The Game
It's more content to set different characters apart, and build the world up.
Also, the biggest thing?
Dwarves and other fine craftsmen now have REASON to shun and despise Kobolds.

### DIFF
--- a/code/game/objects/items/rogueitems/natural/stones.dm
+++ b/code/game/objects/items/rogueitems/natural/stones.dm
@@ -349,9 +349,9 @@ GLOBAL_LIST_INIT(stone_personality_descs, list(
 
 		if(iskobold(M))
 			if(M == user)
-				user.visible_message(span_warning("[user] is attempting to eat the [src]!"), span_warning("I begin to eat the [src]!"))
+				user.visible_message(span_warning("[user] is attempting to eat [src]!"), span_warning("I begin to eat [src]!"))
 			else
-				user.visible_message(span_warning("[user] begins to force [M] to eat the [src]!"), span_warning("I attempt to force [M] to eat the [src]!"))
+				user.visible_message(span_warning("[user] begins to force [M] to eat [src]!"), span_warning("I attempt to force [M] to eat [src]!"))
 			if(do_after(user, 40))
 				M.reagents.add_reagent(/datum/reagent/consumable/nutriment, magic_power*1.2)
 				var/healydoodle_again = magic_power+1
@@ -359,9 +359,9 @@ GLOBAL_LIST_INIT(stone_personality_descs, list(
 				playsound(get_turf(src), 'modular_azurepeak/sound/spellbooks/icicle.ogg', 100)
 				qdel(src)
 				if(M == user)
-					user.visible_message(span_danger("[user] eats the [src]!"), span_danger("I devour the [src]!"))
+					user.visible_message(span_danger("[user] eats [src]!"), span_danger("I devour [src]!"))
 				else
-					user.visible_message(span_danger("[user] forces [M] to eat the [src]!"), span_danger("I force [M] to eat the [src]!"))
+					user.visible_message(span_danger("[user] forces [M] to eat [src]!"), span_danger("I force [M] to eat [src]!"))
 
 
 		else // if theyre not either a construct or kobold, and we're not in cmode, beat them 2 death with rocks.


Fatherless1Xx:
I’m not a fan of this because of the precedent it sets. Once kobolds have a unique racial ability, it will encourage others to create exclusive racials for different races, which will turn into a powergaming arms race. That said, I do think it would work well as a general trait rather than something exclusive to a single race.

ophaq:
Fuck yes. Kobolds should eat your zenarii too.

code/game/objects/items/rogueitems/coins.dm
```diff
@@ -198,6 +198,26 @@
 		return
 	return ..()
 
+//Kobolds eating coins. Weep.
+/obj/item/roguecoin/attack(mob/living/M, mob/user)
+	testing("attack")
+	if(!user.cmode)
+
+		if(iskobold(M))
```
dawsonkeyes, line 206: this should require a lengthy do_after and should not be instant

code/game/objects/items/rogueitems/gems.dm
```diff
@@ -14,6 +14,26 @@
 	static_price = FALSE
 	resistance_flags = FIRE_PROOF
 
+//Kobolds eating GEMS. Dwarves, behold, your BANE.
+/obj/item/roguegem/attack(mob/living/M, mob/user)
+	testing("attack")
+	if(!user.cmode)
+
+		if(iskobold(M))
```
dawsonkeyes, line 22: ditto

code/game/objects/items/rogueitems/gems.dm
```diff
@@ -171,6 +191,27 @@
 		slapcraft_recipes = slapcraft_recipe_list,\
 	)
 
+//Kobolds eating RIDDLES. PSYDON WEPT.
+/obj/item/riddleofsteel/attack(mob/living/M, mob/user)
+	testing("attack")
+	if(!user.cmode)
+
+		if(iskobold(M))//Not as effective, on a per-value basis. But it's still MUCH better.
```
dawsonkeyes, line 199: ditto

coffeecoffeecoffee:
I don't know about mammon, coins have much more use throughout the game and even if they provide the tiniest amount of nutrition, those can be eaten in _huge, huge_ quantities..

Fatherless1Xx:
> > > > I’m not a fan of this because of the precedent it sets. Once kobolds have a unique racial ability, it will encourage others to create exclusive racials for different races, which will turn into a powergaming arms race. That said, I do think it would work well as a general trait rather than something exclusive to a single race.
> > > 
> > > 
> > > If we give a race something nifty, it'll encourage people to give OTHER races nifty traits too.
> > > AND THIS IS A BAD THING!!!
> > 
> > 
> > It is. You remember early-day Ratwood, right? Back when tabaxi were basically demigods? It took weeks of powercreep to get them there, we had to fight with the maintainers to get them to dial it back and be satisfied with just slightly better stats. It's never really gone away, we still have it in this codebase. Strength +1 is still locked behind furries or half orcs (which no one wants to play because they're despised and ugly which is the point) The furry cabal always finds a way. Always.
> > Racials are cool on paper, but there's too much developer bias. This is opening that flood gate.
> 
> you CANNOT mention dev bias as the GPTcoder that butchered drakians being an anthro race because you wanted the +1 STR buff for your own fucking character dude

Developer bias is exactly why that change happened, the best stat blocks in the game (+1 STR, +1 CON and +1 END) is still locked behind furry races, because wildkin are allowed subspecies while halfkin are not. I can already point to racial stats being used to benefit furries. Racial passives will be the same.

Savotta:
> Developer bias is exactly why that change happened, the best stat blocks in the game (+1 STR, +1 CON and +1 END) still remains locked behind furry races or ugly races few want to play (orcs/dwarves), because wildkin are allowed subspecies while halfkin are not for some reason. I can already point to racial stats being used to benefit furries. Racial passives will be the same.

boiling down your entire argument to 'WE CAN'T TRUST LE FURRIES!' saves me from actually having to break down your bait. thanks!

PR #480, opened 2025-08-06
New race: Half-drakes

Fatherless1Xx:
## About The Pull Request

Adds a new race of half-drakes.

## Testing Evidence
image
image
image
image

## Why It's Good For The Game

Half-drakes are to half-kin what drakians are to wildkin, a subspecies. We half half-elves. Why not half-drakians?

And yeah, some of you are thinking this is about the stat block. You’re not wrong. +1 Strength shouldn’t be locked exclusively behind furry races or races designed to be ugly and loathed by design. There should be at least one race that isn't furry/despised that can access it.


dawsonkeyes:
this species is covered by halfkin already

PR #486, opened 2025-08-07
Priest's curses do not work on bandits, wretches VL or lich

Fatherless1Xx:
## About The Pull Request

Trying to use a priest curse on bandits or wretches worshippers will do nothing but still apply the cooldown.

## Testing Evidence

image

## Why It's Good For The Game

Curses are a tool to punish towner roles ho defy the priest. This shouldn't be used to shut loud, known antags.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/church/priest.dm
+++ b/code/modules/jobs/job_types/roguetown/church/priest.dm
@@ -380,9 +380,9 @@ GLOBAL_LIST_EMPTY(heretical_players)
                     to_chat(src, span_warning("[H.real_name] is already afflicted by another curse."))
                     return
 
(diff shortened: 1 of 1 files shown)


Fatherless1Xx:
going to rework this so it applies to roles (bandits/wretches) rather than faiths

benj8560:
This should definitely include lich, and VL as well. Incase they are admin spawned as a event or ever enabled again.

ThinkArbies:
It's literally the point of the curses to allow the priest to use them against open antagonists that explicitly and overtly act against the church. It's the priest calling the direct attention of one of their gods onto them. Just because they worship a different god isn't automatic protection from another. Why does everyone seem to fail to understand this?

PR #493, opened 2025-08-08
ser to sir

Randumb-Phantomb:
## About The Pull Request

turns the honoraries ser to sir for the captain, knights, and the heartfelt knight(thought I'd miss it didn't ya)

## Testing Evidence

TBA

## Why It's Good For The Game

Roguecode has a habit of making words spelled stupidly. Ser is probably the most egregious of these, so I'm removing it. The default accent doesn't even replace sir with ser, so after this change there will be no mention of ser in the code period.


KiteAE:
People like it enough as it is.

PR #502, opened 2025-08-09
Adds new entrance into garrison, removes men-at-arms(and veteran) manor key

Mannybrado:
## About The Pull Request
title

## Testing Evidence
image

## Why It's Good For The Game
Men-At-Arms lounging in the keep as though they're supposed to sit around and do nothing in there is getting frustrating. Since, like most servers, this kind of quality issue is never going to be looked over, I can do this instead to make it annoying for everyone so some of them stop doing it. The side entrance is so they can still get inside the garrison.

### DIFF
--- a/code/game/objects/items/rogueitems/keyrings.dm
+++ b/code/game/objects/items/rogueitems/keyrings.dm
@@ -227,10 +227,13 @@
 	keys = list(/obj/item/roguekey/warden, /obj/item/roguekey/walls)
 
 /obj/item/storage/keyring/guardcastle
-	keys = list(/obj/item/roguekey/dungeon, /obj/item/roguekey/garrison, /obj/item/roguekey/walls, /obj/item/roguekey/manor, /obj/item/roguekey/armory)
+	keys = list(/obj/item/roguekey/dungeon, /obj/item/roguekey/garrison, /obj/item/roguekey/walls, /obj/item/roguekey/armory)
 
 /obj/item/storage/keyring/guardsergeant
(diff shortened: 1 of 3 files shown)


code/game/objects/items/rogueitems/keyrings.dm
```diff
@@ -227,11 +227,14 @@
 	keys = list(/obj/item/roguekey/warden, /obj/item/roguekey/walls)
 
 /obj/item/storage/keyring/guardcastle
```
BadAtThisGame302, line 229: Then the keyring's not guardcastle anymore, is it.

Mannybrado: They can still access all parts of the castle, just not the manor inside of said castle.

Great-Beebo:
MAA are literally the duke's guard and retinue. How is them being in the keep a problem? Am I smoking crack or something? They are SUPPOSED to be in the keep.

emsuhr:
They are guards. Their job is to guard. That's why they're there, to act as both town guards and the guards of the manor.  Unlike watchmen, its implied by the class select that Men at Arms are trusted warriors and generally expected to be reliable enough to guard the dukes home.

Troontjelolo:
luluwpp
Me waiting on the useless fucking shitass duke to finally let us into the manor so we can finally arrest the duke's killer that's still at the scene of the crime

KiteAE:
I do not like how this change was mapped, im not against them being disallowed into the main building but i do not think this is the way to do it.

PR #514, opened 2025-08-10
Better Necrans Part 1 DeathDoor fixes spell+Field Burials+Speak With Dead spell

DelineFortune:
## About The Pull Request

Applies fixes of Necran's death door https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/3272
Adds two new spells:

1. Speak with Dead

Allows you to communicate with the spirit of a deceased person. Their ghost supposed to be in the body

2. Collect Coins

If a body has been dead for more than five minutes, you can remove it from the reality to get 1 silver coin. You are supposed to be next to the body.

## Testing Evidence

image

image

## Why It's Good For The Game

Less cringe looc from dead corpses 
A reward for morticians to bury bodies and get money/make server less laggy


ophaq:
One big concern I have with the miracle that removes bodies from reality is well, that. What if some rando monk with the spells or even an acolyte uses it on someone in the medical wing (or they're just too lazy to bring them back to church) and RR's that person? I really like the idea of the spell but I have a feeling it is going to be used with ill intent at times.

DelineFortune:
image
image

Its 15 mins for ANY CORPSE now

Vultchurr:
Yeah let's just skip the part where we decap people we don't like and hide their heads by adding an ability to delete their bodies outright. 

Nice! The server already suffers from a fraggoid problem, I don't see this further enabling it at all!!

DelineFortune:
> Yeah let's just skip the part where we decap people we don't like and hide their heads by adding an ability to delete their bodies outright.
> 
> Nice! The server already suffers from a fraggoid problem, I don't see this further enabling it at all!!

This opinion is irrelevant. 
You are supposed to be dead for 15+ minutes
They are supposed to be necran and have 3 tier miracle

This pr gives no effect to current problem right now, because any person can simply kill you, dig a grave and put a cross on top of it. It takes less than one minute, any person can make it and it makes you not-reviveable.

PR #517, opened 2025-08-10
Gives Jester Master Athletics + Lets Acrobats Do Flips

EvenInDeathIStillServe:
## About The Pull Request

makes it so that jesters get master athletics (they had 0 skill)
ppl with the leaper trait (from the acrobat virtue) can also do flips now
makes it so that you need to be on combat mode to do flips

problem: flips only show up for _other_ people, due to the way vision cones work in roguecode.

## Testing Evidence

breakpoint of the flip triggering under the appropriate conditions, since i can't actually get a screenshot of the flip happening on my own client
image

## Why It's Good For The Game

flip


roarbark:
sillycreep

PR #519, opened 2025-08-10
Lightning bolt balancejak

ShadowDeath6:
## About The Pull Request

Lightning bolt and blood bolt now apply a much longer click delay (6s up from 3s), and they also make you drop your held items.
Blood lightning now immobilizes you for a full second, normal LB is still 0.2 seconds.

## Testing Evidence

image

Works as intended

## Why It's Good For The Game

Lightning bolt and blood lightning got nerfed pretty hard, hopefully this brings them back a bit more without making them too oppressive since I think the spells are cool


coffeecoffeecoffee:
yeah it always felt a little weird it doesn't make you drop your things

anything86:
> maybe sacrifice the click delay then i think the dropping your items is a more flavorful and useful effect even if it's very momentary

Sure, but you know people will still moan about it because what's wrong here is their expectation of what lightning bolt should do and be. They want it to be a hitscan fight-ender. That's stupid. If you want it to be useful you can do things like make it disable people's parry or dodge, or do a good amount of damage, or slow them down a lot, or disarm them, but not ALL of those things at once, there has to be some compromise between flavor and mechanics here.

The opportunity cost of a lightning bolt is tiny. It's 1 second of holding down your mouse tracking someone's movement. Before, it didn't even matter if they grabbed your mouth after you began charging the lightning bolt, because you could still shoot it after the grab and it would stun them, forcing them to drop it. I don't think that is in any way intended and it shouldn't be facilitated by making it disarm you again.

Maromar15910:
> This is pretty jakked. You're unable to move for 5x the previous duration and can't even pick up your weapon within a reasonable amount of time. You, a mage, can hit someone with this and laugh in their face for 6 whole seconds while you hit them with either a chain of spells because they can't grab your ass, or you just take their weapon and throw it up a tree. Like, c'mon.

The immobilization is 0.2 seconds

ShadowDeath6:
> The immobilization is 0.2 seconds

I assume he was looking at blood bolt, which immobilizes for 1s (normal LB is still .2)

anything86:
The combination of hitscan disarm with 6 seconds of not being able to pick up your disarmed weapon (or fight at all) is egregious no matter the flavoring.

ShadowDeath6:
dropped clickcd back to 3s, should still be very powerful defensively provided you keep distance from your target

Maromar15910:
> dropped clickcd back to 3s, should still be very powerful defensively provided you keep distance from your target

I'm not so sure about that considering the wind up time and cost. Three seconds with a stun of only .2 seconds and next to no damage makes the spell pretty unattractive compared to gravity and repulse which both do this spell's job better without as large a windup. 

At this point, maybe it'd be better as a .2 second stun representing a 'flinch' but with better burn damage, armor piercing if it doesn't already have it.

ShadowDeath6:
> armor piercing if it doesn't already have it.

image

Unless there was a change after this comment was made I assume that burns bypass armor
I'll increase the potency on both spells though if they're that weak in terms of damage

PR #520, opened 2025-08-10
doubles nutrition loss from stamina usage and walking

ShadowDeath6:
## About The Pull Request

Spending stamina and walking around will now cost twice as much nutrition as it did before

## Testing Evidence

it's numbers so there really isn't anything to test or show

## Why It's Good For The Game

I think people should get hungry more often, I swear you only need to eat like one meal every 2 days
also might possibly give soilers / cooks more of a reason to exist if people need more food? 

PLEASE testmerge this before fullmerging because this could turn out to be somewhat oppressive at 2x values


benj8560:
I sware I am eating non-stop when I'm in long fights. Strongly not a fan of this idea. We literally had a lich siege the other round we where begging towners to bring us food. You only need to eat a meal every 2 days if your sitting on your ass doing nothing.

This wont make Soilers/Cooks more valued when walking to a stockpile for 2 mammon fruits exist or just a bush.

sippykot:
i like this change

PR #524, opened 2025-08-11
Handcart refactor: Refactored the code, also now you can load things like bins and chests into them, added new feedback messages for doing this and that ADDITIONALLY: refactored bins to be structures and use signals, instead, because it was necessary in the commission of handcart objectives

Metekillot:
## About The Pull Request

This refactors the code on handcarts, as well as giving them the ability to load things like barrels, chests, crates, and so on. As well, I added some feedback messages for why something can't be loaded, or if the cart is too full, and so on.

As well, though it wasn't the primary objective, bins are now refactored to be structures instead of items; they use signals to handle the special considerations necessary for not putting a bucket inside the bin when you meant to dump it in, and so on.

Finally, it adds some extra meat to the description of handcarts so people know.

## Testing Evidence

https://github.com/user-attachments/assets/53db7940-2153-4a22-a815-43ef1b27596d

## Why It's Good For The Game

Handcarts are a cool little thing and this makes them more robust for carrying things like bins and chests, which they are currently unable to do. The changes to bins clean the code up so there isn't a random proc defined on the absolute base /obj/item to handle their work, and instead uses signals, which are cool.

### DIFF
--- a/_maps/map_files/dun_manor/azure_coast.dmm
+++ b/_maps/map_files/dun_manor/azure_coast.dmm
@@ -333,7 +333,7 @@
 /turf/open/floor/rogue/blocks,
 /area/rogue/under/cave/mazedungeon)
 "bh" = (
-/obj/item/roguebin/water/gross,
+/obj/structure/roguebin/water/gross,
 /turf/open/floor/rogue/cobblerock,
 /area/rogue/indoors/shelter)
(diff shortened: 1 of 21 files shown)


Metekillot:
> Please check your code before making a PR.

Please check code before you merge it into main.

PR #529, opened 2025-08-11
Fixes travel-tile interactions w/ leashes, blindfolds & AFKs, + MORE! (Ambush adjustment!)

zoktiik:
## About The Pull Request
- Makes any source of TRAIT_BLIND prevent you from learning a travel path, not just head-bags.
- Travelling while restrained prevents others from learning travel paths when you travel. Solely intended as a mechanical anti-metagamer workaround.
- Added `last_client_interact` mob var tracking the last tick you either manually moved your character or clicked on an atom. Travel tiles requiring specific traits now check for this being within 0.3s, or 3 ticks. No client interaction, no travel.
  - This should not be CPU intensive at all; this is a very "dumb", non-stateful system that doesn't need checks to update.
  - This is to prevent others from OOCly torturing you as their prisoner (repeatedly pushing/kicking/throwing/teleporting you in) until you go AFK and automatically travel thru. Also just prevents them from abusing you if you're AFK flat out.
  -  This doesn't stop you from getting pulled in or leashed in by someone else, just from starting the travel yourself (and thus potentially outing the location unwillingly, without your input.)
- Others learning the travel point's trait happens when you actually travel-teleport in view of people, not when you touch the tile.
- **You are notified on teleport** when someone gains the trait themselves from watching you use a travel point.  HOWEVER, this notification is **suppressed** if they are using SNEAK intent.
- Getting grabbed or held on a leash completely blocks you, both from starting travel and from teleporting at the end. You can still travel if someone is pulling/leashing you in.
- **BUGFIX:** Travelling while having others leashed to you takes them with you, and no longer just throws them down a z-level.
  - You can take as many people with you at once as you currently have pulled and leashed. Theoretically the cap for this is around 4: 1-2 connected leashes in-hands, 0-1 others grabbed, 2 connected leashes in the hip slots.
  - Leashing yourself, or leashing + grabbing someone, does not trigger a double-teleport.
- Travelling with dropped leashes takes them with you.
- Moderate refactor of travel tile code, deduplicating it and splitting it into several different procs, as well as reorganizing the procs themselves within the file.
- Added new global procs for checking if someone else has mob leashed, getting list of mobs this one is leash "master" of, and getting the dropped leash of a mob.
- Couple miscellaneous additions I made during testing:
  - *MISC:* added text naming the currently leashed person to leash examine(). Does not reveal real_name, however.
  - *MISC:* having compliance mode on makes you get leashed faster.

- **Additionally**, to apply `last_client_interact` to something else it might be useful for, **random ambush checks** (with `always = FALSE`) **need you to not be AFK in order to trigger.** In other words, those ambushes get triggered by actions *you* took.

## Why It's Good For The Game
- Travel tiles now have proper behavior to handle leashes whenever they're involved.
- Anything that makes someone blind can be used to prevent them from learning the entrance, not just checking for whether a sack is on someone's head.
- Guards can't bring an AFK/SSD/unwilling antag and mechanically force them to give up camp locations.
- However, you can be stopped from escaping via a travel tile if you get grabbed during the travel time. Make sure you're not being directly tailed.
- Makes it much less uncertain whether the location is being given up or not. 
  - But if you manage to stay undetected while someone is travelling, you can still covertly watch them and get the location without them knowing.
- Leash QoL.
- Modular, cleaner code.

- **Ambush change:**
  - Pulling others in the forest should be significantly less agonizing now. If you SNEAK or have `TRAIT_AZURENATIVE`, the benefit is also conferred to them, so long as they're not making movements or atom clicks of their own.

## Testing Evidence
I am fairly certain I tested absolutely everything in this PR (*without the luxury of using the testmap*),  but here's a couple demonstrations.

https://github.com/user-attachments/assets/e17de4d5-1be5-4232-b419-bbe270a3ba35

2oq1cn

s4qbta

### DIFF
--- a/code/modules/mob/living/ambush.dm
+++ b/code/modules/mob/living/ambush.dm
@@ -9,8 +9,11 @@ GLOBAL_VAR_INIT(ambush_mobconsider_cooldown, 5 MINUTES) // Cooldown for each ind
 /mob/living/proc/consider_ambush(always = FALSE)
 	if(prob(100 - GLOB.ambush_chance_pct))
 		return
-	if(HAS_TRAIT(src, TRAIT_AZURENATIVE) && !always)
-		return
+	if(!always)
+		if(HAS_TRAIT(src, TRAIT_AZURENATIVE))
+			return
+		if(world.time > last_client_interact + 0.3 SECONDS)
+			return // unmoving afks can't trigger random ambushes i.e. when being pulled/kicked/etc
 	if(mob_timers["ambush_check"])
 		if(world.time < mob_timers["ambush_check"] + GLOB.ambush_mobconsider_cooldown)
 			return


roarbark:
Huge

PR #532, opened 2025-08-12
Gives the WW's Ungrabbable.

arnociiroc:
## About The Pull Request
The Ungrabbable trait was added to ww's on transformation

## Testing Evidence
image

## Why It's Good For The Game
big beasts probably shouldn't be stoppable by a tiny man grabbing them


coffeecoffeecoffee:
they should almost certainly have stats that make them _functionally_ extremely difficult to grapple down but removing the ability to do so wholesale really fucking sucks from the perspective of... doing anything but super murdering them on the spot, actually
why not just give them grapple expert or something so you can still, say, lead one by the hand or drag a crit one down the road, but can't realistically chestgrapple one into the dirt

ShadowDeath6:
> They can be grabbed if-
> 
> 1. Combat mode is off.
> 2. is unconscious.

this is good

Metekillot:
This doesn't account for werewolf on werewolf wrasslemania or something bigger and meaner grabbing them, but it seems an acceptable stopgap until someone ( almost definitely probably going to be me ) reworks the grab spaghetti code

Metekillot:
My plan is just to run the code for resisting a grab (minus cooldowns) with tiered stamina cost based on how hard it was for you to resist a grab from whoever, instead of forcing you to burn a click cooldown and a moment of helplessness, so the chain will be:

grab_exploiter clicks you -> code to resist runs automatically if you're in combat mode -> if you would resist successfully it's as if you were never grabbed, possibly minus a bit of stamina, depending on how tough grab_exploiter is as a wrestling opponent

PR #536, opened 2025-08-13
Refactor the initial grab code (no balance changes, added an animation)

Metekillot:
## About The Pull Request

This just refactors the grab code to be, perhaps, slightly, if only a little, less fucking awful. I amputated dead code with no effect in any circumstance and also removed commented out code (which is sloppy code practice to have).

Finally I added the wiggle attack animation to grabbing somebody to make it just a bit more obvious to onlookers what is happening.

Ignore the line diff; I didn't add that much code, I just moved a couple of files for eventually busting the mammoth amount of grab code into separate files to make the chain less confusing.

## Testing Evidence

https://github.com/user-attachments/assets/79f0ad8e-8ff1-44a7-a75e-d18235e6f044

## Why It's Good For The Game

Makes the code slightly less goddamn awful.


benj8560:
> Can you add the animation to struggling/upgrade attempts too?

there is a powerful meta involving grabbing peoples weapon arms in the chaos of combat and hoping they don't notice it. _that I totally don't partake in at all :3_ so yeah would be a really good idea to add more feedback.

Metekillot:
Yeah I'll take a look

PR #541, opened 2025-08-14
Buffs MILF spawnrate

SylvinSeik:
## About The Pull Request
Changes the stat modifier from being middle aged from: 
-1 Speed +1 Endurance to:
-1 Speed +1 Constitution +1 Fortune (I'm not updating the rest of the PR description, just look at anything86's comment)

This is in an effort to encourage players to make different types of characters without COMPLETELY stunting themselves.  It's still a "net negative" if you consider speed as 2 statpoints like statpacks and racestats do, but it'll be a bit less punishing towards the middle aged users.

## Testing Evidence

It's a one line change bro

## Why It's Good For The Game

image

I don't expect this PR to work magic, but I'm hoping it helps people feel a bit more free to make characters they enjoy without feeling like they are dropping an anvil on their foot for no trade.

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -1497,7 +1497,7 @@ Slots: [job.spawn_positions] [job.round_contrib_points ? "RCP: +[job.round_contr
 							if (AGE_ADULT)
 								to_chat(user, "You preside in your 'prime', whatever this may be, and gain no bonus nor endure any penalty for your time spent alive.")
 							if (AGE_MIDDLEAGED)
-								to_chat(user, "Muscles ache and joints begin to slow as Aeon's grasp begins to settle upon your shoulders. (-1 SPD, +1 CON)")
+								to_chat(user, "Muscles ache and joints begin to slow as Aeon's grasp begins to settle upon your shoulders. (-1 SPD, +1 CON, +1 FOR)")
 							if (AGE_OLD)
 								to_chat(user, "In a place as lethal as PSYDONIA, the elderly are all but marvels... or beneficiaries of the habitually privileged. (-1 STR, -2 SPE, -1 PER, -2 CON, +2 INT, +1 FOR)")
 						// LETHALSTONE EDIT END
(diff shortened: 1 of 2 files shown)


zeratino:
Ohhh u so sexy

LostLittleMeowMeow:
but end is much more important than con, especially in a fight. this'll drive the milfs into hiding

SylvinSeik:
> but end is much more important than con, especially in a fight. this'll drive the milfs into hiding

My original plan was to give them +1 con +1 end -1 spd, but just about every person smarter than me said to just make it +1 con because it's both an improvement, and not a "statistical improvement" that might impede the PR.  I suspect that even if it was those numbers, we'd see several times more adults than middle aged even with the current middle aged job locks.  If maintainers like that idea I would GLADLY change it though.

coffeecoffeecoffee:
> > but end is much more important than con, especially in a fight. this'll drive the milfs into hiding
> 
> My original plan was to give them +1 con +1 end -1 spd, but just about every person smarter than me said to just make it +1 con because it's both an improvement, and not a "statistical improvement" that might impede the PR. I suspect that even if it was those numbers, we'd see several times more adults than middle aged even with the current middle aged job locks. If maintainers like that idea I would GLADLY change it though.

not a maintainer but _I_ like that idea

Wesreidau:
> > but end is much more important than con, especially in a fight. this'll drive the milfs into hiding
> 
> My original plan was to give them +1 con +1 end -1 spd, but just about every person smarter than me said to just make it +1 con because it's both an improvement, and not a "statistical improvement" that might impede the PR. I suspect that even if it was those numbers, we'd see several times more adults than middle aged even with the current middle aged job locks. If maintainers like that idea I would GLADLY change it though.

You were this many seconds old when you learned a collection of people is, in fact, dumber than an individual.

anything86:
My suggestion is make the statistical bonus be to some irrelevant stat, like Fortune. Fortune is the perfect pick for this because people love to think it affects everything and wins you fight when the most notable thing it does (when not below 10) is reduce dodge weapon clipping chance.

+1 CON +1 FOR -1 SPD. There, solved your dilemma. Also makes some sense that someone who managed to live for that long may have 'good luck'.

Wesreidau:
I'd sooner say anyone who lived that long has learned a little more about everything, such that 'fortune' reflects their broader skill and lack of naivety. Just not the same amount of skill as might be a full level increase.

Randumb-Phantomb:
Just give adult a -1 int lol

PR #548, opened 2025-08-14
Sanctuary miracle + no grab trait

Fatherless1Xx:
## About The Pull Request

This is a new t4 miracle for Eorans, it's additional rather than replacement, given how underwhelming the tree miracle is.
This applies godmode, no_grab, grab_immune and pacifism for 40 seconds. It's a way to force a fight to end peacefully.
More or less, it's the hugbox protection adventurers get when starting but turned into miracle form.
Comes with a new spell icon.

Also adds trait_no grab. This stops you from being able to grab others.
## Testing Evidence

image

image

image

## Why It's Good For The Game

This is a thematic way for Eorans to stop killings and reduce fragging.

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -67,6 +67,7 @@
 #define TRAIT_PERMAMUTE "Permanent Mute"
 #define TRAIT_HERESIARCH "Forbidden Knowledge" //allows entry to ascendant church
 #define TRAIT_EXCOMMUNICATED "Excommunicated" //cannot be target of healing miracles of Ten or anastasia or cure rot
+#define TRAIT_NOGRAB "No Grab" //cannot grab
 
 
 //Hearthstone port (Tracking)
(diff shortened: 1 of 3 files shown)


Fatherless1Xx:
Fixed it by adding a new nograb trait. It does what it says on the tin. If you have it, your grab gets qdeleted. This is so you cannot use sanctuary to enable uncounterable grab spamming.

image

NotGoldenglow:
I'd suggest making it AoE or slowing people down. Maybe rooting them? Because otherwise it might go from 'forcefully stops a fight' to 'the receiving party can slip out of and run away while being unstoppable', or the complete opposite: Used to completely shut down an antagonist while they get physically surrounded by people and is unable to do anything to slip free.

Inveldy:
As Notgolden said. This should be an AOE (possibly a fairly large one) and should root the people who received it in place. This way it forces involved parties to stop and talk rather than run away or surround someone while waiting for it to end.

PR #549, opened 2025-08-14
Standing Up Fix

anything86:
## About The Pull Request
Makes it so looking in a different direction does not stop you from standing up.

## Testing Evidence
image

## Why It's Good For The Game
Makes knockdowns in combat less of a death sentence. If you change directions you no longer have to start the countdown all over again, though if you are grabbed you still cannot get up.


roarbark:
good change

RyszardParys:
You could've already done that with your eyes fixed ngl.

PR #550, opened 2025-08-14
Customizable Donor Text and Making SCOMs more modular/slightly less messy

Multigamer99:
## About The Pull Request

Requested by Kumie

CHANGELOG
- Unified the SCOM code for easier future modifications.
  - Merged the crown code into the SCOM file so it's all in one place instead of arbitrarily spread out.
  - Changed most SCOM behaviour to be on a datum attached to SCOM devices so that functions and behaviour can be reused far more readily.
  - Updated some text to make it more clear to the user what is going on, specifically regarding Jabberline errors.
  - Altered the internal IDs of some SCOM devices so that they can better inherit behaviours and be organized a bit nicer.
    - Updated the maps and other code files accordingly.
  - Did not fix the janky implementation of thaumaturgy for now.
- Implements a rudimentary patreon verb system.
- Redoes part of the current patreon specialty voice color code to allow the color to be chose per-player.
  - Note: this is PER PLAYER not PER CHARACTER
  - Currently this feature is always available but it can be restricted to lobby only if so desired.
  - Implemented via two verbs:
    - One to toggle it on and off (so you don't have to use custom colors if you don't want them).
    - One to change the color.
      - The same color enforcement that is applied to character voice color (the color that appears for their name) is applied here.
  - Cleans up code from the previous implementation that isn't used anymore.

KNOWN ISSUES
The only one I encountered in my testing is a runtime where it is claiming a value is null that isn't null in both variable view and based on a temporary debugging check I had used. The code also runs as if it wasn't null. Honestly, I've got no fucking clue what causes it since there shouldn't be a runtime happening there at all, but I can poke at it later since I'm putting this up since I'm about to go to bed. I'll come patch it later.

## Testing Evidence

Donor Level used for testing.
A

Chat output from using the patreon verbs.
Voice Changing
voice change error text

Patreon verbs in the status panel.
C

Status panel without patreon rewards.
Nodonor

Talking without patreon rewards.
Nodonortalk

Various testerings.
Choker
crown
HoundstoneandSerfstone
jaberline
jaberlinecolored
MatthiosCoin
portables mute
scom

## Why It's Good For The Game

Request by Kumie

Noone changed the default color for patreon donors from the last patreon text pr and it was a color that wasn't really the most pleasant. This lets the donor pick the color themselves (which also helps it be a better incentive for folks to donate and support the server which is good for helping the server stay running). As for the SCOM changes, yeah I figured I'd clean it up while I was in there. Might slightly improve runtime performance due to not having to look through every SCOM device to figure out which ones to message at the cost of slight ram usage? Still good for development since the SCOM code was (and sorta still is) a mess and this helps remedy that.


modular_azurepeak/code/modules/clothing/rogueclothes/crown.dm
```diff
@@ -1,108 +0,0 @@
-#define GARRISON_CROWN_COLOR "#C2A245"
-
-/obj/item/clothing/head/roguetown/crown/serpcrown
```
BadAtThisGame302, line 3: Since you're deleting this, is there another Crown that will take its place, or will the Duke wear no Crown?

Multigamer99: The crown still exists, it's just in a different file now so it can be with the other SCOM equipment.

This is mostly because every other piece of SCOM equipment had its own file and the crown was uniquely in its own file for some reason.

code/modules/roguetown/roguemachine/scomm.dm
```diff
@@ -1,4 +1,248 @@
...
+		throw EXCEPTION("YOU MUST CALL 'setup()' ON A SCOMMODULE BEFORE USING IT!!!")
+	if(A == src)
+		return
+	//Vrell - moved these checks to the repeat side to make the code look less ass and make it a bit more dynamic.
+	if(target == SCOM_TARGET_COMMONS && mute_commons)
+		return
+	if(target == SCOM_TARGET_GARRISON && mute_garrison)
+		return
+	if(tcolor)
+		parent_object.voicecolor_override = tcolor
```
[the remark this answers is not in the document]
Multigamer99: FIgured it out. Sending a commit shortly after I check another potential minor oversight.

Multigamer99:
CHANGELOG:

- Merged this shit with the loudmouth shit
- Made the loudmouth reflect the behaviour stated in messages on its respective PR
  - Namely, the public loudmouth is unmutable on portable non-crown scom devices
    - I'm pretty sure that was unintentional since if anyone would be able to mute the public loudmouth, it would be the grand duke.
- Minor cleanup.
- Makes the SCOM directory only include SCOMs that can be called (since that's how the new numbering system works)
  - Yes, this means that calling a number not on the directory but between directory entries does call a given SCOM on the map.
- Makes the SCOM directory visible from ALL devices that can use jabberlines.
  - Still only the SCOM currently but hey it could be helpful down the line if someone makes a portable jabberline thing.
  - The current implementation also allows dynamically adding things down the line. Maybe someone will make a scomstone change/variant that allows you to set its name on craft/first interact (like how (old? idk i haven't checked the current iteration) tabards let you set their colors/patterns) which would then appear on the scom registry.

Multigamer99:
> Now that Loudmouth is merged, you'll have to resolve your conflicts. Let me know if you need help with such.

Should be fixed now.

Multigamer99:
Changelog:

- Maybe fixes compatibility issues with an existing testmerge even though there isn't an open textmerge-labeled PR that messes with the preferences files.
  - Working entirely on assumptions since I do not know where exactly the conflict was. I imagine it was with the save version number I chose.
  - This action was brought on by a discord message informing me of a conflict with a testmerge that changes the preferences file(s).

PR #553, opened 2025-08-14
Hotfix an oversight from my leash proc that will cause runtimes

zoktiik:
## About The Pull Request
Saw the notif that the main PR got merged and had a small nagging feeling that there's something I overlooked about those procs. Sure enough, there was.

When optimizing the earlier checks that see if they're still nearby and that they belong to the person, I neglected the fact that, after that block, we still don't know for sure if the leash even has an assigned pet. This will runtime for people holding a mix of connected and unconnected leashes at the time they teleport to the other travel tile.

## Testing Evidence
None. Did this on my phone.

## Why It's Good For The Game
Less weird runtimes yippee

### DIFF
--- a/code/game/objects/items/rogueitems/leash.dm
+++ b/code/game/objects/items/rogueitems/leash.dm
@@ -391,7 +391,7 @@
 					nearby_pet = target
 					break
 			if(!nearby_pet)
-				if(!do_not_remove) // leash will unregister them next process(), to not spontaneously throw pet up a z-level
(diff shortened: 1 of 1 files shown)


EvenInDeathIStillServe:
> * [Fixes travel-tile interactions w/ leashes, blindfolds & AFKs, + MORE! (Ambush adjustment!) #529](https://github.com/Scarlet-Reach/Scarlet-Reach/pull/529)
> 
> @EvenInDeathIStillServe I ah make ah the mistake 🤌

We gonna kill you vro.

PR #555, opened 2025-08-14
Revert "Pre-optimization frameworking: /datum/Destroy() should not sleep and it is now so; also updated dependencies.sh, fixed some style crimes"

EvenInDeathIStillServe:
Reverts Scarlet-Reach/Scarlet-Reach#522

THIS DOESNT EVEN COMPILE


Metekillot:
There was one var that needed to be fixed. I fixed it.

PR #556, opened 2025-08-14
The 🪢 (How to train your Lupian)

tmyqlfpir:
## About The Pull Request

This PR adds 🪢 mechanics for every 🚀 equipped characters

## What does it add

In the ERP menu there are now three new (aggro grab only) exclusive choices that will lock the two characters together by the hip upon 🍾.

To untie the 🪢, the lupine must...
- Run in any direction (a str vs con check determines if the tie remains)
- No longer be aroused
- Be greater than 2 tiles apart (jump over fence, fall down cliff, etc)
- Start another action in the ERP menu
- Be in a crit/asleep/dead state
- Either person teleporting further than 10 tiles (this will enact an emergency 🚀 removal)

Players can simply opt out by turning off their ERP menu to OOCly undo the 🪢, or discuss in LOOC.

## Testing Evidence

I have tested and shown video examples of this working on the discord channel.

The new ogg sound effect is from the OpenNSFW Sound Pack, and listed as under the Attribution Free category (c00 license).

## Why It's Good For The Game

People were asking for it, and I wanted to see it.

### DIFF
--- a/code/modules/client/customizer/customizers/organ/genitals.dm
+++ b/code/modules/client/customizer/customizers/organ/genitals.dm
@@ -63,6 +63,7 @@
 		/datum/customizer_choice/organ/penis/tapered_double_mammal,
 		/datum/customizer_choice/organ/penis/tapered,
 		/datum/customizer_choice/organ/penis/tapered_knot,
+		/datum/customizer_choice/organ/penis/tapered_knot_mammal,
 		/datum/customizer_choice/organ/penis/tapered_double,
 		/datum/customizer_choice/organ/penis/tapered_double_knot,
 		/datum/customizer_choice/organ/penis/barbed,
@@ -83,7 +84,7 @@
 	customizer_choices = list(
 		/datum/customizer_choice/organ/penis/human_anthro,
 		/datum/customizer_choice/organ/penis/knotted,
-		/datum/customizer_choice/organ/penis/tapered_knot,
+		/datum/customizer_choice/organ/penis/tapered_knot_mammal,
 		)
 
 /datum/customizer/organ/penis/feline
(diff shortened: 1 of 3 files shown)


sippykot:
you forgot about other knotted pinthle types unless im missing something
image
theres the double knotted and the barbed knotted

PoltavskaPraca:
image

sippykot:
tbh can we not have brute damage from knot pulling? i feel like it will just cause problems

coffeecoffeecoffee:
> tbh can we not have brute damage from knot pulling? i feel like it will just cause problems

why would you remove soul
this is like the best part of it

sippykot:
> > tbh can we not have brute damage from knot pulling? i feel like it will just cause problems
> 
> Maybe, what would you suggest then?

just pain and stun?

Murderfoxen:
Um, do you need beta testers? Asking for a friend

Murderfoxen:
> > tbh can we not have brute damage from knot pulling? i feel like it will just cause problems
> 
> why would you remove soul this is like the best part of it

I need my naked lupien stuncrit martial artist build special finisher and this man isnt bering very supportive to the dream smh

Maromar15910:
> > There's some players that are prefbroken by knotted members, but don't mind rping as long as the part is omitted. Adding unavoidable mechanics kills that happy medium. I'd like it if there was a dedicated button on the sex menu to employ the knot/tying so it can knot (hah) be pressed in such cases and to simply simulate being able to control yourself and not stick it all the way in, especially if it's a partner's first time or they were being cautious for some reason or another.
> 
> I've considered this and partly through the PR I made it so knot tying is only reserved for three of the sex actions (Sodomize with knot, Knot their throat, Knot their cunt) which are all aggro grab only actions. By using these actions alone it'll only let the players tie together, otherwise every other action will behave as it did previously.

Thanks! ^.^

arnociiroc:
getting knotted several times stacks the debuffs, and doesn't restore your stats.
image
I had Ten Int originally, Two people were knotting at the same time

Randumb-Phantomb:
> getting knotted several times stacks the debuffs, and doesn't restore your stats. image I had Ten Int originally, Two people were knotting at the same time

you should get permanent brain damaged for being knotted several times in one sitting lol.

Maromar15910:
Are you sure about the fucked stupid duration? Is there a way to end it early? I think smacking yourself or getting slapped should cancel it or something. It's flavorfull but 10 mins on compliance mode is a no-go for this kind of server. 

(maybe when we make another one focused on that kind of stuff...)

PR #557, opened 2025-08-14
Revert "Handcart refactor: Refactored the code, also now you can load things like bins and chests into them, added new feedback messages for doing this and that ADDITIONALLY: refactored bins to be structures and use signals, instead, because it was necessary in the commission of handcart objectives"

EvenInDeathIStillServe:
Reverts Scarlet-Reach/Scarlet-Reach#524

It's throwing warnings.


Metekillot:
Which warnings?

PR #576, opened 2025-08-17
Continuous Butchery

EvenInDeathIStillServe:
## About The Pull Request

makes it so that butchery rolls are per item, instead of all-or-nothing

it'll continuously butcher the carcass, dropping butchery results with each iteration until there's nothing left, then gibs

## Testing Evidence

image

image

## Why It's Good For The Game

I am a code tyrant.


Genevesh:
> so how is this PR going to affect butchery times on average? 30 seconds per cut as unskilled seems comically long if you're cutting 6 or 7 parts off of a thing you killed

Smarter characters will learn butchering faster. Another reason not to minmax combat stats!

PR #581, opened 2025-08-17
Etruscan accent

Randumb-Phantomb:
## About The Pull Request

Adds an Etruscan accent to the game preferences menu.
Links this new Etruscan accent to the italian replacement.
Adds a whole host of words and phrases to the italian replacement.

## Testing Evidence

TBA

## Why It's Good For The Game

Etrusca doesn't have an accent, and the tiefling accent doesn't count. 
Plus, italian representation in SS13 that isn't "Mama mia pizzeria"


Just-Sim:
no, etrusca isn't solely italian, it's spanish AND italy in one place because whoever decided it to be this way thought all latin-based countries share the same culture (they don't), i tried to fix this somehow in AP's lore by making the north italian and the south spanish

this is accentuated by the fact that there are 2 mercenary classes from the 2 different cultures, the etruscan condottieri (italian) and the etruscan vaqueros (spanish)

Just-Sim:
> Then I'll just say north Etruscan 🤷

i'm fine with this

PR #587, opened 2025-08-18
Passive Grab Nerf + Grab Resist CD change

anything86:
## About The Pull Request
Before, all 'resists' (be they putting out fire, or trying to break a grab) had an unreasonable cooldown of two (2) whole seconds in between them. For comparison, a pick intent hit is 1.4 seconds of click delay. This made breaking out of even the most pitiful passive grab a herculean effort in combat and enabled people to disable others by spamming passive grabs.

This PR reduces the CD on grab resisting to 0.8 seconds, the same as a dagger stab (and close to the 1 second of click CD for grabbing), and makes it so whenever you passively grab someone who is actively resisting, they get an automatic, free attempt at breaking the grab. 

In practice this means that grabbing someone in combat will be ineffectual if you don't have the stats or skills to actually properly grab them, and that resisting out of passive grabs isn't taking you twice as long as it takes for them to grab you right back.

**Note: Putting out fires still prompts a 2 second click CD. That wasn't changed. This does not touch fire.**

## Testing Evidence
image

## Why It's Good For The Game
Passive grab spam is incredibly zozozexual and makes fighting at a numerical disadvantage very unfun if any of them remembers to just grab your arm repeatedly.

Because resisting a grab manually still costs you stamina, it shouldn't be a problem if people can resist out of them at reasonable speeds rather than once every two seconds.

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -486,7 +486,8 @@
 	if(isliving(AM))
 		var/mob/living/M = AM
 		if(M.cmode && M.stat == CONSCIOUS)
-			M.resist_grab(freeresist = TRUE) //Automatically attempt to break a passive grab if defender's combat mode is on. Anti-grabspam measure.
+			if(M.get_skill_level(/datum/skill/combat/wrestling) > 4 || src.get_skill_level(/datum/skill/combat/wrestling) < 5) //Grabber skill less than Master OR grabbed skill at Master or above.
+				M.resist_grab(freeresist = TRUE) //Automatically attempt to break a passive grab if defender's combat mode is on. Anti-grabspam measure.
 
 /mob/living/proc/send_pull_message(mob/living/target)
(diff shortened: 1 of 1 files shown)


DorfGuy1:
> In no unclear words, the current state of passive grabs is such bullshit we had to go and code a whole ass 'GRAB IMMUNITY' trait for antagonists who have 20 (!!!) Constitution and Strength and Legendary Wrestling (VL, WWs), because its action economy is so fucked that even at 95% resist chance it's still crippling to get grabspammed in a fight.

ACTION ECONOMY: That is SUCH a good concept to incorporate into the discussion.

coffeecoffeecoffee:
i'm now unable to grab my own saiga to drag it through a door

you could make tamed saigas not resist grabs at all ideally because it was already incredibly annoying for no gameplay reason

Vultchurr:
pain

Nigh impossible to move livestock around, really have to spam the shit out of them constantly. Pls fix.

PR #591, opened 2025-08-18
ports omniwounds from azure peak

ShadowDeath6:
https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/3165
https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/3203
please read the original pull requests
**PLEASE** testmerge this pull request extensively before merging to avoid any unforeseen issues since it touches a lot of different things across the game

## About The Pull Request

CON now influences bleed rate from wounds. The higher, the slower, up to 50% slower at 15 CON.
Dagger pick penfactor has been reduced to 75 from 80.
There is now only one wound that is created per damage type (stab, bite, blunt, etc).
Attacking the same bodypart over and over again will increase its bleed rate, pain and it'll take longer to sew, until you get a major wound that's equivalent to an artery hit

also ports blood transfusion and wound cure miracles (https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/3203) since this was balanced around the existence of those spells

## Testing Evidence

image

image

You fall over EXTREMELY quickly from bloodloss once the artery gets hit, like maybe 4 seconds from my test

## Why It's Good For The Game
(taken from the PR in question)

-No more hidden "dozens" of wounds. Sewing is now way simpler. Might last about as long, though.
Everyone is now more fragile. At least, as far as taking raw unarmored hits goes. The risk of bleeding out even without any of those crits is significantly higher. I do not know if this is going to make things anymore or less fun, since it's all in the numbers.
Grabbing your bleeding bodyparts is now going to be a loooot more important.
Higher damage intents / weapons will be able to very quickly cause an artery bleed. Rend and other higher-damage intents in particular.

-This was mostly meant to replace existing behaviour 1:1, however when I saw that "large" wounds have only 1 bleedrate and you'd need TWENTY of them to match an artery crit (at which point you've definitely already created more than one crit on that limb), I decided to experiment with the wound system, making it more lethal the more you're hit in the same bodypart.

-Whether this is more fun or not remains to be seen. If worst comes to worst I can simply replicate the numbers of the previous system 1:1 and move on, as reducing the nonsense in the backend of wounds was the primary goal here.

### DIFF
--- a/code/datums/gods/patrons/divine_pantheon.dm
+++ b/code/datums/gods/patrons/divine_pantheon.dm
@@ -63,7 +63,6 @@
 					/obj/effect/proc_holder/spell/targeted/beasttame			= CLERIC_T2,
 					/obj/effect/proc_holder/spell/targeted/conjure_glowshroom	= CLERIC_T3,
 					/obj/effect/proc_holder/spell/invoked/wound_heal			= CLERIC_T3,
-					/obj/effect/proc_holder/spell/targeted/conjure_glowshroom	= CLERIC_T3,
 					/obj/effect/proc_holder/spell/self/howl/call_of_the_moon	= CLERIC_T4,
 	)
 	confess_lines = list(


ShadowDeath6:
> How do the blunt weapons tie in here?

from what I can glean from the code, bruise wounds are much more likely to occur against armored foes than artery cuts are (meaning, you rarely ever hit arteries if the foe is fully armored) and cause a LOT more pain (and the wounds can't be stitched up with a needle)
I don't think they're nearly as lethal as arterial bleeds are

PoltavskaPraca:
Made by FreeStylaLT
omniwounds-min

DorfGuy1:
I seem to remember the AP PR mentioning that it would be impossible get wounded through armor with sheer penetration power alone. I don't feel well about that particular bit.

DorfGuy1:
One last thing: The pick onniwound. I'm not sure if this will mean that you can't skull pierce people with picks or daggers anymore.

ShadowDeath6:
image

new numbers, extremely massively increases lethality potential on high damage stab / slash weapons versus unarmored

ShadowDeath6:
image

lesser miracle now reduces wound pain (I don't _think_ it did already, but it was mentioned in documentation)

FreeStylaLT:
oh thats simple since that same proc upgrades / creates a wound, it can just return that.

dreamseeker_GRolDoE0Lc

https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/3586

ShadowDeath6:
image

message for when a wound is first made

ShadowDeath6:
I just realized bruises don't upgrade so this spams chat AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAIIIIIEEEEEEEEEEE

ShadowDeath6:
Not at my pc currently to fix conflicts

DorfGuy1:
> * You **can** get wounds through armor, you just won't be able to stack the bleed up to an artery. You can change those thresholds and limits in the code, it's just one number.
> * Gouge wound was added for pick intent and projectiles because intents that were built around penetrating with full damage AND can crit through the armor ALSO being able to make deadly wounds seemed inzane.
> * These wounds have no influence on crits like skullcrack whatsoever.

If things are still like they used to be: which is to say, any bleeding that wasn't a crit bleed was almost useless, I see a problem with that, otherwise...good. 
I would seperate the pick, dagger pick, and arrow wounds if I had my way. Picks from things like pickaxes and hammers and stuff should be nastier than daggers, but daggers could use a bit of a nerf. However, i think bows and stuff are in a bit of a bad place rn. If you go whole hog into investing on an archer build and rock a crossbow or a yew bow, you SHOULD be able to REALLY screw someone up. If that's not a thing than maybe we give arrows the option to always trigger the movement debuff for damage if targeted at the legs. Then at the very least they'd be good for keeping people from running away. This is my two cents as a player.

DorfGuy1:
> dagger stab wounds and pick wounds (gouges) are different, I'll buff up pick intent a bit

I'm more worried about arrows and such.

PR #600, opened 2025-08-19
Dead nose prevents sewer water mood loss

SylvinSeik:
## About The Pull Request
Makes dead nose prevent the "Putrid stinking water!" mood debuff from sewer water.  Simple.

## Testing Evidence
Quick hack-jobbed image. Left is with dead nose, right is without.
image

## Why It's Good For The Game

Ugly people can now find solace in the sewers, for that is where they belong.


Heheniy:
breathless should have this too

PR #610, opened 2025-08-21
construct buff

RiceGateau:
## About The Pull Request

Makes constructs feel no pain

## Testing Evidence

Not really needed with such a minor alteration.

## Why It's Good For The Game

Constructs cant use drugs, sleep, be revived without the church, and it just makes sense for them not to have pain since they don't have nerve endings.


coffeecoffeecoffee:
> As if golems needed to be even better than rot-cured.

Rotcured can like... Learn skills, can't they?

benj8560:
Golems are in a really good spot rn. They can heal effortlessly by just hoarding rocks. They never need the rest. Coal is argueablely better then a mana potion since its instant and DIRT cheap. I don't think they need colossal buffs right now. You will notice Rotcured is pared along side Easy Dismember giving it a ton of drawbacks for painless.

PR #612, opened 2025-08-21
Improve mail feedback

tmyqlfpir:
## About The Pull Request

This PR adds more additional feedback to the mailing logic.

* Ask the player if they still wish to send the letter if the recipient is not found
* Adds a mention of leaving the from name field blank to be able to send mail anonymously
* Adds a chat notice message hint if tried to collect their mail if their mail was stolen/burned/missing
* If the recipient tried to collect their mail and there was none to be found, silently readd the `ugotmail` status effect for the recipient if their mail is readded into the mastermail machine

## Testing Evidence

I have tested this logically and confirm it does exactly as above describes.
tested

## Why It's Good For The Game

More feedback helps players get less frustrated and cuts down on false bug reports.

### DIFF
--- a/code/modules/roguetown/roguemachine/mail.dm
+++ b/code/modules/roguetown/roguemachine/mail.dm
@@ -170,7 +170,7 @@
 					if(H.real_name == send2place)
 						mailrecipient = H
 						break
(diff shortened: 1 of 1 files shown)


EvenInDeathIStillServe:
Can you add an "are you sure?" Alert to sending mail if there's no recipient found with that name?

PR #613, opened 2025-08-21
wild goblin eyes now hurt you while implanted

ShadowDeath6:
## About The Pull Request

non-goblins will take tox damage and extremely high eye damage, going blind in a few minutes
also fixes the check that was in there because **YOU CAN'T CHECK A SPECIES LIKE THAT!!!!!!!**

## Testing Evidence

image

also tested with a goblin and it worked fine 
## Why It's Good For The Game

easily-accessible super darkvision for literally anyone who cared to get the eyes implanted massively undermines the value of other sources of darkvision
also pseudo goblin buff maybe since they can use these and still be fine

### DIFF
--- a/code/modules/surgery/organs/eyes.dm
+++ b/code/modules/surgery/organs/eyes.dm
@@ -178,9 +178,13 @@
 
 /obj/item/organ/eyes/night_vision/wild_goblin/on_life()
 	. = ..()
-	if (!istype(owner, /mob/living/carbon/human/species/goblin))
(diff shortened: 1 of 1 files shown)


RiceGateau:
Would be cooler if you added a medicinal mix that can be crafted by alchemy to suppress these effects, like an immunosuppressant. This would allow people to either pay or make meds for the eyes (specifically bandits) allowing it to be an option.

Overall, a good balance change that makes sense.

PR #619, opened 2025-08-22
Prevent buffs from removing built-in traits. [Bugfix]

zoktiik:
## About The Pull Request
Buffs now use their own IDs instead of "generic" for adding and removing traits. Prevents them from unintentionally removing traits you spawned in with when their effects wear off.

It's possible that I might have missed some, but this was mainly just me looking over roguebuff.dm while also checking cases where these traits get removed.

For this PR's searchability, this affects:
- Ozium
- Herozium
- Starsugar
- Warden area buff (longstrider)
- Dungeoneer area buff (pugilist)
- Noc darksight ritual
- Necra death bargain ritual
- Dendor lesser wolf rite ritual
- Eora pacifism (pacify) ritual
- Wretch Deserter "brotherhood" or "stand proud" order buff
- Martyr Ultimate Weapon traits ("Vengeance") granted during ult

## Testing Evidence
Compile-tested. There's too many effects for me to try and manually review them all, and there wouldn't be much of a point in doing so. HOWEVER, I did check these traits to make sure anything calling REMOVE_TRAIT on them wouldn't be heartbroken that they can't remove the buff.

Trait-code works by allowing a trait to have multiple "sources" that get added or removed for a given trait, and only when all of those sources are removed is the trait itself removed. See REMOVE_TRAIT and ADD_TRAIT macros.

## Why It's Good For The Game
Fixes a bug with many of these buffs where they remove built-in traits for no good reason. Ozium removing TRAIT_NOPAIN after wearing off is what prompted this, and it's a great example.

### DIFF
--- a/code/datums/status_effects/rogue/roguebuff.dm
+++ b/code/datums/status_effects/rogue/roguebuff.dm
@@ -102,11 +102,11 @@
 /datum/status_effect/buff/ozium/on_apply()
 	. = ..()
 	owner.add_stress(/datum/stressevent/ozium)
-	ADD_TRAIT(owner, TRAIT_NOPAIN, TRAIT_GENERIC)
+	ADD_TRAIT(owner, TRAIT_NOPAIN, id)
 
 /datum/status_effect/buff/ozium/on_remove()
 	owner.remove_stress(/datum/stressevent/ozium)
-	REMOVE_TRAIT(owner, TRAIT_NOPAIN, TRAIT_GENERIC)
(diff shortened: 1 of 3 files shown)


ShadowDeath6:
So do Ozium and Herozium properly work together? (Meaning, running out of one in your system won't dispel TRAIT_NOPAIN while you have the other in your system)
I assume yes since you mentioned multiple sources working properly, but I think that's probably the only test in this PR that I can imagine might throw issues.

PR #620, opened 2025-08-22
Extremely minor Lunacy Embracer balancejak

ShiningM00N:
## About The Pull Request

I changed one number for Lunacy Embracers since I've heard they can be sometimes too fragile. 

HOWEVER.

This is more of a discussion PR. I want to know if the github side of the community is fine with this minor change or not. People have also discussed that the -2 perception they get is too harsh for a speed based class. So instead, it could change it to -1 perception, but no +3 con. Or +3 con but keep -2 Per.

What do you think?

## Testing Evidence

Single number change.

## Why It's Good For The Game

antag strong gud


ShiningM00N:
> Wretches still don't have the stat buff they should have after the removal of supernatural wretches. Something close to a +1 across the board (for all "normal" wretches, like it used to be) would be really neat.

elaborate on normal wretches, which are which

PR #632, opened 2025-08-22
The great unpeelening of everything metal above neck level

Lioneliuss:
## About The Pull Request

Makes everything above iron tier that is also above neck level fully unpeelable

## Testing Evidence

too much to post, but here's the bevor
image

## Why It's Good For The Game

Is it really good? Im unsure. Did this because i was asked to.


coffeecoffeecoffee:
wh... why
this was the one thing the gorget had over the bevor that made it good and useful

Lioneliuss:
I did it because some people asked for it, i dont really care if this actually gets merged, up to the maintainers and whatnot

dawsonkeyes:
not the change to peel that we're looking for

PR #637, opened 2025-08-23
wildshape (and zad form (and bat form) (and mist form)) no longer fully heals the user

ShadowDeath6:
## About The Pull Request

it puts them in stasis mode where on_life() doesn't happen, therefore keeping wounds but nullifying oxyloss and bleeding and stuff that would otherwise kill them

## Testing Evidence

cut arm off -> wildshape -> come out of wildshape 10 minutes later -> health and blood is the same -> die from bloodloss 1 minute later because my arm is cut off

## Why It's Good For The Game

it was always a really weird way to handle the person not dying while in wildshape and I have no idea why it was done this way
this brings it slightly more in line with dnd type stuff where your wildshape is basically a shell and your original body retains the damage it took

### DIFF
--- a/code/_onclick/item_attack.dm
+++ b/code/_onclick/item_attack.dm
@@ -84,14 +84,13 @@
 	var/tempatarget = null
 	var/pegleg = 0			//Handles check & slowdown for peglegs. Fuckin' bootleg, literally, but hey it at least works.
 	var/construct = 0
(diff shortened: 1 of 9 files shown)


ShadowDeath6:
don't fullmerge yet, there seems to be a bug with werewolf transform

ShadowDeath6:
image

hopefully fixes the issue of wounds not properly healing on WW

ShadowDeath6:
vlord shapechange spells no longer heal you

PR #638, opened 2025-08-23
Team Fortess 2 [Роскомнадзор] verb

DelineFortune:
## About The Pull Request

Adds emote button that kills you

## Testing Evidence

image

## Why It's Good For The Game

Dying and loosing is fun
Its part of story and being grabbed-chained-middle mouse clicked or get red from retarded heal bot good boy yap yap *wag tail boy is not a thing that i want 
Let people die in tf2 style 

UPD: its called set name = "Pray for suicide" now

### DIFF
--- a/code/modules/mob/living/emote.dm
+++ b/code/modules/mob/living/emote.dm
@@ -1774,20 +1774,20 @@
 
 	emote("ffsalute", intentional =  TRUE)
 
-/datum/emote/living/lastwords
-    key = "lastwords"
+/datum/emote/living/suicide
+    key = "suicide"
     key_third_person = "utters their last words"
     message = ""                   
     emote_type = EMOTE_AUDIBLE
     stat_allowed = UNCONSCIOUS      
     show_runechat = FALSE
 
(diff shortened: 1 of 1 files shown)


TuskTusk:
> this seems like a button that will inevitably be exploited by bad faith actors for example if I wear someone down with boredom or frustration until they give up it's easily framed as their own choice rather than me fragging them

On the other hand, the ability to instantly respawn allows you an out out of an otherwise incredibly dull scenario. Yes, some folks may instantly unalive themselves rather than be captured, but its also not fun for jailers to deal with an indignant, silent captive who doesn't want to roleplay anyhow.

> you can literally say 'fuck' in pray to get max fire stacks lmao
That is neat, but fire still takes a while to kill you, and even if it DOES expedite the process, the 'I CANNOT DIE YET' timer is kind of in your way.


All in all, I like this. Sometimes, you just wanna stop playing a character for a round, and this allows you to dramatically pass away after doing something dumb, like jumping from a roof or getting burned on a psicross.

PR #643, opened 2025-08-23
hides smith quality

Randumb-Phantomb:
## About The Pull Request

Smithing quality no longer changes the name of an item. It still buffs item value.

## Testing Evidence

TBA

## Why It's Good For The Game

The smithing quality is a nuisance when you inspect somebody and you see a massive list of masterwork masterwork masterwork. Plus, it gives people a false idea that maybe it nerfs durability or damage or protection of weapons and armor, tricking players.


Randumb-Phantomb:
> Quality only affects durability if its masterwork

it never effects anything but value, and if it's masterwork it shines, that's all it does

KiteAE:
Not a fan of it tbh, people like seeing the little text that shows how good it is.

PR #645, opened 2025-08-23
berserker small lil +1

kiwedespars:
## About The Pull Request

gives berserker a +1 to endurance

## Testing Evidence

it's a value change if it makes the game explode that'd be fucking awesome

## Why It's Good For The Game

image

it was not the same stats as atgervi shaman which were buffed slightly statwise in #367 

comparing it to all the rest of the master unarmed classes it's pretty mediocre

- infinitely summonable katar dodge expert, miracles shaman
- 3str 3con 3end disciple. the only benefit is that it's +1speed rather than -1, but general consensus says grabs are broken so
- 3str 2con 2end 2speed templar w/ unique, better variants of unarmed weapons, alongside having miracles.

i'll give it some more love later but this should be a small little thing for now


Bobinater:
Berserker is already one of the most powerful wretch classes by a good margin, capable of 5% escape grabs and worse even against garrison statlines. In no way does it EVER need to be stronger.

PR #648, opened 2025-08-24
golem skin tones that I stole from vanderlin but I'm the guy who coded the pr that added it to vanderlin so it's not actually stealing

Randumb-Phantomb:
## About The Pull Request

Rips a butt load of aasimar skin tones from Vanderlin and adds them to the golem's color list. Replaces the marble here with the Vanderlin marble which is off-white and not pure white.

## Testing Evidence

image_2025-08-25_202655126

## Why It's Good For The Game

More color variety for golems. Don't worry, none of them are neon. Gold, rust, silver, marble, etc. If any colors need vetoing, I'm fine with that.


ShadowDeath6:
I haven't looked at any of the colors but I'd appreciate it if you gave them all a quick once-over and made sure any of them that are too human-colored be put on the doll race instead since doll needs more colors (Specifically looking at Marble, which I just removed from the material options)

ShadowDeath6:
also we have a good portion of these in game already, I'd recommend doing ingame comparisons across the material colors that we have to make sure that you're adding distinct color options
lapis = cobalt
obsidian = coal 
copper = brass or bronze, both are very close

PR #661, opened 2025-08-24
Adds THE KLAPPENLONGER

Lioneliuss:
## About The Pull Request

Adds THE KLAPPENLONGER. Its a ridiculously long, egg shaped Etruscan bascinet. Statwise exactly the same, but doesn't have detail and the visor doesn't work.
Currently only obtainable via smithing, requires Master level armorsmithing and 3 steel bars total.

## Testing Evidence

image
image
image
image
image

## Why It's Good For The Game

Its a funny looking helmet, provides no advantage whatsoever, and its not too different from the owl helmet except you can actually get this one, so why not add it?


KiteAE:
Coneheads the movie starring Dan Akroyd.

PR #664, opened 2025-08-25
Makes glaives harder to get

ConstantineII:
## About The Pull Request
- Removes glaives from drow NPCs, replacing it with bardiche.
- Removes glaives from anvil recipe
- Removes the glaive from the knight armory.

## Why It's Good For The Game
They are currently the best weapon in the game and seeing everyone and their mother run around with one gets old pretty fast.

### DIFF
--- a/_maps/map_files/dun_world/dun_world.dmm
+++ b/_maps/map_files/dun_world/dun_world.dmm
@@ -56424,7 +56424,7 @@
 /area/rogue/outdoors/mountains/decap/stepbelow)
 "siF" = (
 /obj/structure/rack/rogue,
-/obj/item/rogueweapon/halberd/glaive,
+/obj/item/rogueweapon/spear/billhook,
 /turf/open/floor/rogue/tile,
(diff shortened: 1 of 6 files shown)


ShadowDeath6:
what are the ways to get them now? if it’s too oppressive then it might just be better to rebalance them to be more in line with similar weapons

Bobinater:
> what are the ways to get them now? if it’s too oppressive then it might just be better to rebalance them to be more in line with similar weapons

Same way you get the rest of your gear, spawning or getting it made IG.

Vizendel:
If it's been removed from the crafting recipe and from the Drow who drop it, how exactly are you supposed to get it spawned or made?

arnociiroc:
you buy it from the steel category in the merchants

Aracino:
I would prefer we just balance the numbers. It being hidden behind money won't stop people from gaming the broken economy to get it in 15 minutes. Then when someone who's bad dies- a good bandit gets it and we're back to square one.

PR #667, opened 2025-08-25
changes beautiful self-examine mood bonus

ShadowDeath6:
## About The Pull Request

you get the mood bonus by looking in a mirror now, also ugly people will get a mood malus by looking at themselves in a mirror

## Testing Evidence

image

tested with ugly and beautiful and both work

## Why It's Good For The Game

flavor / realism value, if you can see your own face without a mirror then please let me know how
slight nerf to beautiful since it's not an infinite perma mood buff anymore but I added it to mirror so it's not like you're completely barred from getting the benefit of looking at yourself
potentially also a slight nerf to ugly but you really have no reason to click mirrors outside of mirror magic and this only happens if you don't have mirror magic active


PoltavskaPraca:
SOVL

ophaq:
One of the best things about beautiful was examining yourself for the mood bonus so you wouldn't go crazy during long sessions of erp. Feels kind of unnecessary to do this. Not to mention the fact there are only like, 5 mirrors on the map or something and they're locked behind rooms.

necti-dev (approved):

KiteAE:
The are not enough mirrors for this, if we had a hand mirror though... i would change my mind.

ShadowDeath6:
fine I'll add a hand mirror
not going to have it work with mirror magic though

ShadowDeath6:
image

new craftable hand mirror sprite so you can look at your sexy face mid coitus

PR #669, opened 2025-08-25
unarmed misscost adjustment

kiwedespars:
## About The Pull Request

changes misscost and releasedrain to 1 for unarmed

## Testing Evidence

numbertweak.

## Why It's Good For The Game

image

why swinging big twohanded stick cost 1 but using hands cost 5 on miss and 2 on hit.

image


kiwedespars:
> I'm fairly sure the whole reason behind it being so high was to make grab spamming less viable and could be a fault of that change. I don't know, I should probably code dive it.

oh yeah to clarify this has no impact on grabs at all
image

PR #672, opened 2025-08-25
adds racism

ShadowDeath6:
## About The Pull Request

species with stress_desc now have that desc visible near the end of your examine if you're paranoid or stressed

## Testing Evidence

image

Works properly with: 
Paranoid trait (always displays as non-same species)
Stress while not being an empath (does display)
Stress while being an empath (doesn't display)

## Why It's Good For The Game

stress_desc has always existed despite never being used and I want to make it used since it adds flavor


KiteAE:
Its time to speedmerge racism

PR #674, opened 2025-08-25
coif on face

Randumb-Phantomb:
inspired by this tiktok

https://github.com/user-attachments/assets/6095413e-e09f-4809-b1a8-37d14d2d0f21

## About The Pull Request

Lets you wear coifs in the mask slot, this includes leather, chain, and padded coifs.

## Testing Evidence
Layering is fine, thought there might be an issue with the coif going over the neck armor, but it's good.

image_2025-08-25_073416381

## Why It's Good For The Game

Less skullcracks, less decaps, more armor for people to spend on which means more work for the smiths.

### DIFF
--- a/code/modules/clothing/rogueclothes/neck.dm
+++ b/code/modules/clothing/rogueclothes/neck.dm
@@ -41,7 +41,7 @@
 	item_state = "coif"
 	max_integrity = 125
 	flags_inv = HIDEHAIR
-	slot_flags = ITEM_SLOT_NECK|ITEM_SLOT_HEAD
+	slot_flags = ITEM_SLOT_NECK|ITEM_SLOT_HEAD|ITEM_SLOT_MASK
 	blocksound = SOFTHIT
 	body_parts_covered = NECK|HAIR|EARS|HEAD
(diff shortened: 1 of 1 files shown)


coffeecoffeecoffee:
> I mean, we already have padded caps that can do that. The actual benefit I would see here is, that it would allow the neck to not being the one spot with only one layer of protection. Also I don't think that more work for the Smith is something you can sell as a good thing.

I don't think it's more work for the smith given the way the armories are set up; we are given plentiful coifs and they go unused as they are strictly worse in the neck slot than gorgets or bevors unless you're specifically out to layer head armor.

Randumb-Phantomb:
> > I mean, we already have padded caps that can do that. The actual benefit I would see here is, that it would allow the neck to not being the one spot with only one layer of protection. Also I don't think that more work for the Smith is something you can sell as a good thing.
> 
> I don't think it's more work for the smith given the way the armories are set up; we are given plentiful coifs and they go unused as they are strictly worse in the neck slot than gorgets or bevors unless you're specifically out to layer head armor.

Fair, but the iron coifs can always be replaced with something better, and so can the steel coifs.

PR #675, opened 2025-08-25
New, unique weapon option for the Cavalry knight captain

Lioneliuss:
## About The Pull Request

Adds a special glaive option for the Cavalry Captain subclass of the knight captain.
Removes the "Law & Order" weapon option for the Cavalry captain, replacing it with the aforementioned glaive.

## Testing Evidence

image
image
image
image

## Why It's Good For The Game

The Cavalry and Infantry captain had the same special weapon option, which was fine, but i decided to change that fact after someone proposed it, to add more sovl to it. Different special weapons between the two subclasses is good, just like the Marshal gets.


Lioneliuss:
did an oopsie, will return the pr in a bit..

PR #679, opened 2025-08-26
snow elves codified

Randumb-Phantomb:


BadAtThisGame302:
I'm not sure I'm a fan of Snow Elves existing today, everyone knows they were culled, Yes. Everyone knows they are the true Elvish Ancestors, Yes. 

But still keeping them around? I don't agree. 

_And at the very least, change the SKIN_COLOR_GHOST and SKIN_COLOR_GRENZEL_WOODS to SKIN_COLOR_SNOWBORN/SNOWBRED_

PR #686, opened 2025-08-26
Lenticular tattoo armor + regeneration

john-pawson:
## About The Pull Request

This PR changes a few things from the lenticular tattoos:
- Adds an armor value, so now they'll actually be able to absorb damage. 30 Blunt, 50 slash, 50 stab, 20 piercing.
image

- Adds crit prevention against regular cuts and blunt damage, so you'll need to use at least chop or smash if you want to land a crit before breaking it's integrity.
image

- Adds an integrity of 600, it'll be very hard to break, but this seemed fair considering the amount of time they take to repair themselves and the low armor protections.
image

- Adds regeneration, this means they'll slowly repair themselves when damaged. The rates are 1% (6) every 20 seconds, making it so it takes 33 minutes for it to go from 0% (0) to 100% (600). These numbers can be easily changed on the repair_amount and repair_time variables if it needs tweaking.

- Adds an icon. It's the same as the onmob, but I consider it better than the blank square on the armor slot.

**Note**: I probably wouldn't have chosen exclusively arms to give them this, considering how they can wear steel bracers just fine, and would've made it cover part of their chest too considering how it's their weakest part, but the tattoos *are* on their arms and it would probably mean changing the durability to be less, so... please give your opinions on it!! 

## Testing Evidence
The tattoos repairing themselves slowly while nothing happens:
image
Special message that appears when the tattoos finish repairing themselves:
image
Item Icon:
image
No runtimes: 
image

## Why It's Good For The Game

Previously, kashira and ruma clan had their armor (shirt if you're a female kashira) replaced by these tattoos, which did nothing other than being unremovable and prevent them from equipping armor. This PR makes the tattoos do as their description says, and helps them slightly by giving them some natural protection.


ShadowDeath6:
cool

MickDick1:
Well balance wise this pretty much leaves them as useless ever still. Nobody aims arms or hands and for good reason. At least it fixes the fact it was entirely obscuring the chest. Breasts were cut off as a result and bodyshot inaccessible even when fully nude. So, in a way it's a bug fix.

PR #689, opened 2025-08-26
The tiniest Map PR PT 1

Murderfoxen:
- changes the tavern aesthetics, and the bath-house below; going for a more lavish and gothique feel.
- changes the towncenter to have a more central noticeboard so people are really attracted to it; just behind the fountain. Removes old one to the right which blocked alleyway. -gives the keep a private noticeboard for all sorts of RP and such.

## About The Pull Request

image

image

image

image

## Testing Evidence

It will work. Trust me!

## Why It's Good For The Game

Because out of all the **_cool effort and time I put into making the map more distinct;_** 

 This is possibly the only stuff people could unanimously agree on; for now. 
 _(Yes, I am still kind of salty about it..!)_ 
 
 But don't think my druid-grove map PR's are over.. they loom in the distance..
I hope it makes everybody as happy as I can imagine and gives the map more distinction as it evolves.

### DIFF
--- a/_maps/map_files/dun_world/dun_world.dmm
+++ b/_maps/map_files/dun_world/dun_world.dmm
@@ -787,6 +787,10 @@
 "aoe" = (
 /turf/open/floor/rogue/naturalstone,
 /area/rogue/outdoors/rtfield)
+"aoh" = (
+/obj/structure/flora/ausbushes/lavendergrass,
+/turf/open/floor/rogue/grass,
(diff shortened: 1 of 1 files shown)


Rutobia:
I think the curtains are a really nice touch but they might get in the way for performers who do want to use the stage. I feel like it would make it look really cramped up there given how big the curtain sprites are.

Honestly amazing idea to put a hearth in the middle though so I think leaving the music box up on stage is fine.

PR #690, opened 2025-08-26
The Very Big Fat Juicy Vampire Lord Rebalancing

ShadowDeath6:
## About The Pull Request

vampires can no longer drain undead mobs
vampires can no longer drain NPCs
Regen costs 600 vitae, up from 300
Removes gas form spell (bat form still exists) 
Massively lowers the stats you gain from ascending (you now only get +1 to all upon reaching level 4)
Thralls no longer gain stats from Vlord ascending to maximum
Bloodsteal now behaves similarly to bite draining, meaning you can't gain from undead/bloodless/NPCs with it

## Testing Evidence

wip

## Why It's Good For The Game

there was an ultimatum put forth saying that VL would be disabled until it was properly rebalanced so I'd like to take a crack at it to preserve the flavor and soul of the role

### DIFF
--- a/code/modules/events/antagonist/solo/vampires.dm
+++ b/code/modules/events/antagonist/solo/vampires.dm
@@ -19,7 +19,7 @@
 	earliest_start = 0 SECONDS
 
 	typepath = /datum/round_event/antagonist/solo/vampire
-	antag_datum = /datum/antagonist/vampirelord/lesser //vlord once again disabled by host request for now
+	antag_datum = /datum/antagonist/vampirelord
 
 	restricted_roles = list( // basically just keep and church roles
 		"Grand Duke",


roarbark:
thanks for working on this, these changes are way overdue

i think one main problem with VL is that you can just convert as much as possible - i believe the best nerf would be that conversion would be a drain on your vitae instead of a net gain, perhaps the more thralls you have the more and more vitae would passively drain from you.

as it stands right now as a VL you will NEVER get into the situation where you would pass up on converting someone since it costs you nothing - on ratwood they gave you a huge text saying YOU DONT QUITE REMEMBER WHAT HAPPENED when you just drank from someone and dumped them - perhaps this could be added ontop of a change like that?

it would force the VL to at least choose his converts as well as starting to leave a trail of breadcrumps for people to follow instead of Sudden Army Of 1000 Darksteel Combatants

kiwedespars:
makes **all** people unable to drink from mobs.

i wanna drink bandits for water please

PR #692, opened 2025-08-26
Limit hunger loss if naked and on bath turf

tmyqlfpir:
## About The Pull Request

This PR prevents hunger dropping down to starving levels if standing on a `/turf/open/water/bath` turf and naked.

This is so characters do not drain to starving status while submerged in a bath tile, or cause that _dreadful_ stomach growling sound effect until your character eats something.

## Testing Evidence

Untitled

## Why It's Good For The Game

It'll decrease interruptions while roleplaying.

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -922,8 +922,8 @@ GLOBAL_LIST_EMPTY(roundstart_races)
 	if(HAS_TRAIT(H, TRAIT_NOHUNGER))
 		return //hunger is for BABIES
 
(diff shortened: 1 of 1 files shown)


Neliahawk:
if you are really "roleplaying" then being hungry is an "interruption" ? your character eating is part of roleplay.
being hungry makes you roleplay with the inkeep and townspeople.

if you write paragraphs for hours so long in the pool without eating that you starve to death then are you really roleplaying your character?
you can also just order food while in the bath. order some grapes or apples and feed your partner and such.

and what about thirst? sleep? if hunger is a interruption is tiredness too? grab your partner and cuddle in a bed and sleep together and continue in the morning...
these are part of roleplaying your character having a daily life arent they?

PR #693, opened 2025-08-26
naledi disicple balancejak

Lioneliuss:
## About The Pull Request

makes the "my bare hands" option for disciples set your skill to master instead of adding a +2, which led to naledi trained disciples getting legendary fists (doesnt seem like an intended feature)
## Testing Evidence

its one line brah
image

## Why It's Good For The Game

balance or whatever


Lioneliuss:
no longer necessary after the big inq port

PR #697, opened 2025-08-26
gives vulps more snout options, hairs and general layering for fluff features

arnociiroc:
## About The Pull Request

gives vulps more hairs since it was completely untouched for the longest time. 
gives them access to more snouts aswell as the current options are horribly hideous and not **every** fox is a long nosed hideous abomination.
https://en.wikipedia.org/wiki/Vulpini
adds another option for the fluff neck feature, allowing it to layer under hair but over clothing
## Testing Evidence
image

## Why It's Good For The Game
an entire species missing customization because the specifications for it were hidden at the bottom of the file are bad, another option for fluff laying is good.


roarbark:
furry cabal W

PR #699, opened 2025-08-27
Strong Intent = Heavy Wbalance

anything86:
## About The Pull Request
Adds a simple alternative check on the heavy balance check, giving any attack on Strong stance the same effect as Heavy weapon balance.

## Testing Evidence
image

## Why It's Good For The Game
Strong stance is pretty awful right now. It triples your stamina cost (5 -> 15) per hit and all it offers is +10% damage. This enables strength build to make use of heavy, hard-to-parry attacks with additional weapons at the cost of their own stamina economy.

### DIFF
--- a/code/modules/mob/living/roguetownprocs.dm
+++ b/code/modules/mob/living/roguetownprocs.dm
@@ -290,7 +290,7 @@
 
 			if(parry_status)
 				if(intenty.masteritem)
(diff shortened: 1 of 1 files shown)


necti-dev:
seems like a bad idea to be able to turn any weapon into heavy balance just like that

PR #709, opened 2025-08-27
Ravox persistance miracle auto targets most damaged limb

Fatherless1Xx:
## About The Pull Request

Ravox's t3 miracle automatically casts on the most damaged limb rather than the one chosen

## Testing Evidence

image
## Why It's Good For The Game

Quality of life, it's not easy to use this miracle effectively in combat


benj8560:
Ravox gets a really depressing lineup of ''attack'' spells. I ironically had a better time in combat as noc then ravox...ravox buffs good. Keep em coming I say.

PR #711, opened 2025-08-27
Acolytes start with a surgery bag

Fatherless1Xx:
## About The Pull Request

Acolytes begin with surgery kits

## Testing Evidence

image
## Why It's Good For The Game

The church often gets overwhelmed and this should make it easier to manage
More healing means less round removal


Plutolover99:
The church shouldn't be where most of the non-miracle healing is done. There's a clinic for that

roarbark:
church spawns with two in a locked chest already

benj8560:
> church spawns with two in a locked chest already

3 if you count the one by the crypt. 4 if you include the martyers. There is already a surplus of bags.

Just-Sim:
a surgery bag is at most 30 mammon at the smithy

PR #716, opened 2025-08-28
advanced classes for consort

Randumb-Phantomb:
## About The Pull Request

Adds 3 subclasses to consort, for each suitor subclass


Wesreidau:
I think this is a great concept if implemented well. Please note your commits clearly.

PR #719, opened 2025-08-28
Gives all MAA classes swimming 3

arnociiroc:
## About The Pull Request

Gives all MAA classes swimming, level 3.
## Testing Evidence

its a one line change
## Why It's Good For The Game
They live on the coast, they're often floundering in the moat.
There's no reason for trained soldiers(As per the MAA description) to be falling over in their own moat which they've operated with for an amount of years vs a rogue thief, who gets three levels in swimming.
Also I've seen(and am one of) so many Maa's floundering in water at roundstart just so they don't have to deal with it later. 
This would give more time for Other things, or hell... roleplaying instead of grinding stats, so they could do their job semi-decently if water was involved.


Fatherless1Xx:
Swimming remains the only real way for antagonists to escape Garrison. The map is too small, there aren’t many places to run. The wretch escaping is actually a good thing for the round in the long term, it keeps the conflict alive instead of letting it end too quickly.

PR #725, opened 2025-08-28
Acrobats drain stamina from landing falls

SylvinSeik:
## About The Pull Request

Acrobats lose 50 stamina from landing from a zlevel fall.  This is HOPEFULLY enough to add a cost, but its little enough that an acrobat can drop 2 zlevels in quick succession without stamming out.

## Testing Evidence

image

## Why It's Good For The Game

It's not nearly as good as having the trait thats literally too good for anybody but werewolves to have removed, but alas.

Now instead of being "entirely" uncatchable, acrobats will only be "mostly" uncatchable upon using their ability to descend 1 zlevel instantly, or 2 zlevels safely.  They'll at least have to wait for their stamina to regen.

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -46,6 +46,7 @@
 	if(HAS_TRAIT(src, TRAIT_NOFALLDAMAGE1))
 		if(levels <= 2)
 			Immobilize(10)
+			stamina_add(50)
 			if(m_intent == MOVE_INTENT_RUN)
 				toggle_rogmove_intent(MOVE_INTENT_WALK)
 			return
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -46,7 +46,6 @@
 	if(HAS_TRAIT(src, TRAIT_NOFALLDAMAGE1))
 		if(levels <= 2)
 			Immobilize(10)
-			stamina_add(50)
 			if(m_intent == MOVE_INTENT_RUN)
 				toggle_rogmove_intent(MOVE_INTENT_WALK)
 			return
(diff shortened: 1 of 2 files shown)


Just-Sim:
there's many ways to nerf acrobat, this one isn't it

SylvinSeik:
> because it's a virtue... and werewolves replace all ur traits for its own set of traits

The trait has existed and been exclusive to werewolves for much longer than virtues have existed.  Likewise, if it was expected to be in normal peoples hands, it would be in peoples hands, such as rogues, knaves, etc.

But whatever. if the host and maintainers don't like it currently, I'll add some stam drain to the fall to accompany the stun, as suggested on discord. Don't close it until then maybe ty

roarbark:
changing it into stamina drain fixes the issue of people using it for chases. anyone climbing down after you knows which way you went and knows you couldn't have sprinted that far because half of your green is gone...

this also still lets me use it for flavour which im happy about because All The Things I Like Get Nerfed usually

SylvinSeik:
> That said? Gutting it entirely though by removing the singular thing that made it worthwhile(zfall damage removal) is the equivalent of just removing it. Very weird. Not a fan. The stamina impact on use is acceptable, I'd argue, because it makes sense. It really isn't the biggest thing in the room that needs a change, however. I can only imagine what prompted the PR.

Thank you for typing this comment after I had already changed the PR.

But to answer your question, I was informed that leaper does a few funny little tricks that most people aren't aware of, and it was the straw that broke the camels back, so I PR'd it.  I've not had any issues with acrobat users in the past few days, but acrobat is a "the only people that don't hate it are the people that use it" virtue, so of course I was willing to do this since it was so easy.

PR #727, opened 2025-08-28
slurbow removal

notafurryy:
## About The Pull Request

Removes the slurbow

## Testing Evidence
image

## Why It's Good For The Game
The slurbow is actually disgusting when used by knave-ish builds, it can penerate armor as well as the normal crossbow just deals less damage, which doesnt matter if you can spam them out without ever being in range of the person, Also since it's a crossbow the aim time is nigh instant at no stamina drain.


benj8560:
slurbow is like the only good offensive thing the confessor gets though..... ;-;

notafurryy:
> slurbow is like the only good offensive thing the confessor gets though..... ;-;

the slurbow is way too powerful though, it has more opportunites to deal crits than a normal crossbow, deals more damage than it, penetrates all armor like it and you can reload it while moving ontop of the reload time being almost instant.. Not to mention the 0 stamina usage it has

PR #728, opened 2025-08-28
Psydonite nobles

Randumb-Phantomb:
## About The Pull Request

Removes the allowed patron: NON_PSYDON_PATRONS from every role it is mentioned in. Keeps the define around in the code just in case.

## Testing Evidence

image_2025-08-28_140504611

## Why It's Good For The Game

If the inhumen are (mechanically)allowed to be all of these roles, so should the Psydonites. You can create a guy named "Richter Von Evilstein" have him worship Zizo and queue for duke, yet Psydonites are barred.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/courtier/councillor.dm
+++ b/code/modules/jobs/job_types/roguetown/courtier/councillor.dm
@@ -9,7 +9,6 @@
 	disallowed_races = list(
 		/datum/species/lamia,
 	)
-	allowed_patrons = NON_PSYDON_PATRONS
 	allowed_sexes = list(MALE, FEMALE)
 	display_order = JDO_COUNCILLOR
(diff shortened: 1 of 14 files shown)


SFox63:
So there's some update to the picture above with a humen. I still think you should be able to play a Royal as Psydonian. There's no excuse. You can be a Zizite Duke which makes no fucking sense, why can't I play a Psydonian Duke?
image

Kregle:
> Inhuman are allowed to be nobles because inhuman are generally expected to hide their identity and who/what they worship, which is not an expectation Psydonites carry.

Frankly, that's a rather stupid excuse. Only Tennites are allowed nobility per the rules of the Church. I'm sure Psydonites are quite capable to hide their conversion - despite being a quite boisterous, relatively accepted faith - to maintain their power. If a Graggarite or Matthiosite Duke is able to get to power, so too should a Psydonite. If we're looking at analogues for a faith that considered hiding their faith somewhat sinful- Christians in the Roman Empire engaged in clandestine worship. 

If there's an outwardly Psydonite Duke, the Church should declare them a heretic, simple as.

PR #729, opened 2025-08-28
does nothing at all

chemistrymain3:
## About The Pull Request

removes acrobat entirely

## Testing Evidence

works
## Why It's Good For The Game

virtues aren't and weren't originally designed to give you advantages- they're meant to be cute little flavor bits for your role that fit your character, things to build RP off of, and if they incidentally give you slight advantages, that's okay

the majority of virtues are hit and miss- some are stronger than others, like blacksmith's apprentice, and ultimately i do not care too much about balancejakking

however acrobat is a problem because it gives you A Lot of free shit in exchange for Absolutely Fucking Nothing, and it became an obligatory combat virtue because it's just good at everything- it lets you get away pretty effortlessly, ignore 2z falls (by climbing down 1z), and so on so forth blah blah blah

goodbye. alternative to the nerf pr, which is well intentioned and also being brigaded

### DIFF
--- a/roguetown.dme
+++ b/roguetown.dme
@@ -2454,7 +2454,6 @@
 #include "modular_azurepeak\statpacks\wildcard.dm"
 #include "modular_azurepeak\virtues\combat.dm"
 #include "modular_azurepeak\virtues\items.dm"
-#include "modular_azurepeak\virtues\movement.dm"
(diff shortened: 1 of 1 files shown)


dawsonkeyes:
going forward with #725

PR #733, opened 2025-08-29
properly paths stamina poison / makes it lower stam

ShadowDeath6:
## About The Pull Request

it used to lower energy, I changed it to lower stam to be consistent with the name
I will say that this absolutely obliterates your stamina bar at lower END levels (at END 9, 2u of stam poison will take half of your bar, and 2u of strong stam poison will exhaust you) so I'd prefer it get TMed to test numbers and make sure it's not just "I poison u to death and u die"

## Testing Evidence

image

## Why It's Good For The Game

it's a bugfix because the poison is otherwise completely nonfunctional, and also I think it makes it consistent with its name since it's called stamina poison and not energy/mana poison


Heheniy:
you are adding energy, not removing

PR #739, opened 2025-08-29
Ports Weapon Shaft Removal from Azure

anything86:
## About The Pull Request
Ports https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/3551
Some values have been changed where the codebases diverge but it should generally be faithful.

## Testing Evidence
image

## Why It's Good For The Game
It hugboxes the shafts.
Shafts SUCK. Your weapon breaks in 6 slaps if it has the wrong shaft against the weapon you're facing. This MAY be better.


FreeStylaLT:
this works fine but you guys got different weapon balance vibes so don't be afraid to do your own math, this one has a particularly soft curve that even I would like to change later
(damfactor + strfactor steps down to 0 as it reaches 25%)

so you don't really "notice" any difference for penetrative weapons until it chunks off at 25%, as it still "penetrates" for the most part

PR #744, opened 2025-08-29
adds normal bone as a grinding recipe

ShadowDeath6:
## About The Pull Request

it gives bonemeal (less than tailbone grind does)

## Testing Evidence

image

works as expected

## Why It's Good For The Game

allows people to get bonemeal from bones, which I think makes sense
tailbone already exists but it has its own uses and isn't as plentiful so it's not really that enticing to grind up
also you can just get bonemeal from fish infinitely without any combat so it's not really a matter of bonemeal being difficult to access


necti-dev (approved):

ShadowDeath6:
ermm this isn't conflicting THO

PR #749, opened 2025-08-30
struggler rework

Randumb-Phantomb:
## About The Pull Request

changes struggler statpack to be called brutish
changes struggler statpack to give -2 to SPD, PER, and INT, instead of -3 to INT and PER and -2 to FOR.

## Testing Evidence

TBA

## Why It's Good For The Game

TBA


Xaytan:
This just makes it Better Muscular and pretty much the perfect statpack for Big Strong Guy.

Also, Struggler is pretty obviously a Berserk reference? Guts was never slow, but he was incredibly unlucky.

PR #757, opened 2025-08-30
Knife(lock)picking of windows

PoltavskaPraca:
## About The Pull Request

+ It is now possible to knifepick windows from the other side, enabling those with a lockpicking skill to open the windows without having to bust them the fuck down, as well as close them.
+ Speed and success chance are dependent on the lockpicking skill, PER and SPD.
+ With each passing check there is a chance to fail. The higher the skill and the relevant stats, the less checks you have to pass at the higher speed. The failstate is a broken window and a loud sound of it being broken.
+ Daggers have multipliers attached to them. Hunting knife is 1.0 multiplier, steel dagger is 1.1, Du Tace is 1.2 mult, Cleaver is 0.8. Throwing knives are 0.95, for example.
image

## Testing Evidence

https://github.com/user-attachments/assets/7c089008-5d97-4475-a149-5531fbb41da2

## Why It's Good For The Game

Rn thieving while someone is in the building below you is straight up impossible unless there is a door you can lockpick on the higher floors. Smashing the window is simply not an option, save for the exit.

### DIFF
--- a/code/game/objects/structures/roguewindow.dm
+++ b/code/game/objects/structures/roguewindow.dm
@@ -288,7 +288,7 @@
 					playsound(src, break_sound, 100)
 					log_admin("Window broken at X:[src.x] Y:[src.y] Z:[src.z] in area: [get_area(src)]")
 					loud_message("A loud crash of a window getting broken rings out", hearing_distance = 14)
-					new /obj/item/natural/glass/shard (get_turf(src))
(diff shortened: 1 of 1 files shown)


sippykot:
this just makes it so now EVERY window is shit and needs to be bricked up immediately
instead of just most of them

PoltavskaPraca:
Have you both considered not being little powergaming gremlins?

PR #758, opened 2025-08-30
Atgervi: Back to the Roots

Just-Sim:
## About The Pull Request
The Atgervi have completely lost their identity, they're supposed to be inhumen-following mercenaries from the Gronn Highlands (Vikings) who act **together** for best results. Both classes can fight on their own, sure, but they'll struggle against more than 1 opponent.

The Varangian is your warrior-viking, they can use bows and axes, and will decimate in battle with its axe, but will struggle against quick opponents.

The Shaman is your barbarian-cleric, they use their fists and grapples, using miracles to buff their Varangian. They **struggle** against anything other than a 1v1 due to their lack of armor.

## Testing Evidence

image

## Why It's Good For The Game

Because I just don't like the changes that have been made and would like to see the class I play actually be what I chose it for.


pengisbungholius:
if you want to make shaman a support class that isn't able to handle many opponents at once, drop the unarmed to 3 or 2 since successful grappling should remove parry/dodge well enough for 1v1s? +3 str +1 speed AND unarmed skill at 4 makes them speedier than iconoclast (with 1 less skill point in unarmed, but mostly comparable stat spread otherwise).

also they get strong bite (which is rather crazy if your opponent has uncovered flesh) for 1v1s

Dagonlives:
This was a disappointing P.R for me. I was playing the vagarian precisely because it was a holy skill viking badass. Now it's just a really subpar man at arms.  How is bow skill useful? Where am I going to keep my satchel between my shield, axe, arrows and bow?  This 'bck to basics p.r' just ensured I won't play what was my favorite mercenary.

PR #770, opened 2025-08-31
Map Fixes

Murderfoxen:
- changes curtains to be closed at roundstart because it looks nicer. >:|
image

- adds a noticeboard that was missed by @ConstantineII (north-east gate)
image

- removed an additional noticeboard that is found near the new central one. (cleans up the alleyway like the beforetimes..)
image

- fixes a tile issue with the bath-house that I missed in my sleepy mapping. u-u my biggest shame
image
image

### DIFF
--- a/_maps/map_files/dun_world/dun_world.dmm
+++ b/_maps/map_files/dun_world/dun_world.dmm
@@ -6532,7 +6532,9 @@
 /obj/structure/closet/crate/chest/crate,
 /obj/machinery/light/rogue/torchholder/c,
 /obj/structure/curtain/magenta{
-	color = 54202a
+	color = 54202a;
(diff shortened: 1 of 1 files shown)


Aracino:
If you could, there's also some invisible walls on the north most pews in the charge. On the like, 2nd section of the pew from the left, there's an invis wall that blocks north and south movement. I believe it's also there on the top right pew.

Murderfoxen:
> Yeah the church has a lot of weird issues. There's a black square in the dining area, the pews are all wonky and reversed, the borders are often messed up.

this and the above comment; there doesnt seem to be anything indicating this and it sounds like it might be an issue with mapgen at the start of rounds causing these anomalies.

PR #777, opened 2025-08-31
knight balancejak

Lioneliuss:
## About The Pull Request

mostly just cuts the mage level intelligence buffs knights got, adds one more int to both knight captain subclasses and slight stat alterations otherwise, review and decide whether its good or not

## Testing Evidence

changed a few numbers, is it really nescessary?

## Why It's Good For The Game

balance i guess


Aracino:
Isn't the most common complaint at the +4 int should be removed- not something that should be matched

pengisbungholius:
if you're balancing stats you should probably keep in mind that str/spd = 2 stat points, all others = 1, you have several instances of replacing 2-3 int with... 1 con and/or end, which does not add up 

additionally, just ADDING int to knight captain seems unreasonable? you're not taking away anything from them as well (AND they get unique equipment... might be overloading them with stuff?)

also, lastly, int is the feinting stat: I assume knights have insanely high int to represent how they're trained professional soldiers who have the knowledge and free time to learn how to feint people, they're kind of loosing out on their gimmick of being feint machines with this pr? since they're now losing ALL their int and are now on par with footmen and cavalrymen MAAs stat boost-wise

Randumb-Phantomb:
It should be 2 INT on ALL of them, these are educated knights, not slightly above average. 10 INT is the kind of shit commoners get.

roarbark:
whats the real, in game problem with knights having 14 INT? if this is an issue of flavour, the mages get the arcyne trait and reading skill to signify that they're smarter at academics

sippykot:
knights have hight int to feint

Lioneliuss:
*fart

PR #778, opened 2025-08-31
Restore and rebalance poison knives

tmyqlfpir:
## About The Pull Request

This PR returns the poison knife behavior, and attempts to balance it to be more fairer.

What has changed:
* The knife will not poison the target if it has hit armor
* On armor hit, there is a 20% chance that the knife will lose it's poison coating
* Brute/Lashing/Twist/Punch intents will not inject the target
* The knife can be cleaned to remove the poison
* More text feedback has been given (although successful inject will NOT display text to the player)
* Players with `TRAIT_LEGENDARY_ALCHEMIST` will be allowed to double dip the blade, allowing it to store and inject 2 units of reagent
* When slicing food with a contaminated knife, the reagents will be transferred into all of the the slices (or single slice) and be cleaned
* Throwing will NOT inject other characters

## Testing Evidence

I have tested this locally and confirm that the above points are accurate to the implementation.

## Why It's Good For The Game

This adds emergent gameplay to poisoned knives and forces the attacker to wear down the armor of the target, before they can get a strike in.

### DIFF
--- a/code/modules/food_and_drinks/food/snacks.dm
+++ b/code/modules/food_and_drinks/food/snacks.dm
@@ -535,7 +535,7 @@ All foods are distributed among various categories. Use common sense.
 			var/obj/item/reagent_containers/food/snacks/slice = new slice_path(loc)
 			slice.filling_color = filling_color
 			initialize_slice(slice, reagents_per_slice)
-			if(add_reagents_from_knife)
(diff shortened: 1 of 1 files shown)


ShiningM00N:
can you still dip knives in emberwine.

PR #779, opened 2025-08-31
Removes Dungeon keys from guards, knights and councillors

Vizendel:
## About The Pull Request

Tired of guards interrupting your dungeon (e)rp? This will fix it.

## Testing Evidence

It is just one line changes, should require no testing.

## Why It's Good For The Game

Guard players deserve less.

### DIFF
--- a/code/game/objects/items/rogueitems/keyrings.dm
+++ b/code/game/objects/items/rogueitems/keyrings.dm
@@ -215,25 +215,25 @@
 	keys = list(/obj/item/roguekey/farm, /obj/item/roguekey/butcher)
 
 /obj/item/storage/keyring/sheriff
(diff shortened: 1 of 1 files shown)


benj8560:
councillors maybe. But like...these people need to be able to access to the cells. Its their job to fill them. ERP is definitely a last priority to when it comes to balance assuming you are not just joking (if you are raise some real points for this).

sippykot:
> councillors maybe. But like...these people need to be able to access to the cells. Its their job to fill them. ERP is definitely a last priority to when it comes to balance assuming you are not just joking (if you are raise some real points for this).

councillors are theoretically supposed to be lawyers so them having dungeon keys isnt really out of the place
this whole pr is kinda dumb, just fuck people in your bedroom as dungeoneer

PR #783, opened 2025-09-01
Deathless changes + if you leap/jump/snore no UH!! sound + a little shoddy jump emote bugfix

PoltavskaPraca:


Randumb-Phantomb:
I have a very strong opinion on this. A lot of this seems a bit railroady. Not everyone plays their deathless character as a... vampire, or whatever you're suggesting they are now. Why the fuck do they get deadite immunity? They're not a deadite already, and the rot doesn't care if you're dead or not, you can be a walking unbreathing vaaaaampire for all it cares and it will still infect you. Also restrict this from any role that gets silver equipment, for obvious reasons. It's death_less_ they're not dead, nor undead, they're _inwardly_ unliving. On the outside, they're fine, they shouldn't be burning at the touch of silver or exploding from miracles.

Fatherless1Xx:
This should be a vice rather than a virtue because the negative far outweighs the positive

notafurryy:
>  You can't die from bloodloss

You still get blodloss debuffs and you die from it

Randumb-Phantomb:
> Stop talking outta your ass please. What does rot have to do with deadite infection? 

'The rot' is the disease that causes people to become deadites.

Xaytan:
This prevents lux revival.

PoltavskaPraca:
> This prevents lux revival. Is this supposed to prevent lux revival? The only way to revive deathless right now appears to be either the funny revival chair or brain transplants.

I'll test lux revival on local but as far as I'm aware, they are not correlated

PoltavskaPraca:
> This prevents lux revival. Is this supposed to prevent lux revival? The only way to revive deathless right now appears to be either the funny revival chair or brain transplants.

you are correct, sorry for having missed it. Didn't know you actually can't lux undead mobtypes

PoltavskaPraca:
alright so. I'm reverting any changes to the deathless and instead adding a new virtue. The bugfixes and other stuff is remaining in this PR.

PR #784, opened 2025-09-01
de-nimrodification of code

ShadowDeath6:
## About The Pull Request

nimrod doesn't mean dumb ok

## Testing Evidence

didn't test but it compiles

## Why It's Good For The Game

image

### DIFF
--- a/code/datums/special_traits/traits/traits.dm
+++ b/code/datums/special_traits/traits/traits.dm
@@ -348,12 +348,12 @@
 	character.set_patron(/datum/patron/godless)
 
 //negative
-/datum/special_trait/nimrod
-	name = "Nimrod"
+/datum/special_trait/foolish
+	name = "Fool"
 	greet_text = span_boldwarning("In the past I learned slower than my peers, and I tend to be clumsy.")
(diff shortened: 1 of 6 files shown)


thehighpriestofhydra:
"uhh, but nimrod doesn't mean stupid!"
Okay; but "no shit, Sherlock/Einstein/(insert genius here) is a thing, it auto filters because it's intended to be ironic, in the same way you would call something stupid someone said a "bright idea".

And indeed, that's how it first originated as an insult in popular media - Buggs Bunny calls Elmer Fudd (a shitty hunter) "Nimrod" (a great hunter) as an example. Saying "nimrod doesn't mean stupid so it should be removed" is silly.

ShadowDeath6:
nimrod isn't even tangentially related to "stupid", unlike sarcastically calling someone einstein / sherlock
also I doubt Nimrod would even exist as a biblical figure in psydonia's canon given it's not real life so it makes even LESS sense to call someone that

sippykot:
using this word like that is literally just a bugs bunny reference and completely out of place for the setting

PR #787, opened 2025-09-01
homesteader port

Lioneliuss:
## About The Pull Request

ports https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/3172

## Testing Evidence

it works, i tested.
image
image

## Why It's Good For The Game
more stuff stolen from azure yippie


necti-dev:
Awful feature

FreeStylaLT:
> not really a fan of this, I can't really tell why this would be good for the game baseline and all I can see it doing is detracting from other towner roles because it can just level up better than them the original PR on azure was 10:19 downvoted btw

That was because the PR initially removed nearly every other towner role (Except Hunter, Barber & Witch iirc) so this one would replace them. They're better than any other towner barring those exceptions because they were meant to replace them all, but people liked the idea of playing a guy named "Cheesemaker" and doing nothing but making cheese for 3 hours, so the other roles were conceded to remain.

Either way though, up to you guys. The names are funny, and I don't think any individual towner role can fill 3 hours of round content on its own, hence why I opted for a sandbox role that can bounce between activities and has all the gear to do anything they'd want.

PR #788, opened 2025-09-01
lunacy embracer faith adjustments

Lioneliuss:
## About The Pull Request

makes lunacy embracers dendorite or graggarite only
warns you about the change in the role description
spawns them with a dendorite amulet on the neck

## Testing Evidence

https://github.com/user-attachments/assets/77b2f1a8-b41e-4d6a-a8e8-bd91c4f2c8f4

## Why It's Good For The Game
makes the roles description true
shouldve been done a long time ago if im being honest, OR changed the roles description

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/wretch/lunacyembracer.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/wretch/lunacyembracer.dm
@@ -20,6 +20,8 @@
 	ADD_TRAIT(H, TRAIT_WOODWALKER, TRAIT_GENERIC)
 	ADD_TRAIT(H, TRAIT_HERESIARCH, TRAIT_GENERIC)
 
+	neck = /obj/item/clothing/neck/roguetown/psicross/dendor
+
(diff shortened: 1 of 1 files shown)


zVineatic:
I think some deity limitations are okay with some roles, exactly for this one. Can you imagine an astratan lunacy embracer? I can't. But limiting the role to a SINGLE deity is a bit too much I think. I would say they are fine with being graggarites alongside dendorites as well.

CandleJaxx:
let 'em be abyssor or dendor, locking 'em feels silly

pengisbungholius:
probably an overall good change as is though, a malum worshipping lunacy embracer would be quite strange? really just about no matter how you swing it.

lunacy embracers get woodwalker and legendary(?) farming (both seemingly because of the implied denor worship), they should probably remain locked to dendor or have whatever whitelisted gods change that to a more fitting skill (like, adding fishing 2 if you're an abyssor lunacy embracer) instead of just allowing gods in a list without changing or checking for what the player picks

Lioneliuss:
fine ill add fraggar

roarbark:
lock them, they'll be absolutely awful to balance should we ever touch any miracles at all

PR #793, opened 2025-09-02
Recall changes

Lioneliuss:
## About The Pull Request

Makes recall require an open mouth to cast (no actual spell, i couldnt get it to work proper)
While casting recall, you now emit smoke. Why?.. Because you disappear in a big cloud of smoke i guess, the real reason is making it more obvious youre casting it

## Testing Evidence

https://github.com/user-attachments/assets/1788aa10-07dc-4bff-bf30-dde4714ee30b

## Why It's Good For The Game
Balance, recall used to be a free get out of jail card with very little signs youre casting it besides a small message in chat and a progress bar, and also made it physically impossible to capture any mage that had it, as if you left them for literally 15 seconds they would be GONE.

### DIFF
--- a/code/game/objects/effects/temporary_visuals/miscellaneous.dm
+++ b/code/game/objects/effects/temporary_visuals/miscellaneous.dm
@@ -467,3 +467,36 @@
 	else
 		return INITIALIZE_HINT_QDEL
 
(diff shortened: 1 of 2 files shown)


Lioneliuss:
screenshot provided as the video may not load
image

PR #794, opened 2025-09-02
Meet new bandit class: the pioneer 

DelineFortune:
## About The Pull Request

Adds support bandit class
They can dig they can kill you with their strong saperka shovel (it has low speed so dont stay on one place to avoid getting hit you retard)
They can place traps
Stupid people cannot disarm traps
Bandits/Wardens can see placed traps well
Other ppl can barely see it
Bandits/wardens/vl/lich/wretch or **any person without armor training** or **dodge expert** cannot trigger those traps
If you dont want to get hit by it dont pick combat classes and travel across antags lands???

## Testing Evidence

image

## Why It's Good For The Game

coz 95% of our prs are:
cool unique balance jacks upgrades like ''remove antag class or numberfuck it''
9 of 10 antag design are 'rush to target click them until horisontal' so game design, so cooooooooooooooool


notafurryy:
bandits aren't really the kind of antag that can fortify stuff, it's mostly highway robbery and smash and grab stuff, you aren't sticking around long enough to build stuff or set up traps, unless it's a combat door wall to block off access which is ass

S-Z-F:
> Strange WW2 german officer looking coat... Also, why can specific people just walk through the traps and not trigger them? I understand bandits, but why liches or wardens?
> 
> Another thing - "Saperka" is a Polish name for a modern, foldable shovel-multitool. That shit didn't exist until about 1860.

As i was gonna add before you blocked my comments over any signs of criticism - the idea doesn't seem bad, but we would have to actually put it in a round to see if its good or not. Maybe adding a timer for the traps before they dissapear would be a good idea? That way bandits wont just leave these around randomly in bad faith to frag people remotely with their invisible trap.

Just-Sim:
> > Strange WW2 german officer looking coat... Also, why can specific people just walk through the traps and not trigger them? I understand bandits, but why liches or wardens?
> > Another thing - "Saperka" is a Polish name for a modern, foldable shovel-multitool. That shit didn't exist until about 1860.
> 
> As i was gonna add before you blocked my comments over any signs of criticism - the idea doesn't seem bad, but we would have to actually put it in a round to see if its good or not. Maybe adding a timer for the traps before they dissapear would be a good idea? That way bandits wont just leave these around randomly in bad faith to frag people remotely with their invisible trap.

did u make an alt just to comment on this post??? how much free time do u have??

PoltavskaPraca:
Looks cool and interesting. Saperka name is a bit too on the nose, for me personally but u know why.

Gorilla warfare is real.

DorfGuy1:
We already have a trap system that's just criminally underused. Letting the bandits get a special snowflake way of making them just feels lame. Expand the systems we already have.

pengisbungholius:
something else I'd just realized, the snowflake interactions between wretches and wardens with the traps seems a little odd:

what if wretches DON'T side with bandits what if they side against them? what about when a guard/merc/whatever sides with bandits? rontz rings have to be put onto people already, shouldn't it check for the bandit scomstone instead of checking for roles? or is that not possible? it would be quite lame to not only get hurt by your new allies, but to also be actively ruining the trap placement by walking onto/being dragged onto your new bandit ally's traps.

and, not sure if the duke/whoever else job change thing sets them to be a warden for the effects of the trap, but it might also be a chafing point for the intended effect (of dissuading MAA/knight/squire/adventurer/adv+ roles from charging towards bandits alone: what if adventurers could just get retitled to warden to dodge traps?)

DelineFortune:
> something else I'd just realized, the snowflake interactions between wretches and wardens with the traps seems a little odd:
> 
> what if wretches DON'T side with bandits what if they side against them? what about when a guard/merc/whatever sides with bandits? rontz rings have to be put onto people already, shouldn't it check for the bandit scomstone instead of checking for roles? or is that not possible? it would be quite lame to not only get hurt by your new allies, but to also be actively ruining the trap placement by walking onto/being dragged onto your new bandit ally's traps.
> 
> and, not sure if the duke/whoever else job change thing sets them to be a warden for the effects of the trap, but it might also be a chafing point for the intended effect (of dissuading MAA/knight/squire/adventurer/adv+ roles from charging towards bandits alone: what if adventurers could just get retitled to warden to dodge traps?)

will check after tms quality of life going to be added later

>are they supposed to get more bows/crossbows skill than knaves? knave's code comment says it's supposed to be the sneaky ranged class, and they get specialized into crossbows OR bows, while the pioneer gets 5 in each?

>also WOW they get like 18 stat points? I assume this is because of their lack of dodge expert/armor training? should that be code commented so that it's clearer for the future?
for reference: the iconoclast gets 12, and has heavy armor training; the knave gets 14 and has dodge expert - its not that critical. 

They have no speed no dodge and light armor only - there literally nothing they can do again any 2 tiles weapon 
they are too addicted to other bandits/their traps

necti-dev:
* This is really not fitting with the medieval fantasy theme, as implemented. I understand you enjoy modern warfare servers, but this shouldn't be one.
* The `has_exempt_role()` proc is bad. This is what traits are for.
* Same with `has_required_trigger_trait()`. You add an immunity trait for this kind of thing. Which is what the other proc does. The mixture of both is just confusing.
* Same for `is_exempt_viewer()`
* The skill spread and stats of these guys are really inflated

PR #796, opened 2025-09-02
"freeman" accent

Randumb-Phantomb:
## About The Pull Request

adds a freeman accent, which is carefully crafted to be role-versatile. A bandit, knight, or duke could take this accent and not be too out-of-place.
It's based off of helldivers obviously.

## Testing Evidence

image_2025-09-02_110716587

## Why It's Good For The Game

This is hilarious and it's not too setting-breaking.


roarbark:
seems unfitting

PR #802, opened 2025-09-03
Lunacy Embracer additions which make it no longer just worse berserker (hopefully)

AdvilPMSevere:
## About The Pull Request

Adds the following to Lunacy Embracer

- Steelhearted trait
- Skill 4 in athletics, from 0 (wtf)
- Skill 3 in swimming, from 2
- Skill 3 in sneaking, from 0
- Skill 2 in butchering, from 0

## Testing Evidence

Dream Daemon would not work for some god forsaken reason but otherwise I would have loaded it up. Adds a few lines directly from the berserker code, should work.

## Why It's Good For The Game

This brings Lunacy Embracers in line with other wretches in terms of the steelhearted trait. Gives some common sense skills to the fellas which they lacked seemingly as an oversight, without giving them anything that berserker didn't already have.


AdvilPMSevere:
Also, didn't include holy skill because of the other pr about that currently open.

ShiningM00N:
Since you're already doing this, I'll close my PR adressing these kind of changes.

I do want to say, how does the idea of giving them polearms 2 base, 4 if they choose it sound? Of course they get lesser unarmed if they do choose that.

AdvilPMSevere:
> No, this class does not need another buff.

Ok, but this should really not even be considered a buff in the traditional sense. This just adds things every other wretch has that lunacy embracer doesn't for some reason. If this is too much than why can berserker have all the above skills and more without any of the downsides of lunacy embracer?

PR #816, opened 2025-09-04
Gives Lunacy Embracers the Steelhearted and Beastly Digestion Traits

AdvilPMSevere:
## About The Pull Request

As the title says, gives Lunacy Embracers steelhearted and allows them to eat raw food and drink murk.

## Testing Evidence

Two lines.

## Why It's Good For The Game

As people whose whole thing is being insane and living in nature, it only makes sense they are able to sustain themselves without "civilized" things like cooking and well water. It also doesn't make much sense that those who embrace the natural world so wholly would be affected by things like dismemberment and other brutality. The added traits will rectify this and let lunacy embracers actually reject civilization like they're meant to.


Xaytan:
Return to monke

PR #819, opened 2025-09-04
Ports Slick trick, a spell which creates a 3x3 slippery area

E231kei500bandai:
## About The Pull Request

Ported from https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/1803

## Testing Evidence

https://github.com/user-attachments/assets/484170e7-06c7-4641-8e95-f5a56faec061

image

## Why It's Good For The Game

It's a funny spell that is powerful but can't kill someone on its own.

The spell can also encourage teamwork, as you could use it as a defensive spell to protect your allies or an offensive spell to slip your enemies.

It also has a fairly long cooldown so it can't be spammed too much.

### DIFF
--- a/code/game/turfs/open/_open.dm
+++ b/code/game/turfs/open/_open.dm
@@ -113,8 +113,8 @@
 		if(buckled_obj)
 			buckled_obj.unbuckle_mob(C)
 			lube |= SLIDE_ICE
-		if(lube&SLIDE_MAGIC) //send them flying
-			new /datum/forced_movement(C, get_ranged_target_turf(C, olddir, 7), 1, FALSE, CALLBACK(C, TYPE_PROC_REF(/mob/living/carbon, spin), 1, 1))
+		if(lube&SLIDE_MAGIC)
+			new /datum/forced_movement(C, get_ranged_target_turf(C, olddir, 6), 1, FALSE, CALLBACK(C, TYPE_PROC_REF(/mob/living/carbon, spin), 1, 1))
 		else if(lube&SLIDE)
 			new /datum/forced_movement(C, get_ranged_target_turf(C, olddir, 4), 1, FALSE, CALLBACK(C, TYPE_PROC_REF(/mob/living/carbon, spin), 1, 1))
 		else if(lube&SLIDE_ICE)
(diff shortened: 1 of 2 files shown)


chaoko99:
this is for a bounty, I wanted this to be a spell for wizards because A: I think it's funny
and B: They don't have enough spells that are not directly killing someone or making someone better at directly killing someone. It goes both ways, after all.

E231kei500bandai:
balance update, it sends you flying, but no longer stuns you for extended amounts of time. this makes it super annoying and funny but not a death sentence

https://github.com/user-attachments/assets/93252ac3-bf42-4281-a7ce-b004169cca08

E231kei500bandai:
people skidding off a roof to their death is hilarious how could you be against that

Making it have an activation period could make it more skilled
Like for the first 5 second it can't slip anyone

E231kei500bandai:
also keep in mind the range is low, you can't spawn a slip on someone unless you're already in their vision range

pengisbungholius:
> people skidding off a roof to their death is hilarious how could you be against that
> 
> Making it have an activation period could make it more skilled Like for the first 5 second it can't slip anyone

it is funny, that's why I'm saying it should be locked off away from frag-centric roles, since it'd VERY quickly become unfunny when unfunny-"gimmick"-abuser # 1,523 decides that for the next week they'll run the spell into the ground. 

a long activation period maybe, but the strength of it is still there... if someone's just spamming it on a gate's/north stairs' threshold it'll slip people and allow you to gank them with a bunch of wretches/bandits

> also keep in mind the range is low, you can't spawn a slip on someone unless you're already in their vision range

doesn't really change too much for combat or escaping? if you slip someone as you go in an L shape and they continue on straight, it's just a free escape tool on TOP of being a z level funny spell, and an easily top tier spellblade-y combat tool. 

this is ss13, players WILL run your fun funny addition into the ground, they WILL codedive for exploits and exploit them, they WILL ruin the game for everyone

putting the spell behind a role, like court magos or a new role that ISN'T able to attack people easily (meaning like, probably full on pacifism or something so they can be FUNNY and not completely meta) would give it a much better reason to ever be added

E231kei500bandai:
Now with a 10 second effect and a 45 second recharge and a 4 second charge, this gives you 39 seconds of time inbetween spells so it shouldnt be spammable in front of a door

PR #831, opened 2025-09-05
Papier

EvenInDeathIStillServe:
## About The Pull Request

removes the broken background from scrolls
adds a title bar to paper so you can close it from there instead of trudging through the text itself to find the close button. and you can move the window around easier now

adds a smash intent to the tome of justice so u can lay down the law. literally

## Testing Evidence

image

image

## Why It's Good For The Game

don't you argue with me


RyszardParys:
Why did this get closed? It was peak :(

PR #832, opened 2025-09-05
unquisition racelocks

Lioneliuss:
## About The Pull Request

changes the racelocks on the inquisition to the same ones as the knight ones. Clearly the previous ones were made by an evil azurite

## Testing Evidence
 ඞ ඞ ඞ ඞ ඞ

## Why It's Good For The Game

racism i mean better racelocks

### DIFF
--- a/code/modules/jobs/job_types/roguetown/Inquisition/absolutionist.dm
+++ b/code/modules/jobs/job_types/roguetown/Inquisition/absolutionist.dm
@@ -5,7 +5,7 @@
 	faction = "Station"
 	total_positions = 1 // THE ONE.
 	spawn_positions = 1
-	allowed_races = RACES_SECOND_CLASS_NO_GOLEM
+	allowed_races = RACES_NOBILITY_ELIGIBLE_UP
 	disallowed_races = list(
 		/datum/species/lamia,
 	)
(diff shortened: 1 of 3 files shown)


Randumb-Phantomb:
Allow golems you goof

AdvilPMSevere:
> Being unbias: There's no reason this should be more strict then templars/priest. The hosts specific concerns are around the noble court of the keep.
> 
> Also golems are highly suspicious beings being powered by Lux and towing the line of necromancy. I don't think think the Inquisition would allow them directly into their ranks.
> 
> I would also say that tieflings would be told to fuck off. if their not already.
> 
> Being Bias: racelocks stupid. Raaa. remove them all!! Free the furries.

Isn't there a tooltip about how Otava is a human supermajority, like Grenzelhoft? That being the case the racelocks being more strict than templar would be consistent in the "lore".

SFox63:
Why? What's the point of this? Inquisitors aren't nobles of the court, or knights.

AdvilPMSevere:
> I never play Inquisition. I can see racelocks for Inquisitor themselves, but perhaps loosened for the followers - so long as they abide a form of racial hierarchy. A lupian absolver knows its place, or.. whatever?
> 
> Just throwing ideas.

Could be even that the Inquisitor is making use of more "local" people in order to help with the mission in Scarlet Reach.

AdvilPMSevere:
> Why? What's the point of this? Inquisitors aren't nobles of the court, or knights. I can see tainted races like Tiefling, being hell spawn and all, but limiting other races? Wouldn't your church want to reach out to as many as possible? What's the point in limiting that?

If Inquisitors are sent from Otava, it makes the most sense for them to be human. With all round start roles, we necessarily have to assume that the world existed before the round started that week, which means whatever racelocks are in place need to be consistent with whatever lore is formulated so far.

Lioneliuss:
someone made this but different so ill close this one

PR #835, opened 2025-09-05
Adjusts recall spell flavor text

tmyqlfpir:
## About The Pull Request

This PR adjusts the recall spell flavor text when attempting to cast the spell.

`H.visible_message(span_warning("[H] closes [H.p_their()] eyes and begins to focus intently..."))`

is changed to

`H.visible_message(span_warning("[H] lowers [H.p_their()] head and begins to focus intently..."))`

## Testing Evidence

works

## Why It's Good For The Game

The character does not close their eyes, so it shouldn't reflect in the flavor text.

### DIFF
--- a/code/modules/spells/spell_types/wizard/utility/recall.dm
+++ b/code/modules/spells/spell_types/wizard/utility/recall.dm
@@ -16,7 +16,6 @@
 
 	var/turf/marked_location = null
 	var/recall_delay = 15 SECONDS
-	var/casted_with_eyes_open = FALSE
 
 /obj/effect/proc_holder/spell/self/recall/cast(mob/user = usr)
 	if(!istype(user, /mob/living/carbon/human))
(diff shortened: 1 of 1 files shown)


ShadowDeath6:
> What would you suggest to buff it, in that case?


make it not say “closes their eyes” and leave the spell as-is, I assume

Brabodie:
> I always found it strange that the text mentions the character closing their eyes, and the mechanics didn't mirror this. Given the wind up to cast this spell, having vision muted doesn't change much as you pretty much have to be isolated to cast this spell, as you're a sitting duck.
> 
> What would you suggest to buff it, in that case?

Nah honestly at this point it's so chopped I don't think you CAN buff it without making someone mad. Maybe make it cast quicker since it's so obvious you're doing it now? But I can see that getting insta denied.
My biggest complaint wasn't your PR, just needed somewhere to vent the fact that mages are so ass and I'm just seeing spells getting even more chopped while summoning continues being untouched, left as it was in its poorly ported state. I want to enjoy mages but it's impossible to stand a chance when all odds are against them and everyone seems to want them beat into obsoletion.

PR #836, opened 2025-09-05
Fixes thief attributes

SylvinSeik:
## About The Pull Request

When the "Eastern Agent" thief equipment loadout was added, it also added the statblock a second time. This PR removes it.

## Testing Evidence

I didn't test it yet, I'll do that later, but it's a pretty safe bet.

## Why It's Good For The Game

As funny as it is for a thief to have 19 speed roundstart, that's probably something we should patch out.


RyszardParys:
Ah i was suprised to see how high some of my stats were. This explains it 🎉

Issue #584, opened 2025-08-17
[BUG] Necras vow ritual does not allow original limb reattachment


benj8560:
## Describe the bug

The necras vow ritual is suppose to disallow non-original limbs to be attached as replacements. However. Even original limbs can not be reattached.

## Steps to Reproduce:

Use the necras vow ritual, cut off your leg, try and reattach your leg, get ''This one has pledged a vow to Necra. The sinews reject the false flesh. It requires their own flesh and blood.'', cry about being a cripple the rest of the round.

## Additional context
Strangely in testing this privately I found that limbs replaced via admin healing could be reattached after amputated again without issue.

ShadowDeath6:
I'm working on something closely related to this, please make a report in the discord bug reports channel with a bit more details if you can recreate it once more

PR #838, opened 2025-09-05
no longer can cut off lamia tails

PoltavskaPraca:


benj8560:
the bug with necras vow can happen regardless see #584 so moot point. But it sounds like theres plenty of other issues with it so +1

ShadowDeath6:
#841
alternative to this PR

PR #841, opened 2025-09-05
misc bodypart fixes / necra's vow

ShadowDeath6:
## About The Pull Request

the issue was bodyparts not properly updating when delimbed so there was never an owner set
necra's vow now works as intended
lamias now get their tails back when full_healed in any way, also respects their tail color
you now gain / lose lamia tail or legs respectively when your species is changed via VV, non-lamia will just get white tails like you get when selecting the species in game preferences

## Testing Evidence

image

reattaching limb properly with necra's vow active

## Why It's Good For The Game

bugfixes #584 amongst other issues that have plagued lamia for a bit


ShadowDeath6:
also fixes an oversight I had with the golem rework PR, golems can no longer do the vow to Necra ritual since they'd get no benefit from it

PR #849, opened 2025-09-06
Clean up prepare_step() species check

tmyqlfpir:
## About The Pull Request

This PR optimizes a commonly executed proc `prepare_step()`. By using AND operators instead of MOD, it will perform faster as MOD is one of the most computationally expensive instructions to execute.

The logic is functionally the same as the old code, but should run faster.

## Testing Evidence

Untitled

## Why It's Good For The Game

Rewrites `prepare_step()` to use cheaper AND operators.

### DIFF
--- a/code/datums/components/footstep.dm
+++ b/code/datums/components/footstep.dm
@@ -59,7 +59,13 @@
 	if(steps >= 6)
 		steps = 0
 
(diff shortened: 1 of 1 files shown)


PoltavskaPraca:
post-merge code review :face_holding_back_tears:

PR #855, opened 2025-09-07
makes shortsword and *small* messer swift

PoltavskaPraca:
## About The Pull Request

+ makes iron and steel shortswords swift. Same for the messer (not the kriegmesser) and gladius.

## Testing Evidence
it doth compile
image

## Why It's Good For The Game

STRchuds never ever use it. The weapon is basically a worse rapier/sabre. Some have even said it deals less dmg than a steel dagger. Having more options for when you lose your rapier/sabre is makes SPD build less miserable.


ShadowDeath6:
why is it good for the game

PR #865, opened 2025-09-07
Rework/Buff Dual Wielder Virtue

tmyqlfpir:
## About The Pull Request

This PR addresses the currently confusing description and behavior of the Dual Wielder.

Currently, the virtue behaves as follows:

* Holding two weapons of the same type and material activates the virtue
* Get ambidextrous ability with single handed weapons
* Reroll attacks if attack was parry/defended/dodge
* Reroll defend/dodge/parry if enemy attack was successful
* If both opponents have the dualist virtue, it'll cancel out and the attack will continue like normal

This PR changes the following
* If successfully hit the enemy (and not their armor), roll a 33% RNG check and do a second attack
* Only hits with the intent blunt/slash/chopping/stabbing/stab/pierce are allowed a second attack
* The secondary attack uses 50% strength
* The secondary attack uses the weapon in the opposite hand
* Attacker and target must equal or less than 1 tile distance apart
* It will not apply when attacking yourself

For werewolves, they do not inherit the traits from their normal form, so Dual Wielding trait not apply to them.

## Testing Evidence

Untitled

## Why It's Good For The Game

This improves the overall usefulness of the virtue.

### DIFF
--- a/code/_onclick/item_attack.dm
+++ b/code/_onclick/item_attack.dm
@@ -219,7 +219,7 @@
 			to_chat(user, span_notice("Success!"))
 		return
 	if(do_double_hit && HAS_TRAIT(user, TRAIT_DUALWIELDER)) // do a second follow up attack if we successfully hit our target
-		if(prob(66)) // 33% chance of doing a second hit
+		if(!prob(33)) // 33% chance of doing a second hit
 			return
 		var/obj/item/offh = user.get_inactive_held_item()
 		if(!offh)


NotGoldenglow:
I don't quite think this makes the virtue any more useful. It's a sidegrade. 
To use it, you already need to one-hand weapons. Onehanded weapons are notably weak, and they're used by people who will likely have more speed than strength... Which you're also diminishing with the /3 weapon strength, AND you're making it hit... Once, randomly? 

I might as well not dual wield. That random hit does absolutely nothing for me, because once I already have their armor punched through, I don't want to hit anywhere els.

PR #866, opened 2025-09-07
gives wretches an alternative exit back in the waterfall so they aren't permacamped in the south forest

zeroisthebiggay:
## About The Pull Request

adds a secondary exit to the wretch camp on the north east end of their camp that leads to behind the hamlet waterfall, back to where it used to be before it was inexplicably moved to the south forest i think from maybe an upstream change
i don't need to tell you that right now the south forest has literally fuckall player traffic other than new adventurers spawning in and guards and wardens stacking up to repel the constant wretch gate storms because there's fuckall else to do int he south forest
yes wretches have the underdark exit but unless you know the routes like the back of your hand it's very unintuitive to escape the underdark
as it is if you want to play wretch and interact with the half of the map with actual player density then you have to either take several detours through the bog, sprint through the town with a mask on, or go around the coast for a like five minute adventure through the water

## Testing Evidence

i had a screenshot of the travel tiles behind  the waterfall on my clipboard but i accidentally erased it pushing to origin
i tested it and it works

## Why It's Good For The Game

okay so this is going to be controversial whether or not it's actually good to enable both of the minor antagonist roles to be able to inhabit one half of the map, but bandits as is have a secondary exit quite literally like five screens away from the current main wretch exit
realistically this'll enable more frontal gate assaults as bandits and wretches invariably team up and likely still making the south forest a bumfuck nothing ever happens place because every wretch in their sane mind is going to go up north where there's Actually People To Interact With Because Nobody Besides Guards Who Wanna Frag And Fresh Spawning Adventurers are in the south forest
it's a bandaid solution to the fact that the map is super duper northsided in terms of content to do and it's naturally where players gather but it can't be fixed without the town literally being moved in its entirety
i'm sorry but the south forest sucks

### DIFF
--- a/_maps/map_files/dun_world/dun_world.dmm
+++ b/_maps/map_files/dun_world/dun_world.dmm
@@ -21163,6 +21163,13 @@
 /obj/structure/fluff/walldeco/wantedposter,
 /turf/open/floor/rogue/ruinedwood/spiral,
 /area/rogue/indoors/town/tavern)
+"gwm" = (
+/obj/structure/fluff/traveltile/wretch{
+	aportalgoesto = "wretchin1";
+	aportalid = "wretchout1"
+	},
+/turf/closed/mineral/random/rogue,
+/area/rogue/outdoors/mountains)
 "gwq" = (
 /obj/structure/roguerock,
 /turf/open/water/cleanshallow,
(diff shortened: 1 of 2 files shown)


KiteAE:
If they need another exit its not gonna be in the north east i specifically moved them out of there due to future changes i am planning to the map. 

I do not want things added to the north east of the map, the vampire lord is being moved out of there soon aswell.

PR #868, opened 2025-09-08
Revert "berserker small lil +1"

BadAtThisGame302:
Reverts Scarlet-Reach/Scarlet-Reach#645

It's one of the best if not the best Wretch class in the game, it doesn't need anymore buffs, especially in stats that matter, like endurance.


kiwedespars:
mostly done because it's supposed to have onpar stats with shaman which got a random +1 buff and i thought it was incredibly strange that shaman ended up being statwise better than wretch until the omnistat

(then it got moved to wretch for a little bit that was kinda funny)

PR #867, opened 2025-09-07
Knuckle Deep

Bobinater:
## About The Pull Request

In my previous PR, https://github.com/Scarlet-Reach/Scarlet-Reach/pull/665 I had touched the more or less attack rate of the katar and knuckles, without realizing that I left the knuckles in a far better place than I should have. As of current, they attack faster than any other blunt weapon in the game and are 2 slots, so you can carry around a LOT of them easily, while I don't know where the size code is, I do know where clickCD and integrity are so I thought I'd put this together after consulting @kiwedespars our resident unarmed abuser.

## Testing Evidence

2 line change, trust bro trust.

## Why It's Good For The Game

Knuckles are still used constantly as just a fucking pocket mace and with how many classes get 5 in unarmed, have become the brunt of armor smashing in antag fights, so. Believe me when I say with the duration of a single feint you can absolutely destroy someone through a steel helm with these things.

PR #875, opened 2025-09-08
knuckles nerf

kiwedespars:
## About The Pull Request

alternative to #867

- knuckles now take up **four** slots of inventory space. no more bagstacking. put those hips to work.
- steel knuckles wdef 8 -> 7
- bronze knuckles wdef 10 -> 8
- smash intent 1.8 integ damage -> 1.5

## Testing Evidence

image

## Why It's Good For The Game

saner alternative that doesn't gut unarmed parry builds. turns out the solution to infinite integrity is to just make them bigger. they still fit in a bag/belt so they should take a hit as a result

wdef nerfs because they're a bit high anyways

smash nerfs because onehanded and maces require twohanded to smash.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/special.dm
+++ b/code/game/objects/items/rogueweapons/melee/special.dm
@@ -69,7 +69,7 @@
 	damfactor = 1.1
 	clickcd = CLICK_CD_MELEE
 	swingdelay = 8
-	intent_intdamage_factor = 1.8
+	intent_intdamage_factor = 1.5
 	icon_state = "insmash"
 	item_d_type = "blunt"
 /// INTENT DATUMS	^
@@ -278,7 +278,7 @@
 	associated_skill = /datum/skill/combat/unarmed
 	pickup_sound = 'sound/foley/equip/swordsmall2.ogg'
 	throwforce = 12
-	wdefense = 4
+	wdefense = 0
 	wbalance = WBALANCE_SWIFT
 	thrown_bclass = BCLASS_CUT
 	anvilrepair = /datum/skill/craft/weaponsmithing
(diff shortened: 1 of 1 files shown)


kiwedespars:
> Why should steel knuckles have ~4-5 more wdef than seemingly every other blunt weapon... when they're literally knuckles? what are you parrying with? why not drop it down to the grand mace's level of... 3?

azure balance on unarmed weapons. you're expected to parry with them and the unarmed classes do not get shieldskill to make up for it.

speaking of azure i looked at their wdef when they did nerf it down to these levels and it results in sameskill parrying having 40% parry, this kills parry unarmed while leaving dodge untouched. dodge is also bloated as fuck rn i don't think it should be incentivized further

> 250 is the amount of durability a grand mace has. Why are dinky knuckles as durable as a grand mace, while also being faster than it and thus being better in every single sense?

logically this doesn't make sense.

however in terms of game balance and a desire to keep the game fun, i'm going to say that, yeah. the weapon that's expected for a class to be able to parry should at least have some integrity. 300 is too much though for a 2 tile item, and making them 4 tiles instead would solve the issue of having 900integ worth of weapon on your belt

> why keep it so much higher than steel knuckles? it's just a flat upgrade still in almost every sense

i agree with this point. it's just that bronze knuckles have always been (randomly) better on defense just to make them viable as an option so i kinda kept that going.

making them a flat downgrade is a good idea

> the mace intents have... no integ damage bonus, the mace also has... just a 1.5 damage bonus on its smash intent. Obviously the mace has 32 force... when wielded, lol, so you're losing out on an offhand which could be a shield or a staff to walk through mud. One handed, a mace has 25 force. Knuckles have 22 force. Knuckles swing significantly faster than maces do? AND maces have to be two handed to have smash intent?

they do 1.35x integ damage baseline. they can be wielded to do significantly more damage whereas knuckles cannot, obviously
knuckles swing slightly faster in exchange for them doing less damage than a cudgel per hit strike for strike.
if you're landing a feint you're not going to need a shield to crack someone's skull, so stowing your shield to just deal 90 damage to armor is the cost of doing significantly more damage yeah

this is more of an issue of smash being too good for a one-handed intent that's fair i'll probably nerf smash more and the initial integ damage on knuckles

> Why does this pr that is supposed to be nerfing knuckles have a katar buff?

katars have less integrity than daggers and break in 1 riposte. it can be dropped down a bit in damage though sure

as for the wdef reduction i was looking at the azure pr and thought "wait that's kind of a neat qol change" but also valid point i'll just save that for another pr

PR #888, opened 2025-09-09
Sword tweaks.

RyszardParys:
## About The Pull Request

Changes integrity values of some swords.

## Testing Evidence

🥀 

## Why It's Good For The Game

It pmo that some weapons had unreasonably high durability, while others had strangely low durability.
Rapiers are the easiest accessible weapon in the game, anyone can just pick "duelist's apprentice" virtue and have probably one of the best swords in the game. (one-handed, swift, strangely high durability even though the blade is supposed to be thin).
Also, so far the blacksteel flamberge just felt like "sword but expensive and fancy" even though it took two blacksteel bars and a rontz to make... Now it's more durable.


RyszardParys:
> My only feedback here is that maybe the Vaquero merc class' rapier (Cup-hilt rapier) should retain a bit more durability than a standard one. Maybe 200? Seeing as its a rather 'unique' variant

Fair, done.

PR #891, opened 2025-09-09
Deadite tags!

SylvinSeik:
## About The Pull Request

Deadites are now marked with a tag similar to vampires, or outlaws, or what have you. Shamelessly brought back from #201, but without the jump removal from deadites.
Also removes a duplicated item in the loadout because someone clearly didn't look through and double check before PRing stuff. What an idiot, right?

## Testing Evidence
Using that rat that we hate:
image

Yes, I also tested to ensure the tag and trait are removed just as the others are, but thats less fun to put together into an image
## Why It's Good For The Game

It's really hard to tell whos a deadite and who is just a creacher with some of these races, man. Doubly so when they are facing north.


ShadowDeath6:
It could be worth revamping vamp code to handle it this way as well, doubly so since I think stuff like astrata sacred flame might not count them as deadites?

PR #897, opened 2025-09-10
Slick Trick fixes so you can jump over it

E231kei500bandai:
## About The Pull Request

now you can jump over the slick trick thing

also fixes duration mistakes by adding a define

## Testing Evidence

image

## Why It's Good For The Game

makes sense that you cant slip midair


E231kei500bandai:
the slip lasting longer than it should is mega annoying but its good now

PR #930, opened 2025-09-12
makes an attempt at making invis useful against simplemobs

ShadowDeath6:
## About The Pull Request

makes mobs not detect you if you're under the invis spell's effect

this can sometimes be bypassed by stealth detection stuff (which mire lurkers have) or just attacking / being attacked

if a mob already has you as its target when you cast invis they won't attack you but they will try and shoot ranged attacks at you

## Testing Evidence

[2025-09-12 04-38-28.webm](https://github.com/user-attachments/assets/f41e0056-e5eb-4469-95a8-897926425910)

## Why It's Good For The Game

makes it a higher value pick if you want to avoid getting touched by spiders, keep in mind this can't be used to cheese simplemobs by going invis and whacking them since you drop out of invis if you click anything

this feels kind of shitty and janky but I don't know why it does so I'd appreciate it being TMed so I can fix any unforeseen issues that appear and make it feel less shitty and janky

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/hostile.dm
+++ b/code/modules/mob/living/simple_animal/hostile/hostile.dm
@@ -218,7 +218,7 @@
 /mob/living/simple_animal/hostile/proc/Found(atom/A)//This is here as a potential override to pick a specific target if available
 	if (isliving(A))
 		var/mob/living/living_target = A
-		if(living_target.alpha == 0 && living_target.rogue_sneaking) // is our target hidden? if they are, attempt to detect them once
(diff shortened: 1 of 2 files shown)


AdvilPMSevere:
> Do you use it before combat or after? I think the way it checks currently means that they'll still go after you if they detect you prior to casting invis

I'm pretty sure I've used both. I'll be doing it in-game soon so I can check

AdvilPMSevere:
> I'm pretty sure I've used both. I'll be doing it in-game soon so I can check

Yeah so upon playing a round in the bog, it seems like Mire Lurkers in particular can track you if you are invis, but only once they have seen you first. The other spiders and skeletons both will not find you if you are invis, and will deaggro if you go invis and they're already chasing you.

PR #939, opened 2025-09-12
Replaces some loadout armours with weaker versions

roarbark:
## About The Pull Request

the spellblade robes have STAB crit resist, something only available to the highest tier of light armour, they also have ARMOR_SPELLSINGER which is formidable protection. most importantly, they cover BOTH THE LEGS AND ARMS something that makes them an actual contender against the studded leather and hardened leather coat, which force you to choose between arm or leg coverage, in turn for having 100 integrity more

you probably shouldn't be able to spawn with this for free

i added "tattered" versions with less strong values, the tattered robes are as strong as a gambeson and the tattered hat is now somewhat worse than a leather helmet

## Testing Evidence
loadout items work
image

## Why It's Good For The Game

the loadout system is meant for you to have your drip ready, not spawn with almost or arguably actually best in slot armour, personally, i believe the player should at least have to interact with a tailor or the merchant for the proper versions of these items, though i made sure that you may still use them if you simply like the look of them


Plaguewalker:
> can you add the dobo robe into this please? I want one for my kazengun character but they're locked to combat classes for obvious reasons

On it already

PR #948, opened 2025-09-13
eora pomegranate tree buffs

ShadowDeath6:
## About The Pull Request

ochre aril (one that makes you explode and die) now drops 2 lux arils as well
also massively increases drop rate of the special arils for the max tier pomegranate
also decreases maximum pruning speed from 25(!!!!!!!!!!!!!!!!!!!!!!!!!) seconds at 0 farming skill to 12 seconds, decreases skill scaling so it's about the same duration when you hit max farming level (previously 4 seconds, now 4.2 seconds)

## Testing Evidence

tested and works as expected
drop rate for lux aril is now ~60% chance per golden pomegranate, up from 6%

## Why It's Good For The Game

image

imagine this but ~18 seconds x4 each time you harvest a pomegranate and also you need to give it food and water (which also each take time to give to the tree) because the tree's prune/food/water levels reset every time you harvest

considering how much time and effort it costs to bring the tree online and that you need to re-feed and re-water and re-prune from 0 every time you harvest a pomegranate (it resets all stats when you harvest the fruit) I think the rewards should be both more consistent and more useful

ochre aril is almost completely worthless because it full delimbs and decapitates you (ONLY Eorans can eat this, so you can't just give it to NPCs), this at least makes it viable in more situations because it gives you two of the revival potion ingredient (I have genuinely never once EVER seen ANYONE make a revival potion because it costs like 3 different things that are rarely ever brought together) as well as 2 lux arils (the math-inclined amongst you will realize that if you use the lux to revive the eoran after stitching them back together that this is actually a positive lux exchange)

I would argue that even after these buffs it's still on-par or weaker than either having an Astratan acolyte just use Anastasis (free lux every 2 minutes with 0 downside or opportunity cost past the devotion it uses), or having some random goblin tied down and lux farmed (1 free lux every 15 minutes with 0 cost at all past procuring the goblin)

### DIFF
--- a/code/modules/spells/roguetown/acolyte/eora.dm
+++ b/code/modules/spells/roguetown/acolyte/eora.dm
@@ -469,7 +469,7 @@
 			to_chat(user, span_warning("The tree has been fully pruned already!"))
 			return TRUE
 		var/skill = get_farming_skill(user)
-		var/prune_time = 25 SECONDS - (skill * 3.5 SECONDS)
+		var/prune_time = 12 SECONDS - (skill SECONDS)
 
 		to_chat(user, span_notice("You begin pruning the tree..."))
 
@@ -480,7 +480,7 @@
 			if(iscarbon(user))
 				var/mob/living/carbon/C = user
 				add_sleep_experience(user, /datum/skill/labor/farming, C.STAINT * 0.5)
-			
+
 			to_chat(user, span_notice("You prune some branches."))
 			update_icon()
 			return TRUE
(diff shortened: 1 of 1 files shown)


Ketrai:
Ochre arils were actually reworked on AP to revive the two nearest corpses at the cost of your own life so it's a bit more generally useful.

ShadowDeath6:
>Ochre arils were actually reworked on AP to revive the two nearest corpses at the cost of your own life so it's a bit more generally useful.
>Also design intent wise, if there's any aril that should remain rare it's definitely the ashen one. Meant as more of a RP item / forbidden fruit. Making it nigh guaranteed just opens up the church for some massive powergaming. Eating an ashen aril when you're already in the effect cures the decay for 30 minutes. That side of desperation is taken away when they're common.

I'll make some changes for these then, thanks for letting me know

>Also what's the point of having an aril used to make revival potions when you're granting the comically easy revival surgery ingredients to them en masse. No one is going to make the potions like this.

nobody does anyway because it's easier to just get lux from a goblin than go hunting for trolls and spiders, the arils were probably the most convenient part of the equation for an acolyte to get

>Eora isn't revival focuses? They still have their own revival spells too if that's been ported here.

They haven't, and I don't really want to give other patrons revival spells
Eora tree wasn't **anything** focused unless you really enjoyed fishing because despite being a major time and effort sink the drop rates were so abysmally low for anything useful that it almost wasn't worth spending your time on the spell
I don't want Eorans to outshine Astratans in being boring revival bots (they still don't even with this change) but with the base numbers it's a 1/20 chance per pomegranate to even get the lux aril which is the ONLY thing that's even remotely worth spending that much time on procuring from the tree

ShadowDeath6:
image

ochre arils now rez two people near you instead of making new arils
credit to https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/3277 for the concept, I had to take some artistic liberties with how the procs are handled so I'd like to keep this TMed just in case it doesn't work properly due to something I did

PR #951, opened 2025-09-13
Add Dominate Mind, a spell which allows you to enthrall targets and force them to obey commands (from ss13 voice of god)

E231kei500bandai:
## About The Pull Request

Add Dominate Mind, a spell which allows you to enthrall targets and force them to obey commands (from ss13 voice of god)

## Testing Evidence

https://github.com/user-attachments/assets/12b294ab-07c0-4793-906b-1b0f6cc0f73c

## Why It's Good For The Game

images

uhh idk yet we gotta test this shit first and get feedback and stuff


E231kei500bandai:
note I was unable to test the hearing functionality because idk how to have two clients at once on scarlet reach

for some reason my dream daemon go button does nothing and i can only use the vscode method

UntoldTactics:
https://github.com/user-attachments/assets/15eba048-f2ba-4165-9404-4febed296439

UntoldTactics:
literally everyone is getting a SUBMIT stat panel and a submit to master button  even if the spell ain't cast on you
kinda wonky

code/modules/spells/spell_types/wizard/invoked_single_target/dominate_mind.dm
```diff
@@ -0,0 +1,244 @@
...
+
+	// Don't reformat the puppet's own speech
+	if(puppet == speaker)
+		return
+
+	// Reformat the raw message text.
+	hearing_args[HEARING_RAW_MESSAGE] = span_phobia(hearing_args[HEARING_RAW_MESSAGE])
+
+/// This verb allows a player under the effect of Dominate Mind to permanently submit to their master.
+/mob/living/proc/submit_to_master()
```
chaoko99, line 229: this verb appears to be available to non-dominated players 

Image

AdvilPMSevere:
I have some thoughts after playing around with this in-game.

First, I wanted to note that it seemed like something was going on with the cooldown being extremely long-testing it I found that it would take over 90 seconds frequently to reset, and even accounting for time dilation would take far longer than the intended 30 second cooldown, in the order of 50-75 seconds of in-game time.

Second, I think it might be TOO difficult to succeed with this spell. For context, I was spent a decent amount of time using this against a skeleton npc, which all have 1 INT, as a character with legendary arcane skill. I found it was very hard to reliably upgrade to tier 3, even after you pass the quite difficult check to get a critical success to enable you to do so in the first place. I imagine that trying to do this against any actual player is nearly impossible. However, it definitely should be difficult to mind-control someone-which is why I've included the third part of this comment.

Third, I have a comment/suggestion. I think it might be better if this was split up somehow into two spells or features; one being the enthrall, long-term mind control side of things, and the other being exclusively the voice of god command system. The reason for this is that the voice of god commands, which although quite fun and flavorful are not very powerful, are locked behind the same restrictions as the very serious aspect of general mind-controlling, which is (rightly) quite hard to do. I think a solution to this might be to simply remove the total mind-control aspect of this spell, while keeping the voice of god commands, and doing away with the tier system for them. The general enthralling aspect might work better as something that needs more investment on the part of the wizard, perhaps as a ritual or craftable thing of some kind.

In any case I think this is super neat and look forward to it being improved.

PR #953, opened 2025-09-13
Kazegun Loadout Update

Plaguewalker:
## About The Pull Request
Adds (hopefully) weak versions of the robes along with the cut-throat pants to the loadout system, the pants typo was fixed.

## Testing Evidence
lodout update 2
loadout update
Three out of the Five just to show for consistency they all have the same weakened state
formal dobo
worn dobo
florals

## Why It's Good For The Game
Further rounds out the loadout for Kazengun items, the robes themselves are severely weakened compared to their actual worthwhile counterparts.

### DIFF
--- a/code/modules/clothing/rogueclothes/armor.dm
+++ b/code/modules/clothing/rogueclothes/armor.dm
@@ -1535,6 +1535,12 @@
 	allowed_race = NON_DWARVEN_RACE_TYPES
 	flags_inv = HIDEBOOB|HIDECROTCH
 
+/obj/item/clothing/suit/roguetown/armor/basiceast/light
+	name = "ronin dobo robe"
+	desc = "A dirty tattered dobo robe with white lapels, robes primarily worn by the nomadic or disgraced ronin traveling from their homeland."
+	armor = ARMOR_PADDED
+	prevent_crits = list(BCLASS_CUT, BCLASS_BLUNT)
+
 //less integrity than a leather cuirass, incredibly weak to blunt damage - great against slash - standard leather value against stab
 //the intent for these armors is to create specific weaknesses/strengths for people to play with
 
@@ -1545,6 +1551,12 @@
 	item_state = "eastsuit2"
 	armor = list("blunt" = 50, "slash" = 90, "stab" = 60, "piercing" = 30, "fire" = 0, "acid" = 0)
 	max_integrity = 200
+
+/obj/item/clothing/suit/roguetown/armor/basiceast/crafteast/light
+	name = "formal dobo robe"
+	desc = "A dobo robe with a red tassel, used primarily for general use by the populace for formal occasions."
+	armor = ARMOR_PADDED
+	prevent_crits = list(BCLASS_CUT, BCLASS_BLUNT)
 
 //craftable variation of eastsuit, essentially requiring the presence of a tailor with relevant materials
 //still weak against blunt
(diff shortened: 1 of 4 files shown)


ophaq:
Can these be upgraded or repaired into regular ones with some effort?

code/modules/clothing/rogueclothes/armor.dm
```diff
@@ -1546,6 +1552,12 @@
 	armor = list("blunt" = 50, "slash" = 90, "stab" = 60, "piercing" = 30, "fire" = 0, "acid" = 0)
 	max_integrity = 200
 
+/obj/item/clothing/suit/roguetown/armor/basiceast/crafteast/light
+	name = "formal dobo robe"
+	desc = "A dobo robe with a red tassel, used primarily for general use by the populace for formal occasions."
+	armor = ARMOR_PADDED
```
zeratino, line 1558: This goes for all of the additions here, but it would probably be better to use something like ARMOR_PADDED_BAD, instead of ARMOR_PADDED, since this is supposed to be a loadout item. 

ARMOR_PADDED gives the same protection classes as a regular gambeson.

code/modules/clothing/rogueclothes/armor.dm
```diff
@@ -1535,6 +1535,12 @@
 	allowed_race = NON_DWARVEN_RACE_TYPES
 	flags_inv = HIDEBOOB|HIDECROTCH
 
+/obj/item/clothing/suit/roguetown/armor/basiceast/light
+	name = "ronin dobo robe"
```
zeratino, line 1539: This is a nitpick, but the Kazengite adventurer clothes are based around Koreans, not Japanese.

Maybe it would be better to title it as a "fragile dobo robe," instead.

code/modules/clothing/rogueclothes/pants.dm
```diff
@@ -589,6 +589,10 @@
 	allowed_race = NON_DWARVEN_RACE_TYPES
 	flags_inv = HIDECROTCH
 
+/obj/item/clothing/under/roguetown/heavy_leather_pants/eastpants1/light
+	name = "cut-throat's pants"
+	desc = "Foreign pants worn by those from the Kazengun region."
+
```
zeratino, line 595: Without any change in an armor value (i.e. ARMOR_PADDED_BAD), this shares all of the same characteristics with the above pants - barring a name and description change.

This can be fixed by adding in an "armor" variable and assigning a definition to it, like how you did with the previous additions.

It would look like:

armor = ARMOR_PADDED_BAD.

zeratino (commented):
Great first PR. Aside from the reviews I made, I don't think the floral robes should be included, since they're on a one-slot mercenary role.

code/modules/clothing/rogueclothes/armor.dm
Plaguewalker: I went with armor padded at the time as I was using the spellcaster robe fix for reference material.

Plaguewalker:
> Great first PR. Aside from the reviews I made, I don't think the floral robes should be included, since they're on a one-slot mercenary role.

I only added it because, they're really the only "proper" formal robes the kazegunites get and I'd rather not leave it on the table as then it feels wholly incomplete. I understand the qualms you have with it, but at the same time if we're editing it to be all ARMOR_PADDED_BAD does it really add much in this regard for it to warrant being cut?

ophaq:
I think the floral robes should stay. They're nice drip and not many people play the class. Feels underused otherwise.

ophaq:
I tested this locally, and got an undefined type path error. Make sure to check if your things got in as they didn't on my end, and it would be a shame to miss out on this!

Plaguewalker:
> I tested this locally, and got an undefined type path error from it on launch. Make sure to check if your things got in as they didn't on my end, and it would be a shame to miss out on this!

Shouldn't be undefined type path, I had a clean compile when I did the code

PR #964, opened 2025-09-14
limits bandit sawbones slots

arnociiroc:
## About The Pull Request
reduces the available amount of sawbones slots to two

## Testing Evidence

tis be a one line change mi'lord
## Why It's Good For The Game
having four feint masters with 16+(or more with potions) int and dodge expert with master swords on a single team, probably wasn't the greatest of ideas.

image


Lioneliuss:
> we only want one
> two slots

arnociiroc:
i can change it to one if a maintainer wants that

PR #968, opened 2025-09-14
disables dungeoneer, adds lost slot to MAA

chaoko99:
## About The Pull Request
Title.

## Testing Evidence

I cannot figure out how to give myself admin perms on my local server
## Why It's Good For The Game

image

less of this (cut a corpse's legs off without ever verifying it was a corpse, masked it, then pillory'd it, AFTER being told by the Castifico that it was dead.)

I don't personally believe the long-term behavior shown by people playing Dungeoneer justifies the existence of the class. It is loved by shitters, and so long as we continue not to punish shitters, it will continue to be an issue. 

The current population of people playing these classes makes antagonists fucking suck to play. They puppy guard the antagonists and make sure they get no more agency in the round with roughly the same candor as someone who would get job banned from warden on TG. 

Suiciding and playing a different character often is your literal only option with some people.

Either people who want to be wardens can volunteer at the beginning of the round as MAA, or they can suggest a better solution.

The reason people instantly try to escape from the dungeons is because you suck, guys. Be better.

### DIFF
--- a/code/__DEFINES/jobs.dm
+++ b/code/__DEFINES/jobs.dm
@@ -111,7 +111,7 @@
 
 #define GUARDSMAN	(1<<0)
 #define MANATARMS	(1<<1)
-#define DUNGEONEER	(1<<2)
+#define DUNGEONEER	(1<<2) //Unused
 #define SQUIRE		(1<<3)
 #define BOGGUARD	(1<<4)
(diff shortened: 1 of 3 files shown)


AttackHelicopter6784:
im sorry chao but this is just silly, removing the role won't do anything to fix this.

PR #969, opened 2025-09-14
zombie immune also stops rotting

ShadowDeath6:
## About The Pull Request

another attempt to stop golem zombification

## Testing Evidence

waited about 20 minutes with a golem corpse and a human corpse, no zombie or rotting on the golem while the human transformed and started making the fly buzzing noises

## Why It's Good For The Game

golem zombies are really bad because you can't rotcure them without the priest spell
TM first please in case this breaks something else

### DIFF
--- a/code/datums/components/rotting.dm
+++ b/code/datums/components/rotting.dm
@@ -64,6 +64,9 @@
 	if (istype(A, /area/rogue/indoors/town))	//Stops rotting inside town buildings; will stop your zombification such as at church or appothocary.
 		return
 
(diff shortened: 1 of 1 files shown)


ShadowDeath6:
> Yeah, I just did a quick check; this would stop anyone with ZOMBIE_IMMUNE from ever rotting. That seems really strong!

This was done intentionally so that golems and dolls don't rot, given they're not made of things that can rot, I assume it would further apply to other people immune to being deadite'd as well
The only people with deadite immune are golems/dolls, zombies (there's a check for zombies here anyway, so the check shouldn't apply to zombies that were killed) and zizo worshipers

PR #970, opened 2025-09-14
[TM PLS] makes charge timer account for ticklag to make it feel less inconsistent.

chaoko99:
…
## About The Pull Request
Multiplies charging progress by subsystem tick delta.

## Testing Evidence
 I cannot test this in a sim environment due to not being able to replicate lag under load.

## Why It's Good For The Game
spellcasting feels BADDDDDD sometimes.

### DIFF
--- a/code/_onclick/drag_drop.dm
+++ b/code/_onclick/drag_drop.dm
@@ -247,13 +247,13 @@
 	L.used_intent.prewarning()
 
 	if(!charging) //This is for spell charging
-		charging = 1 
+		charging = 1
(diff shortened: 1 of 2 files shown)


chaoko99:
so this doesn't appear to be working. I theorize the reason is because the subsystem responsible for tracking the charge time is running once every TWO seconds, meaning it's taking literally twice as long because it's not running in seconds, but in ticks.

For some fucking reason.

chaoko99:
https://streamable.com/evkn0t added a feature to it

PR #971, opened 2025-09-14
Add checks for TRAIT_NOPAIN alongside a few checks for TRAIT_NOPAINSTUN

polymathPretender:
## About The Pull Request

I added specific checks for TRAIT_NOPAIN along a few places that were only checking for NOPAINSTUN. It didn't seem to make sense for someone to be able to have benefits for not being stunned by pain, but not have those same benefits for not feeling pain at all.

## Testing Evidence

psydonite

## Why It's Good For The Game

In my opinion it makes internal logic more consistent and predictable for people playing.

### DIFF
--- a/code/modules/mob/living/carbon/life.dm
+++ b/code/modules/mob/living/carbon/life.dm
@@ -91,7 +91,7 @@
 							stuttering += 5
 							emote("painmoan")
 							return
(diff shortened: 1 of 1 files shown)


dawsonkeyes (commented):
compile is not evidence of testing

ShadowDeath6:
just for posterity's sake, I don't think this PR actually does anything meaningful because species.dm already has a nopain check above everything handling pain and handle_random_events() in living/life.dm is overwritten by carbon/life.dm (which /human inherits)
afaict it would only impact simplemobs that have the nopain trait but not the nopainstun trait

PR #977, opened 2025-09-15
Half-kin can no longer be nobles

Kraseo:
## About The Pull Request

What it says on the title; half-kin can no longer be (most) nobility roles. They're the direct descendants of wild-kin after all, which means they were illegitimate children.

## Testing Evidence

Captura de pantalla 2025-09-15 154725

## Why It's Good For The Game

Wild-kin can't be nobles. It makes no sense from a lore standpoint that their offspring would be allowed to be, say, the grand duke of Scarlet Reach, or its inheritor. Bastards don't exactly look great with the clergy!

### DIFF
--- a/code/__DEFINES/roguetown.dm
+++ b/code/__DEFINES/roguetown.dm
@@ -194,7 +194,6 @@
     /datum/species/human/northern,\
     /datum/species/elf/wood,\
     /datum/species/human/halfelf,\
-    /datum/species/demihuman,\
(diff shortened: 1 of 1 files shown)


ShiningM00N:
The least we need is more role limitations on the specific subsect of jobs that are already incredibly limited.

Brabodie:
great pr but itll never go through because people are far too comfortable with their comfort characters already

PR #938, opened 2025-09-12
Church rework

DelineFortune:
## About The Pull Request

a lot of things people asked to draft it to see the progress

Steward cannot order ton of red for 100 mammons, only lesser one

What’s actually going on? The Church is leaving the city because the second-most influential and thematically important faction should be driving the story, intrigue, and conflict, not just acting as a box where people get free treatment, beds, and food.

The Church gets research content that lets it unlock upgrades (travel the map and jam items into statues to earn research points) and shove valuables into the Heart to buy them. Organs, artifacts, chemicals, upgrades, and the ability to purchase miracles and rituals (with a discount on your patron’s rituals).

Also, the Church becomes strong enough on its clay (smite miracle is stronger+buff). Acolytes get wrestling 5 back and polearms 3. Still no combat traits or stats, so they can’t even counter the usual “mouth grab -> die.”

City clergy start without miracles but with some starting points. They also spawn without automatic Devotion and are expected to unlock it (bitch it takes about 10 minutes if you are solo - you have 1,000 Devotion as an Acolyte or Druid anyway; put some effort into the game). So don’t expect Astratans to have Anastasia at the start - they can just not have it but 10 offensive miracles

Also some miracles are upgrade-able.

image

## Testing Evidence

image

image

image

image

image

image

image

image

image

## Why It's Good For The Game

unique content woah
no more medical bay gameplay only and HEAL ME FOR FREE OR DIE

PR #980, opened 2025-09-15
Church's first part

DelineFortune:
## About The Pull Request

The Church is leaving the city. 
In its place there’s now an unlit park - a good spot to do ERP or to hide and kill or hide from million guards and legendary cooking maids with pans after you.
The Church has moved outside the city. 
Soon I’ll add the ability for them to make offerings to statues, gain research points, steal and sacrifice gold in the name of their gods to unlock new content.  (The code exists already im runing tests https://github.com/Scarlet-Reach/Scarlet-Reach/pull/938) 
Acolytes get master wrestling and joyreman polearms. 
Clergy gain bonus on their land. Will see some rounds, there is still solutions to let them handle better against 15+ antags every round we have.
The Church should be the faction, bring conflict and story to the game - not a box that replaces the apothecary and the inn. Not a place where players get healing, food, and a bed for free.
Clergy gets Temple Smite (new cleric miracle) - usable only within church areas. Smite simplemobs for 50 damage or humans with a 3s stun and -3 Speed / -3 Con / -3 Per for 15s. Won't affect clergy_trait gamers unless they are Excommunicated. Both caster and target must be inside a church.
Also removed invisible barrier from old spawn place

## Testing Evidence

image
image
image
image
image
image
image
image

## Why It's Good For The Game

We are trying to step away from being medical bay and become a fraction.

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -340,7 +340,7 @@ GLOBAL_LIST_INIT(roguetraits, list(
 	TRAIT_ADRENALINE_RUSH = span_notice("I'm invigorated in the midst of battle! I don't feel my wounds!"),
 	TRAIT_DEADITE = span_danger("The Rot has overtaken me."),
 	TRAIT_PUZZLEMASTER = span_notice("I've solved an impossible puzzle!"),
-  TRAIT_CLERGY = span_notice("I am a servant of the local House of the Ten.")
+	TRAIT_CLERGY = span_notice("I am a servant of the local House of the Ten.")
 ))
 
 // trait accessor defines


DelineFortune:
> Sorry that the first thing I said was a criticism. I was just concerned with ending up in a place where antagonist actions get stonewalled or hemmed into specific strategies.
> 
> The changes you're making look quite interesting. The separation from the town will definitely add an interesting dynamic of the church being "outsiders" in that sense.

We can't get experience without taking some radical things. Yes, I fully admit that my actions might fail. But having factions within a one screen run doesn't seem very reasonable to me. Everything is too close to the center

Church is still close to the centre, strong on it's own lands. But also was and still valid to all antags and average scarlet combat "mouth grab -> 5% chance to resist -> death"

Jamdrawers:
1) I'm dubious of spreading out the population too much; Dunworld is kind of too big as it is already, don't want it to feel too empty. It's important to be able to find people so that you can interact with them, and I think it's been to the game's benefit that the church was close to hand to swing by to.
2)  If we think that the robed holy spellcasters are too weak, I don't think the solution should be to make them also top-of-the-range kung-fu masters instead of leaning into their role as holy spellcasters. And not everyone has to be primary combatants.
3) I worry that the place is just all-round too big. Maybe it'd fit better on scarlet where it has the population density to fill it out, but I suspect it'll feel very empty most of the time.

Brabodie:
> We are going to check how its going to be. On low pop there usually no templars but +10 bandits/wretches And as clergy they are valid targets for them and as the church is outside of the city now (no scom rings, no scoms) they cannot call for help.
> 
> From now we have a problem that average acolyte gets mouth grabbed-killed (5% chance to resist)

Never have I ever even seen an acolyte get so much as touched while playing together with bandits, only thing I've seen is them venturing outside the church trying to validhunt. If anything, acolytes dying reduces medbay gameplay which the PR is trying to achieve so I see it as a plus.

code/game/area/roguetownareas.dm
```diff
@@ -40,6 +41,13 @@ GLOBAL_LIST_INIT(roguetown_areas_typecache, typecacheof(/area/rogue/indoors/town
 	if((src.warden_area == TRUE) && HAS_TRAIT(guy, TRAIT_WOODSMAN) && !guy.has_status_effect(/datum/status_effect/buff/wardenbuff)) // Warden
 		guy.apply_status_effect(/datum/status_effect/buff/wardenbuff)
 
+
+/area/rogue/Entered(mob/living/carbon/human/guy)
```
Detective-Google, line 45: this is a duplicate definition of `Entered()` (see L51 of this file); it's also horrible object oriented programming-- you could define this on `/area/rogue/indoors/town/church/Entered()` so that you are only running this check if you enter that area, rather than EVERY SINGLE AREA. (obv would have to duplicate it for any church area that doesn't share a path with that but that's still way better than this.)

code/game/objects/structures/flora.dm
```diff
@@ -415,7 +415,29 @@
...
+    icon_state = "maple1"
+
+/obj/structure/flora/maple/b
+    icon_state = "maple2"
+
+/obj/structure/flora/maple/c
+    icon_state = "maple3"
+
+/obj/structure/flora/maple/d
+    icon_state = "maple4"
```
Detective-Google, line 440: needs to be converted to tab indentation, currently uses spaces.

code/game/objects/structures/fluff.dm
```diff
@@ -1355,3 +1355,109 @@
 	stake.forceMove(drop_location())
 	stake = null
 	qdel(src)
+
+/obj/structure/fluff/statue/shrine/donation
+    parent_type = /obj/structure/fluff/statue
```
Detective-Google, line 1360: there's no reason to mess with the parent_type like this for this structure-- having to do this means something has gone horribly wrong. https://ref.harry.live/datum/var/parent_type 

why not just have it be shrine/donation/abyssorqueststatue (etc for other gods)??? then you could define the `icon` var to `icon = 'icons/roguetown/misc/64x96.dmi'` on only that parent and not duplicate it to the children, same for `max_integrity = null`.

code/game/objects/structures/fluff.dm
```diff
@@ -1355,3 +1355,109 @@
 	stake.forceMove(drop_location())
 	stake = null
 	qdel(src)
+
+/obj/structure/fluff/statue/shrine/donation
+    parent_type = /obj/structure/fluff/statue	
+
+/obj/structure/fluff/statue/shrine/abyssorqueststatue
+    parent_type = /obj/structure/fluff/statue/shrine/donation
```
Detective-Google, line 1363: more space indentation which should be converted to tabs.

code/game/turfs/closed/walls.dm
```diff
@@ -211,3 +211,33 @@
 	add_overlay(dent_decals)
 
 #undef MAX_DENT_DECALS
+
+/turf/closed/wall/churchpass
+    name = "stone wall"
+    desc = "A strange wall, faintly shimmering."
+    icon = 'icons/turf/walls/stone_wall.dmi'
+    icon_state = "stone"
+    explosion_block = 1
```
Detective-Google, line 220: = TRUE?

code/game/turfs/closed/walls.dm
```diff
@@ -211,3 +211,33 @@
 	add_overlay(dent_decals)
 
 #undef MAX_DENT_DECALS
+
+/turf/closed/wall/churchpass
+    name = "stone wall"
```
Detective-Google, line 216: more space indentation, convert to tabs.

code/modules/hotspring/falling_leaves.dm
```diff
@@ -42,3 +42,29 @@
 	particle_effect = new(src, /particles/sakura, 6)
 	//render the steam over mobs and objects on the game plane
 	particle_effect.vis_flags &= ~VIS_INHERIT_PLANE
+
+/particles/maple
```
Detective-Google, line 46: uses space indentation, should be converted to tabs, you should not artificially add multiple spaces to line up code like this, just `color = "#E23B1E"`, *not* `color       = "#E23B1E"`

code/modules/spells/roguetown/acolyte/general.dm
```diff
@@ -1,3 +1,87 @@
...
+	associated_skill = /datum/skill/magic/holy
+	antimagic_allowed = TRUE
+	recharge_time = 10 SECONDS
+	miracle = TRUE
+	devotion_cost = 10
+
+/obj/effect/proc_holder/spell/invoked/temple_smite/proc/_is_church_area(var/area/A)
+	if(!istype(A))
+		return FALSE
+	if(A:church_area)
```
Detective-Google, line 24: use `A.church_area` instead, : overrides the compile time check that the variable actually exists. 
see https://ref.harry.live/operator/colon vs https://ref.harry.live/operator/dot 

the istype checks should also be non-required if you're using a variable to check if an area is in the church, since it should be TRUE on those areas, and is therefore redundant-- 

Adding a variable to the base /area type also sucks if not necessary but isn't the end of the world.-- better solution might be a #define CHURCH_AREAS with a list of desired areas to be effected in; could then check `if(get_area(mob) in CHURCH_AREAS)`.

code/modules/spells/roguetown/acolyte/general.dm
```diff
@@ -1,3 +1,87 @@
...
+		revert_cast()
+		return FALSE
+
+	if(istype(target, /mob/living/simple_animal))
+		target.adjustBruteLoss(50)
+		target.visible_message(
+			span_danger("[target] is scorched by divine wrath!"),
+			span_userdanger("Divine wrath tears through me!")
+		)
+		playsound(target, src.sound, 60, FALSE)
```
Detective-Google, line 62: the `src` in `src.sound` is implied and can be discarded; e.g. `playsound(target, sound, 60, FALSE)

DorfGuy1:
This looks so peak. But it also wipes out a lot of the quiet parts of the map that people use to get things done.

DelineFortune:
> This looks so peak. But it also wipes out a lot of the quiet parts of the map that people use to get things done.

It didn't. I literally used a mointaint to avoid removing anything else

The only one thing that is missing is miner's house.
You can toss examples here ill look what can I do.

AFTERLIVE:
please for the sake of game balance, replace the stun with either a slow or a disarm. point and click stuns have never been fun for anyone to fight against and it's basically a 0 skill spell that can end fights in seconds and make them not interesting.

chemistrymain3:
Good PR.

DelineFortune:
> please for the sake of game balance, replace the stun with either a slow or a disarm. point and click stuns have never been fun for anyone to fight against and it's basically a 0 skill spell that can end fights in seconds and make them not interesting.

Yes its going to be reworked to be less cancer

DelineFortune:
> the inquisition hunts heretics, the keep already mogs people over taxes, and now the apothecary will overtake the church as the main healers.
> 
> what's the church supposed to do now that they are isolated? wait to get raided or go validhunt even harder? start fights with the town and take away from the ACTUAL antags?

Supposed to be because they going to get way more content soon 

Like a way to buy miracles, better organs, artefacts and etc

The code is getting ready here:

https://github.com/Scarlet-Reach/Scarlet-Reach/pull/938

zeratino:
I NEED IT BACK. PLEASE. RE-TM THIS!!

PR #986, opened 2025-09-16
adds expert athletics to lunacy embracer

dawsonkeyes:
lunacy embracer has 0 athletics skill, this fixes that


AdvilPMSevere:
I love you

PR #987, opened 2025-09-17
adds in padded hoods to the merchant shop

zeratino:
## About The Pull Request
see title

## Testing Evidence

image

## Why It's Good For The Game
this seems to be an oversight from when it was first pr'd? either way it'd be nice to see this stuff appear more often


roarbark:
all of these items, including studded leather and hardened coat should probably cost double

zeratino:
> all of these items, including studded leather and hardened coat should probably cost double

probably, but most tailors usually charge 3 zil a piece & the merchant is supposed to up charge for these
it's compounded with the silver face so i don't really see an issue with it

PR #1014, opened 2025-09-20
misc magic changes

ShadowDeath6:
## About The Pull Request

elementals can now mine rock walls
hedge mages now start with guide to magic
you can now get crafting books in loadouts

## Testing Evidence

image

works as expected

## Why It's Good For The Game

gives elemental summons more flavor, especially crawlers since you can bind them early and they can be a mining buddy

other stuff was on request, not sure if there are balance ramifications to allowing people to have a recipe book in their loadout but it does take up a loadout slot


ShadowDeath6:
simplemobs can now be pulled through travel tiles once more

PR #1015, opened 2025-09-20
Makes Lunacy Embracers Embrace the Thing They are Supposed to be Embracing, More

AdvilPMSevere:
## About The Pull Request

Removes many "civilized" skills from embracers that don't fit with their wild nature. Adds some skills that people who scrounge around in the wild naked all day should have. 
Removed:

Farming
Sewing
Tanning
Cooking
Carpentry
Alchemy
Masonry

(Fishing was left as it is to allow Abyssor embracers to fish.)

Adds or buffs:

Sneaking
Climbing
Swimming
Tracking
Butchering

Also gives the ORGAN_EATER and DARKVISION traits to embracers. 

## Testing Evidence

9/10 dentists agree that this works.

## Why It's Good For The Game

This skill overhaul removes a bunch of these weird skills that really make no sense for embracers to have, and helps them in areas where they clearly should be quite good. Organ eater will allow them to eat organs, consistent with their lifestyle of eating raw meat and drinking shit water. For a class which cannot wear clothes, they cannot carry around any source of light without using up a hand slot-and the idea of a lunacy embracer needing to carry around a lantern or torch all the time isn't really consistent with their intended place as creatures of the wild. Night-eyed is almost a mandatory virtue pick for them for this reason, so giving them its benefits from the start will allow embracer characters to play with other virtues and make more varied characters without forcing them to forego seeing.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/wretch/lunacyembracer.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/wretch/lunacyembracer.dm
@@ -16,6 +16,8 @@
 		TRAIT_WOODWALKER,
 		TRAIT_NASTY_EATER,
 		TRAIT_CALTROPIMMUNE
+		TRAIT_ORGAN_EATER
+		TRAIT_DARKVISION
 	)
 	subclass_stats = list(
 		STATKEY_STR = 3,
@@ -31,17 +33,12 @@
 		/datum/skill/combat/wrestling = SKILL_LEVEL_MASTER,
 		/datum/skill/combat/unarmed = SKILL_LEVEL_MASTER,
 		/datum/skill/misc/athletics = SKILL_LEVEL_EXPERT,
-		/datum/skill/misc/sewing = SKILL_LEVEL_APPRENTICE,
-		/datum/skill/craft/tanning = SKILL_LEVEL_APPRENTICE,
-		/datum/skill/labor/farming = SKILL_LEVEL_MASTER,
-		/datum/skill/craft/carpentry = SKILL_LEVEL_APPRENTICE,
-		/datum/skill/misc/climbing = SKILL_LEVEL_JOURNEYMAN,
-		/datum/skill/misc/swimming = SKILL_LEVEL_APPRENTICE,
-		/datum/skill/misc/tracking = SKILL_LEVEL_JOURNEYMAN,
+		/datum/skill/misc/climbing = SKILL_LEVEL_EXPERT,
+		/datum/skill/misc/swimming = SKILL_LEVEL_EXPERT,
+		/datum/skill/misc/tracking = SKILL_LEVEL_EXPERT,
+		/datum/skill/misc/sneaking = SKILL_LEVEL_APPRENTICE
 		/datum/skill/craft/crafting = SKILL_LEVEL_APPRENTICE,
-		/datum/skill/craft/masonry = SKILL_LEVEL_JOURNEYMAN,
-		/datum/skill/craft/alchemy = SKILL_LEVEL_JOURNEYMAN,
-		/datum/skill/craft/cooking = SKILL_LEVEL_JOURNEYMAN,
+		/datum/skill/labor/butchering = SKILL_LEVEL_NOVICE
 		/datum/skill/labor/fishing = SKILL_LEVEL_JOURNEYMAN,
 		/datum/skill/magic/holy = SKILL_LEVEL_JOURNEYMAN,
 	)


AdvilPMSevere:
> I wouldn't call them civilized skills. You can be a wild person who farms, and works leather. Masonry... Yeah, that can go, and so can alchemy, and cooking. They really shouldn't care whether or not their meal is fancy.

You can be wild, maybe, but not so wild you are a lunatic. With tanning and sewing, anyway, considering they can't wear clothes, I figure there is no point in having the skills anyway.

AdvilPMSevere:
> The only reason I say to give it to them, is because every Lunatic player I know usually spends some time working on a satchel and drip items with sewing/skincrafting. Drip is an important component of RP, and there's no reason we should make an already limited (in ability) role even more limited. Let them have their drip, at least.

As far as drip goes, the starting loadout options probably cover most of it. As for the satchels/belts, I do think its a bit strange for embracers to be carrying things at all for the most part. That being said, I don't really have a problem leaving sewing, but I'd rather wait and hear from more people/have the PR reviewed. If it gets merged right away, I might just make another one re-adding sewing or something.

benj8560:
> (Fishing was left as it is to allow Abyssor embracers to fish.)

> Removed: farming

?? isn't this role played by like 80% dendorites? Would the same argument not apply to farming? You know ideally..some day we add failth based extra stats to all the non-church clerical classes.

AdvilPMSevere:
> ?? isn't this role played by like 80% dendorites? Would the same argument not apply to farming? You know ideally..some day we add failth based extra skills to all the non-church clerical classes.

Farming strikes me as a much more "civilized" activity than fishing. It isn't perfect, but I wanted to still let Abyssor embracers have fish other than the one they can summon via ritual, not that embracers get ritualist anyway, which they maybe should. Ideally they are "fishing with their hands" or whatever. 

Farming is basically the opposite of "running free in the wild", and why I think it doesn't make sense. Farming requires clearing land, planting and harvesting, etc. All things that are extremely not natural, and wouldn't be something done by animalistic lunatics for sure.

benj8560:
> Farming is basically the opposite of "running free in the wild", and why I think it doesn't make sense. Farming requires clearing land, planting and harvesting, etc. All things that are extremely not natural, and wouldn't be something done by animalistic lunatics for sure.

you can not fish with your hands (as far as I'm aware I don't do much fishing gameplay) needing a rod/fish traps both things I would call a noteworthy bit of tech in are IC age.

you can however manage an entire farm with nothing but your bare hands, a campfire and the most advanced thing a bucket (crazy) at farming level 3. I do not see your point for fishing making any more sense then farming. I don't think they need anything more then 3 though.



> not that embracers get ritualist anyway,

Huh they don't get ritualist? They likely should yeah....maybe with craftable ritual chalk to keep the naked spawning

benj8560:
I also wanna make a light argument for cooking. They should know how to cook a slab of meat over a campfire (I think level 2 cooking is enough?). We don't need to FORCE the raw food eating onto people. Let people have some flexibility in just how savage their embracer is.

AdvilPMSevere:
> you can not fish with your hands (as far as I'm aware I don't do much fishing gameplay) needing a rod/fish traps both things I would call a noteworthy bit of tech in are IC age.

Unfortunately this is the case, but it would be strange if Abyssor embracers just couldn't fish at all, and so we put up with the cages and fishing rod for the sake of having fish.

> you can however manage an entire farm with nothing but your bare hands, a campfire and the most advanced thing a bucket (crazy) at farming level 3. I do not see your point for fishing making any more sense then farming. I don't think they need anything more then 3 though.

Fishing is like hunting-something primal, where you go out and get the fish, kill it and eat it. Farming is not that at all-it isn't natural, being entirely an invention of and indivisible part of civilization, which is the exact opposite of what lunacy embracers are supposed to be. If there is a lunacy embracer working a plot of land, that means they-cleared land (not very wild), are maintaining it from animals, pests and weeds (not very wild), and presumably have somewhere to store their harvest and seeds (VERY not wild). Fishing, even with a rod or cage, is WAY more embracing of wild craziness and animalistic instinct than farming ever could be. Dendorites who don't go this far in embracing the wild exist, they're druids. Those are the ones doing the farming.

AdvilPMSevere:
> while i agree with the sentiment
> 
> not being able to craft a new leg when you inevitably get legcut by some hubbie for your 100mam bounty is aids.

Great point actually. I assume it's some level of carpentry, but I don't know specifically. I'll add it.

benj8560:
> Great point actually. I assume it's some level of carpentry, but I don't know specifically. I'll add it.

It's crafting skill surprisingly. I'm pretty sure you need level 2 for it.

NotGoldenglow:
> [...] Also, as far as cooking is concerned, that is another very not wild thing to do [...]

Humans have been cooking food since the discovery of fire, hundreds of thousands if not a million or so years ago. While garnishing, spicing and etc. is definitely 'civilized', Grug threw his meat on the fire because it made the meat taste better and Grug was happy. That is quite literally one of the first things the absolute earliest humans we can even call *human* in the first place, started to do.

... Not to mention Lunacy embracers aren't million year old unevolved apes. Either, at this point, we take away every single skill from them because every argument that can be made boils down to 'well they're wild, they shouldn't care about this' or we accept that they're still people who chose a different lifestyle and the players can choose the degree of 'true' wilderness they wish to adopt.

Dagonlives:
I really enjoy this role.  Part of the enjoyment is the variety of crafting option.s They aren't particularly broken, or problematic. Let me play my savant dendor moron who can farm at a master level as a lunatic. It's not an issue with their gameplay. Ditto for carpentry, and other such skills.  Letting a wildman build their own primitive hut isn't weak rp. It's going to lead to more interesting opprotunties then naked fragging.

AdvilPMSevere:
> Humans have been cooking food since the discovery of fire, hundreds of thousands if not a million or so years ago. While garnishing, spicing and etc. is definitely 'civilized', Grug threw his meat on the fire because it made the meat taste better and Grug was happy. That is quite literally one of the first things the absolute earliest humans we can even call _human_ in the first place, started to do.

The given text upon choosing Lunacy Embracer is "You have abandoned your humanity to run wild under the moon. The call of nature fills your soul!"

> ... Not to mention Lunacy embracers aren't million year old unevolved apes. Either, at this point, we take away every single skill from them (that isn't physical like unarmed/wrestling/athletics/climbing) because every argument that can be made boils down to 'well they're wild, they shouldn't care about this' or we accept that they're still people who chose a different lifestyle and the players can choose the degree of 'true' wilderness they wish to adopt.

The given text upon choosing Lunacy Embracer is "You have abandoned your humanity to run wild under the moon. The call of nature fills your soul!" They are quite literally not just "people who choose a different lifestyle", and trying to make them just nuttier Druids completely kills the point of them at all. If people want to play a person who lives in the wild and is a bit crazy or Dendorite, they can do that. Play a Druid, or an adventurer, or any role really. Embracers are and should be another thing entirely.

AdvilPMSevere:
> I really enjoy this role. Part of the enjoyment is the variety of crafting option.s They aren't particularly broken, or problematic. Let me play my savant dendor moron who can farm at a master level as a lunatic. It's not an issue with their gameplay. Ditto for carpentry, and other such skills. Letting a wildman build their own primitive hut isn't weak rp. It's going to lead to more interesting opportunities then naked fragging.

Like my other reply to another comment, I'm just going to say that embracers should really be something apart from "savant Dendor moron who can farm at a master level as a lunatic." That is a perfectly achievable gameplay option as a Druid or Dendorite adventurer, even. 

That being said, the point about having a hut isn't really that unreasonable-I would still argue that building is rather un-wild of someone, but I may leave some carpentry in for said reason. I do want to create subclass options based on patron for LE's in the future, and it may end up that the Dendor subclass will give carpentry, while LE's won't get it by default.

PR #1018, opened 2025-09-20
Grudgebearer Tweaks / Removal of Armor Layers

petwir1:
## About the Pull Request
The Grudgebearer Merc class seems to have been the victim of a previous attempt to change the way armor works. Each piece of armor has layers which with successive hits are shredded and reduce the armor rating. These layers can only be fixed on an iron anvil, which puts them at a serious disadvantage as they cannot fully repair just on a table like everyone else. In addition to that, as I tested armor layers I found a pretty odd bug that again which points to armor layers being a half-finished feature (see below).

This pull request removes armor layers as a concept. It also messes with the armor values of the Grudgebearer armor to bring it more in line with regular plate. I also set the Grudgebearer to spawn with the steel greataxe instead of the steel battle axe because in my mind that is more on-par with the steel goden given for the "mace" option.

I'm more than willing to change things if I may've made the class too strong, just let me know :) this is my first ever PR

## Testing
BEFORE PR:
HelmBefore
^Grudgebearer Helm before damage
HelmAfter
^Grudgebearer Helm after damage
BodyAfter
^Grudgebearer plate (body) after only doing HELM damage... see how the shredding of layers applies to ALL PARTS OF THE ARMOR. So you can get smacked in one place and your whole body becomes weaker- this is the bug I mentioned above.
AFTER PR:
NewArmor
^ This is the modified version of Grudgebearer plate. It's a bit more resistant than fullplate and more durable too.

## Why It's Good For The Game

It makes an underutilized class better and removes a confusing mechanic that only applies to and disadvantages this one armor set.

### DIFF
--- a/code/__DEFINES/roguetown.dm
+++ b/code/__DEFINES/roguetown.dm
@@ -140,7 +140,7 @@
 #define ARMOR_DRAGONSCALE list("blunt" = 100, "slash" = 100, "stab" = 100, "fire" = 50, "acid" = 0)
 #define ARMOR_ASCENDANT list("blunt" = 50, "slash" = 100, "stab" = 80, "piercing" = 80, "fire" = 0, "acid" = 0)
 #define ARMOR_SPELLSINGER list("blunt" = 70, "slash" = 70, "stab" = 50, "piercing" = 30, "fire" = 0, "acid" = 0)
-#define ARMOR_GRUDGEBEARER list("blunt" = 40, "slash" = 200, "stab" = 200, "piercing" = 100, "fire" = 0, "acid" = 0)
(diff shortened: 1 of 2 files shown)


ShadowDeath6:
possibly worth keeping the armor mechanic and just handling it differently

petwir1:
I kept all the code intact just commented it out. So it'll still exist if someone would like to work on it in the future, in it's current form I think it'd be better if it wasn't there.

chemistrymain3:
No. Absolutely not.

petwir1:
> possibly worth keeping the armor mechanic and just handling it differently

Hmm, I was thinking about this. I think the main issue is that it has to be repaired on an iron anvil (of which there is only two in the map that are free to use + it has to be repaired by a dwarf) to regain layers... do you think keeping the layering system and allowing it to be repaired on tables like all other armor would be a better solution?

I would be fine tossing all of the other changes if that change was made.

Bobinater:
800 integ is fucking INSANE.

PR #1035, opened 2025-09-22
Lowers merc slots 8->4

ShadowDeath6:
## About The Pull Request

Too many

## Testing Evidence

I didn’t test but I think it works

## Why It's Good For The Game

mercs are pretty strong combatants and there shouldn’t be as many as there are currently, I think their slots are completely maxed out in almost every round I’m in
Doubly so because there’s an apparent issue of mercs just acting like adventurers but with significantly stronger stats / skills / gear

### DIFF
--- a/code/modules/jobs/job_types/roguetown/mercenaries/mercenary.dm
+++ b/code/modules/jobs/job_types/roguetown/mercenaries/mercenary.dm
@@ -3,8 +3,8 @@
 	flag = MERCENARY
 	department_flag = MERCENARIES
 	faction = "Station"
-	total_positions = 8
-	spawn_positions = 8
+	total_positions = 4
+	spawn_positions = 4
 	allowed_sexes = list(MALE, FEMALE)
 	allowed_races = RACES_ALL_KINDS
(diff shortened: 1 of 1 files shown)


Xaytan:
So they're completely maxed out every round, therefore we should reduce the slot count.

Make it make sense. Please, somebody, make it make sense.

KiteAE:
The are a lot of issues with mercs but I don't think this will fix them, all it will result in is 4 people making an ahelp asking for slot increases to which they will probably get it 70% of the time an admin is on.

roarbark:
this doesn't fix mercs being adv+

RyszardParys:
If a mercenary goes to "act like an adventurer" _whatever that means_, then i think the admins should handle that, not coders. If a mercenary fights antags for free, expecting to get paid after, then that's kinda their problem.

DorfGuy1:
> If a mercenary goes to "act like an adventurer" _whatever that means_, then i think the admins should handle that, not coders. If a mercenary fights antags for free, expecting to get paid after, then that's kinda their problem.

They do dungeons, they fight antags, who said anything about being paid?

PR #1047, opened 2025-09-24
Coin fix

ConstantineII:
## About The Pull Request
Coin fix

## Testing Evidence
Coin fix

## Why It's Good For The Game
Coin fix


RyszardParys:
Coin fix?

Wesreidau:
Was a massive stack of coins actually a problem?

Wesreidau:
Its just that dividing up 21 coins doesn't cause the crash, its dividing up hundreds of coins from the toss command. Coin stacks could simply divide into two piles when thrown and only subdivide into single coins if the thrown pile is 20 or less. Then we have our hundreds of coins in a purse and crash prevention.

PR #1053, opened 2025-09-25
Map spawns for mana-binding collars + crafting rename

NuIIpunkt:
This PR makes mana binding collars and gloves more accessable by spawning 2 collars (one in the dungeon master's bedroom and one in the court magicans bedroom drawer) and leaving an extra pair of mana binding gloves at the magos tower

It also renames the debug item "mother of all gems" to "gem" to remove any potential for confusions during crafting

I think this is a good change because it allows for more RP options as a dungeoneer and court magican (punishing apprentices or actually having alive sorcerers as prisoners). It will no longer be required to gag someone and keep them restrained the entire time and allows them to properly engage in roleplay.

### DIFF
--- a/_maps/map_files/dun_world/dun_world.dmm
+++ b/_maps/map_files/dun_world/dun_world.dmm
@@ -64606,6 +64606,7 @@
 /obj/item/rogueweapon/sword/long/exe/cloth,
 /obj/item/natural/stone,
 /obj/item/storage/belt/rogue/surgery_bag/full,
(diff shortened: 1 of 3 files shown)


NuIIpunkt:
IT DID? The recipe book specified it to be the mother of all gems wich seems to have confused me

NuIIpunkt:
Changed the crafting recipe back, leaving the roundstart spawns wich already do most of the work

PR #1054, opened 2025-09-25
Revert "removes conversion rite"

Just-Sim:
# The issue:
> Deliberate bad faith abuse of the conversion ritual turning wretch into a conversion antagonist (it is not one) to do on-the-spot on-the-street conversions of people, typically guards or known fraglords, who then sneak back in and then start fragging/betraying the keep the moment they're given the opportunity.

# The solution:

image

# Explanation:
I am tired of coders having to be the ones to police the server due to abuse created by bad actors. But I'm even more tired of people coming in and removing stuff instead of addressing the core issue. Deleting stuff should be the last resort solution, when you could instead be working towards an actual fix. 

There was an issue with kneestingers, so I went ahead and found a solution that would work for everyone by giving them a timer, there was an issue with werewolves being too over tuned, so their damage values were lowered, there's no reason to delete a feature when instead you could be fixing the specific issue itself.

# Or an even better solution: 

Don't let a conversion antag snowball, go straight for the converter just like you should be going straight for the lich, or straight for the vamp lord, if you let them snowball, it's a skill issue from the opposing side

This should be an admin situation, not devs having to step in and do it themselves.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defines.dm
+++ b/code/modules/mob/living/carbon/human/human_defines.dm
@@ -138,3 +138,5 @@
 
 	// adds a flag that if we were skeletonized not because we are super dead and rotted, our face can be shown
 	var/ritual_skeletonization = FALSE // ritualcircles.dm path of rituos, prevents the ritual target's name always being unknown ingame. used in human_helpers.dm if( !O || (HAS_TRAIT(src, TRAIT_DISFIGURED)) || !real_name || (O.skeletonized && !ritual_skeletonization && !mind?.has_antag_datum(/datum/antagonist/lich)))
+
+	var/already_converted_once = FALSE // ritualcircles.dm , used to make it so players can't switch around between inhumen gods to stack buffs with conversion rites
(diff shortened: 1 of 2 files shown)


dawsonkeyes (changes requested):
you bring up a valid point as to improving rather than removing, but you have not done that - this is a direct re-add of something we very intentionally had removed; and does not work to solve any of the problems that removing it fixed

conversion should be limited to a specific role or specific means or have some sort of cost to perform the ritual, it should not be a free once-every-ingame-day thing every wretch with ritualist (quite a few classes) can do

kimesik:
> conversion should be limited to a specific role or specific means or have some sort of cost to perform the ritual, it should not be a free once-every-ingame-day thing every wretch with ritualist (quite a few classes) can do

"Quite a few classes" is just two-and-half classes: Heretic, Necromancer (locked to Zizo) and Heretic-Wanderer (which is just a variant of normal Heretic) There's also Towner Witch. It's just that Heretic and Heretic Wanderer are popular and combat-oriented classes on their own.

dawsonkeyes:
you have stated your intent to completely ignore my previous review & we are not merging a simple readd with no strings attached

PR #1056, opened 2025-09-26
Don't uselessly check client for null twice for `update_mouse_pointer()`

tmyqlfpir:
## About The Pull Request

This PR makes an inline optimized for the proc `mob/living/update_mouse_pointer()`.

By only checking client once, it saves on a inherited proc call and additional null check, which makes a difference for hot code such as `update_mouse_pointer()`.

## Testing Evidence
Before:
before
After:
after

The average calls is `828246` per round - this will save up to 1.3 seconds of processing time.

## Why It's Good For The Game

Optimizes the server in a small way.

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -1811,8 +1811,12 @@
 			clear_fullscreen("remote_view", 0)
 
 /mob/living/update_mouse_pointer()
-	..()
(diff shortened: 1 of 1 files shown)


TuskTusk:
You know, these kinds of things get way too little attention, and you doing that to optimise server lag and response is highly appreciated.

PR #1064, opened 2025-09-26
corset changes

PoltavskaPraca:
## About The Pull Request

item has like F in every defense stat, but still counts as a light armor.
been made a subtype of an undershirt, therefore, no values.

## Testing Evidence

image

Screenshot 2025-09-26 224249


Wesreidau:
I disagree.

Go the other way; make it hardened leather protection for the stomach only. That would pair great with corslets. and protect from gutspills.

DorfGuy1:
> I disagree.
> 
> Go the other way; make it hardened leather protection for the stomach only. That would pair great with corslets. and protect from gutspills.

i like this idea, though gutspills are rare...hmmm...having some hot armor is good tho

PR #1073, opened 2025-09-27
[r4m] converts most light sources to movable lights.

chaoko99:
## About The Pull Request

This basically takes a lot of load off the lighting subsystem for something which literally produces no load whatsoever.
There are downsides, namely that blacked out tiles don't contribute light (as normal with movable lights) which is a slight downside in certain circumstances) 
This is NOT done. 

MISC changes: 
removes the snuffing effect when extinguish is called bc they look fucking terrible and run like ass.
## Testing Evidence
it compiles. I need to update some systems to support it.

## Why It's Good For The Game
Less lag = more good

basically just opening this PR for advice, as I'm not sure what to do from here. Updating lights does not yet work.

### DIFF
--- a/code/game/objects/effects/misc.dm
+++ b/code/game/objects/effects/misc.dm
@@ -76,14 +76,15 @@
 	light_system = MOVABLE_LIGHT
 
 /obj/effect/dummy/lighting_obj/Initialize(mapload, _color, _range, _power, _duration)
-	. = ..()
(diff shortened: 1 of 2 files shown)


chaoko99:
can't figure out how to make sconces stop emitting light when they lose their torch.

so any advise si appreciated.

Ephemeralis:
I've pushed a provisional fix that should address orison not functioning correctly (and also causing a lot of lag), as well as restoring its intended range/power scaling.

PR #1075, opened 2025-09-27
Port SOME of the Jannisary changes from RW2

OspreyIsHere:
## About The Pull Request
Ports part of https://github.com/Rotwood-Vale/Ratwood-2.0/pull/72, which includes:
- Jannisaries **lose the +2SPD and the -1 PER** in exchange for **+2 CON and +1 PER**
- They no longer get a Goedendang start, but rather **axe and shield**. This includes jman axes, leveled up to expert if they pick said start.
- Removed 1 of the 2 starting scabbards, added it to the Shamshir start (so you don't just get two scabbards 4noresin)

## Testing Evidence
Little Lines of Code that were literally ripped out from the PR.

## Why It's Good For The Game
Jannisaries were on the less-picked side, mostly because the rider's main attraction was the two dodge expert classes. I've been checking RW2 out and jannisaries are being picked more often since when their scalemail breaks *(As it always does)*,  they at least have some CON to back up the fact they'll bleed like pigs soon enough. Figured it'd be a good test here!

### DIFF
--- a/code/modules/jobs/job_types/roguetown/mercenaries/classes/desertrider.dm
+++ b/code/modules/jobs/job_types/roguetown/mercenaries/classes/desertrider.dm
@@ -1,6 +1,6 @@
 /datum/advclass/mercenary/desert_rider
 	name = "Desert Rider Janissary"
-	tutorial = "The Janissaries are the Empire's elite infantry units, wielding mace and shield. We do not break."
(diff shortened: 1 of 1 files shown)


RyszardParys:
Still no shoes :(

OspreyIsHere:
> Still no shoes :(

image
I mean they **SUPPOSEDLY spawn with them** unless someone fucked w/The atompath

PR #1080, opened 2025-09-29
Frost Magic Rework

Kroto-Ker:
## About The Pull Request

Frostbite has been changed to a sort of 'stacking' debuff. Both frost bolt and snap freeze inflict it. Here's how it works:
If you get hit by any frost magic spell, it inflicts 'Shivering'. You take 25 stamina damage and lose -1 speed for 25 seconds.
Get hit again by frost magic while you have 'Shivering', you get 'Frostbite'. You take another 25 stamina damage. For 6 seconds, move slower, lose -2 speed, and slowly lose stamina over time. This can cause stamcrit.

Shivering and frostbite can not stack, and you can not inflict shivering while the target has frostbite.

**Frost bolt:**
Does 5 less damage to 20 damage total. (Same as spitfire)
Does double damage to NPCs.
Recharge time changed to 6 seconds.
Spell windup unchanged. (About 1 second?)

**Snap Freeze:**
Does 10 less damage to 30 damage total. 
Does double damage to simplemobs.
Recharge time changed to 15 seconds.
Spell windup unchanged.
Spell visuals changed.

## Testing Evidence

image

## Why It's Good For The Game
I believe the other solution of only reducing the timer would render the spell useless, or at least boring! I think this solution makes the spell more interesting while providing it a unique and effective use case.

https://github.com/user-attachments/assets/ee6c9503-034e-4aa1-8f25-bdf0e6bebff6

Also the video shows a version of frost bolt w a 4 second cooldown, I changed it to 6 seconds after looking at the footage.


Bobinater:
Kinda dogshit it slows you so much. Honestly.

zeroisthebiggay:
i like this except for this

> Recharge time changed to 6 seconds. Spell windup unchanged. (About 1 second?)

the spellcasting meta to kill people is to just spam spitfire and wait for spitfire to come off cooldown
giving a second burn damage spell with a low cooldown to cycle between spitfire may inadvertantly give mages a higher effective dps than with frost bolt just being a big slowdown

i can't think of a good way to counterract this but making fire spells purge the frost debuff might lessen the blow

DorfGuy1:
> > i can't think of a good way to counterract this but making fire spells purge the frost debuff might lessen the blow
> 
> Spitfire now fizzles on contact with a player under either of the frost effects, dealing no damage and clearing the effect.

oh wow, so either fire or ice but not both? It'd be neat if it did a little extra damage when it cleared the effects you know, from being hot, cold, hot cold. Elemental combos coool

PR #1082, opened 2025-09-29
In-hand Icon Cache

FrownTown:
## About The Pull Request

• Replaces expensive mid-game resource access and icon operations in obj/initialize with pregeneration at init and a cache
• Used a non-firing subsystem to handle everything on its own code thread and made the code cleaner and easier to read
• Also fixed the weird .json file that had CRLF end formatting for some reason?

## Testing Evidence

• It Just Works™

## Why It's Good For The Game

• Less lag is more gooder


FrownTown:
Fixed a bug with cache path collisions, recommend remerging!

PR #1087, opened 2025-09-29
Add zigboxes

Lioneliuss:
## About The Pull Request

Adds zigboxes, store up to 10 zigs, sold at the purity machine empty for 5 - 10 mammons

## Testing Evidence
it workies
image
## Why It's Good For The Game

saving storage instead of having a handful or two of zigs take up your whole satchel... they still do if you buy them one by one without a zigbox, but at that point, skill issue..


chaoko99:
big day for chronic smokers

PR #1089, opened 2025-09-29
Gives Lunacy Embracers some traits and fixes a typo

chaoko99:
## About The Pull Request

Gives LE's:
Experienced Woodsman: Which helps deal with their absolutely dogshit default perception, which makes me honestly think this was not included by accident. 
Gives them Longstrider, strengthening their territorial advantage and giving them a better chance of getting away without instantly being maneatered and thus, fragged.
Outdoorsman lets them sleep in trees comfortably. I thought it was odd they could not.

also i fixed a typo
## Testing Evidence

string addition, compiles
## Why It's Good For The Game
See description. It basically assists in increasing the home turf advantage embracers are allegedly supposed to have, but do not really feel as if they have.


chaoko99:
from the discord:

image

PR #1091, opened 2025-09-30
fixes sleeves not appearing sometimes

FrownTown:
## About The Pull Request

- Fixes an oversight that caused certain items to generate an extra aux arm sprite that covered their sleeves when the character was facing EAST/WEST. This is due to the aux limb being intended to show the hand poking out of sleeves, but items that had inherited a "sleevestate" from a parent, but had their "sleeved" variable set to null were not handled correctly.

- This will have the side-effect of causing said items to appear to have sleeves that cannot be torn off and used for cloth (although the dismemberment mask will still remove them if someone loses an arm). I think this is a worthwhile tradeoff.

## Testing Evidence

image

## Why It's Good For The Game

It makes things looker more betterer.


RyszardParys:
Yes, while this does fix sleeves not appearing when looking left or right, it doesnt seem to fix the south or north facing sprites.

PR #1103, opened 2025-10-01
fixes an oversight

zeratino:
## About The Pull Request

gives light step to the wretch outlaw

## Testing Evidence

image

## Why It's Good For The Game

wretches should be better than rogues, who have light step


Jackofashepard:
you should add this to bandit knaves too

PR #1116, opened 2025-10-03
Task-based Necromancy (small necromancer QOL change) 

ConstantineII:
## About The Pull Request
Port of: https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/3952

> Necromancers now set a task for their Skeletons, before summoning. This is to give the potential Skeleton player a hint as to what their round might be like.

Unfortunately, not every Skeleton wants to be turned into a Mining Machine for an hour and a half.

## Testing Evidence
494707229-7d097c39-8be8-4053-9fb5-016a6c69a614

## Why It's Good For The Game
Screenshot 2025-10-03 192657

> skeletons won't be hot-dropping into an unknown battlefield. They'll at least get a sense of who they're working for, and what they'll be doing. This helps them invested, and might intrigue other players who might never play a skeleton.

### DIFF
--- a/code/game/objects/items/rogueitems/necro_relics.dm
+++ b/code/game/objects/items/rogueitems/necro_relics.dm
@@ -20,6 +20,13 @@
 		to_chat(user, span_warning("The crystal thrums under your touch, but remains inert."))
 		return FALSE
 
+	// Ask the Necromancer for a task for the skeleton BEFORE the timer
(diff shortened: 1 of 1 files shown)


AdvilPMSevere:
This is great to have to see the name of who summoned you. Only concern I have is that skeletons might start thinking they don't have to do anything but whatever their "purpose" read, when they of course are bound to their masters and must obey them in all things.

PR #1119, opened 2025-10-04
Storage for martyr cloak. Martyr sword sharpness changes.

RyszardParys:
## About The Pull Request

Adds a storage to the martyr's cloak, like all cloaks should have.
Also makes their sword get a little more sharpness integrity when using safe-mode, and becoming high-undullable when committing to their oath.

## Testing Evidence

Soon:tm:

## Why It's Good For The Game

No more 40% sharpness on the martyr's sword after a few parries while they only have 6 or 2 minutes left to live.


RyszardParys:
The sword doesnt return to it's original sharpness (200) when the ult runs out, idk how to fix that. If anyone knows, hmu.

PR #1122, opened 2025-10-04
harpy afterparty init

PoltavskaPraca:


benj8560:
not really sure why lamia knight bandits is a issue. Was the common belief not that we don't race lock people who can be foreigners? Since you know things could work differently in other parts of psydonia and be more acceptable. We allow goblin lords of heartfelt mind you.

For example in a predominantly Lamia inhabited land it would be pretty normal. Proper even.

Harpies not wasting bandit slots on such a stupid build is likely good though but there's a lot of examples of that.

PR #1127, opened 2025-10-05
gives broadheads a little AP

chaoko99:
## About The Pull Request

arrows now have a tiny bit of AP.
## Testing Evidence
var edit

## Why It's Good For The Game

image
this is dumb


anything86:
No. This isn't a tiny amount, this is 10 AP. You didn't even go look into why a simplemob would have armor, you just immediately jumped to buffing broadheads by a substantial amount. Broadheads have no AP because of how their calculation against other players work.

PR #1133, opened 2025-10-07
Adjustcator Upgrade

iMoeMoe:
## About The Pull Request

Some number changes for Adjustcator so the role wouldn't feel like a nothingburger. (Quite frankly worse than regular MAA somehow)

## Testing Evidence

just number changes ser.

## Why It's Good For The Game

rn it's really lackluster compared to the two other roles in the lineup. So just giving it some stat changes and a proper halberd, proper shield skills touchup should be good.


iMoeMoe:
Now they get Spear instead of the Halberd. Defaulted basically

PR #1134, opened 2025-10-07
Priest-from-750-to-500-announcement

iMoeMoe:
## About The Pull Request

750 is a bit much, 500 seems more reasonable

## Testing Evidence

just number changes

## Why It's Good For The Game

rn, it takes like forever for priest to get to announce, so just bring it down to 500 to encourage Priests to change miracle set more to just practice every single set in the Tenth.


iMoeMoe:
erm don't look at the other commits, i was a bit stupiod :3c

PR #1135, opened 2025-10-07
Fix-Powergamey-Exploits

iMoeMoe:
## About The Pull Request

Fixing some powergamey exploits

## Testing Evidence

number changes

## Why It's Good For The Game

I made a severe and continuous lapse in my judgement, and I don't expect to be forgiven.

### DIFF
--- a/code/modules/cargo/packsrogue/Brigand.dm
+++ b/code/modules/cargo/packsrogue/Brigand.dm
@@ -146,7 +146,7 @@
 
 /datum/supply_pack/rogue/Brigand/silverwarhammer
 	name = "Silver Warhammer"
-	cost = 80
+	cost = 180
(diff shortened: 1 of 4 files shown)


benj8560:
sadly this only half resolves the issue. If at any point you can get something from the favor list and get even just _half_ the favor spent back its enough to dupe. 

What we really need is a check for smithing code. If a bandit smelts a bar or forges something the idol refuses it. Bonus points if they get smited by matthios for doing it.

benj8560:
> sadly this only half resolves the issue. If at any point you can get something from the favor list and get even just _half_ the favor spent back its enough to dupe.

this is a thing due to bandits being able to share their favor UI with other bandits. Meaning you can effectively get 8x the favor if all bandits are duping (not a realistic round but possible)

PR #1136, opened 2025-10-07
Psy-Dagger-improved-stats

iMoeMoe:
## About The Pull Request

Kind of annoying having to click it again to get another one

## Testing Evidence

another number change, ser

## Why It's Good For The Game

Since it's a summoned item on hand that can be instantly resummoned, mind as well make it so it doesn't need to break


benj8560:
and keep in mind it draws stamina to summon. There is a reason for it to have durability.

iMoeMoe:
> and keep in mind it draws stamina to summon. There is a reason for it to have durability.

given it can be resummoned for as nothing cost, fair to give it enough DUR to last a whole fight instead of having to replace them every 3 slash

PR #1137, opened 2025-10-07
Consort-Changes-and-Fix

iMoeMoe:
## About The Pull Request

Give Male Consort the Crownstone
Give Female Consort her Satchel
Removing Pickpocketing Skill (Why is this here?) 
Giving them good Medicine and some Cooking, Improving Literacy
Making the stats more auxilliary than ever Combat (why would you fight as Consort?)

## Testing Evidence

Number Changes

## Why It's Good For The Game

Just adding a few things that's supposed to be there, removing something that does not make sense and just fully make Consort an Auxilliary role for pure Roleplayer. Some good Int and Speed so they can maybe dodge assasination and maybe fights if they ever got into one.


Xaytan:
Why is the consort as good at medicine as the professional torturer?

Fatherless1Xx:
A role with zero responsibility shouldn’t have access to a crownstone.

Ephemeralis:
Seems like we're decisively not doing any of this, so.

PR #1140, opened 2025-10-07
fixes invisible scissors

ShadowDeath6:
## About The Pull Request
bugfix

## Testing Evidence

image

## Why It's Good For The Game

bugfix


ShadowDeath6:
> not the fact you cant even store these in any satchels or anything like that

click the satchel with open hand and then place the scissors into the inventory itself rather than clicking the bag with the scissors

PR #1143, opened 2025-10-08
Arcane Trickster: A mage subclass for sneaky bastards

Plum-but-gayer:
## About The Pull Request

Adds the Arcane Trickster subclass, which is a hybrid of the sorcerer and thief adventurers. Similar to Spellblade and Spellsinger, they get 12 spell points and 3 thematic gimmick spells preloaded (invisibility, lesser knock, and vicious mockery). They lose their mage armor in exchange for dodge expert, and in general have equal-or-worse sneaky skills from thief. Their gear is a downgrade from thief, too, but players can invest in conjure weapon/armor if this is a concern for them.

## Testing Evidence

image

image

image

## Why It's Good For The Game

I always remembered there being a spell-thief subclass but I think it got eaten in the big adventurer revamp felinid did back on AP months ago. Spellsinger is fun, but doesn't exactly 1:1 fit the sneaky trickster niche - you lack a lot of climbing and thief skills, so you have to overinvest in virtues to make up the difference. That, and you start with a best in class swift sword and martial spells, which don't really make sense for a thief. My idea with arcane trickster is to fulfill the fantasy of being a sneaky, tricky bastard but not really being much of a stand up combatant. 

I think I have some concerns with a spellcaster that has dodge expert, but I think their lack of an expert weaponskill and low con/STR will make the susceptible to counterplay in the form of feints, charges, and grabs. Worst case scenario is I can make a separate 'utility/trickster' spell list to choose from if we have a suddenly influx of dodgemaxxing spitfire abusers, but that felt a little overkill to do preemptively.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/mage.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/mage.dm
@@ -269,8 +269,8 @@
 		)
 
 	if(H.mind)
-		H.mind.AddSpell(new /obj/effect/proc_holder/spell/invoked/slick_trick_small) //gimmick spell. 1 tile slick trick. honk!
+		H.mind.AddSpell(new /obj/effect/proc_holder/spell/invoked/mockery) //it's back. if they become op feintmeisters remove this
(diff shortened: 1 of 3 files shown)


KiteAE:
> Unless maintainers have any concerns, I'd call this ready to merge. Been up for a bit and I haven't seen or heard any big issues.

done

PR #1146, opened 2025-10-08
Werewolf Limit Cap of 5

iMoeMoe:
## About The Pull Request

Kite asked for a limit of 5 WW so I made limit of 5 ww

## Testing Evidence

https://medal.tv/games/space-station-13/clips/lgeAOuiKabgkB4lB6?invite=cr-MSxLNTUsNDA2NTM0NTI2

## Why It's Good For The Game

Leaving WW unchecked and let them infecting 22+ people is kind of stupid. Put in a hard check so that they don't go bite spree and make Church players weep.


Fatherless1Xx:
This feels like a huge overreaction to something that happens maybe once in a hundred rounds.

Ephemeralis:
Also seems like we're not doing this, and a hard cap wouldn't be the way to address this regardless.

PR #1173, opened 2025-10-13
oceanjak

chaoko99:
## About The Pull Request

Saigas (and all other riding animals, a list that's actually pretty big) can no longer go into deep OCEAN water. This only affects ocean water, and should not stop you from getting your snout deer across a river. Fixing the vampires in the ocean problem comes in another PR, I have ideas.
make a boat if you need to go ocean. The alternative looks really silly and immersion breaking. 

also you no longer splat when you hit water and it breaks your fall.

## Testing Evidence
https://streamable.com/8rdzm1

## Why It's Good For The Game

red-deer-cervus-elaphus-hind-walking-through-shallow-water-in-front-EBK5N2

### DIFF
--- a/code/datums/components/riding.dm
+++ b/code/datums/components/riding.dm
@@ -20,6 +20,9 @@
 	var/ride_check_ridden_incapacitated = FALSE
 
 	var/del_on_unbuckle_all = FALSE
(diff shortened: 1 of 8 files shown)


CVSthePharmacy:
not a rly good change man, just makes another problem......

PR #1177, opened 2025-10-14
Basic loadout item colour preferences

Ephemeralis:
## About The Pull Request

Something about queuing up for dye stations at roundstart shits me through the earth. So, no more. Set your preferred dye station colors in your character's preferences and forget about it.

The color you pick will be automatically applied when you pull it from your stash, regardless of whether you're near a dye station or not.

Note: this does not include "details" or anything like that, because frankly, I just can't be bothered. Also includes some code deduplication so stashing behavior stays consistent across whatever we decide to put it on. It was copy-pasted in like five different places.

## Testing Evidence

image

image

## Why It's Good For The Game

Dyeing shit at roundstart gets extremely old, extremely quickly.

### DIFF
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -893,3 +893,20 @@ GLOBAL_LIST_EMPTY(personal_objective_minds)
 /datum/mind/proc/get_special_person_colour(mob/M)
 	if (special_people[M.real_name])
 		return special_people[M.real_name]
+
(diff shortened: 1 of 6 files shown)


Ephemeralis:
> Would this cover items afforded 3 unique color swatches or not?

No. It is functionally equivalent to the one-color dye that most items use. This is essentially just a quick and dirty PR to facilitate those - doing this "properly" would take a lot more effort..

PR #1183, opened 2025-10-15
Apocrypha gets Peel intent

iMoeMoe:
## About The Pull Request

Balancejak whatever, right now this thing is just worse Creed and could outright does jack against armours and outright worse than Creed options. having an option to actually work against armour instead of hitting praying for armour break while people peel peel peel your shit. This is a GREAT sword that have ONE TILE RANGE. Somehow worse than the smithable Psy Great just because that actually gets peel.

## Why It's Good For The Game

- Changed: Giving it sword peel instead of Executioner's Thrust.


dawsonkeyes (approved):
dont know how i feel about taking away its thrust but w/e

PR #1186, opened 2025-10-15
Alternate Giant Buff Proposal

Plum-but-gayer:
## About The Pull Request
Gives Giant +1 Constitution, and reduces damage from running into doors by 90%. This means they will receive 20 damage per door smack instead of 200. Basic doors take 3 smacks to destroy and 'special' doors (vault, etc) take 6 smacks to destroy - this brings the damage down from 600(!) and 1200(!!) brute to destroy one door to 60 and 120, respectfully. Kool-aid manning into the vault will still probably frac your ribs unless you show up with nice chest armor.

## Testing Evidence
+1 con
image
after smacking into a door once
image
i forgot to take a picture of the 60 damage on her after she smacked into the door 3 times bc she healed bc mobs fall asleep when not directly controlled. 20*3 is 60. Trust.

## Why It's Good For The Game

Oversized is a lot of fun and has a storied history. It existed pre-statpacks as a lucky roll that would give you like +STR +CON - SPEED, but after statpacks were introduced it became problematic bc oversized orcs could hit like 18 STR or something crazy. Some nice contributor made the throw and charge mechanics which rule, but due to balancejakking they were forced to add some utterly ridiculous numbers to them. 200 damage is enough to immediately cripple a limb and put you into critical hit threshold, and 600 is enough to break almost any chest armor available ingame. There's literally **no** situation in which using the charge to break down a door makes any more sense than just smacking it with whatever big weapon you undoubtedly have bc you're oversized and probably running a strength build. This change allows players to actually use the door knockdown feature while still paying a minor price for it. 60 brute to break a door twice as fast and twice as stylish as bashing it seems fair.

+1 Con is because bigass sprites are easier to click, and also the line about being 'stronger and tougher' is a bit misleading if the virtue doesn't actually do that. The con helps with health and charges - having a massive dude be better at knocking people over doesn't feel too gamebreaking. Remember your defend intent, kids.

### DIFF
--- a/modular_azurepeak/virtues/size.dm
+++ b/modular_azurepeak/virtues/size.dm
@@ -8,3 +8,4 @@
 	recipient.transform = recipient.transform.Scale(1.25, 1.25)
 	recipient.transform = recipient.transform.Translate(0, (0.25 * 16))
 	recipient.update_transform()
+	recipient.change_stat("constitution", 1)


SadStarfish7:
I forgive you for jacking my swag. I do think 20 is a bit on the small side, though. Maybe 30? 35?

Plum-but-gayer:
I think 20 adds up with multiple charges but I'm willing to bump it up if maintainers request/we see a problem with charge abuse. Let's keep it low for now and see how it goes - I kept the number low keeping in mind that any number I put there is being multiplied by 3 or 6 (and several stuns) so increasing it even a little  increases the sum total damage quite a bit!

Fatherless1Xx:
What's to stop people taking the virtuous statpack to pick intellectual+giant for +2 stats without losing any in a way that's objectively better than statpacks?
Virtues really should not give stats.

Plum-but-gayer:
Ready to merge upon maintainer approval, also. LMK if anything needs to be changed.

PR #1187, opened 2025-10-15
Removes Lamia racelocks on Soilson, Innkeep, Tailor and Servant

Sniperjolly:
## About The Pull Request

Does what it says on the tin

## Testing Evidence

It's an extremely small change feel free to shoot it down if you think it's tested insufficiently

## Why It's Good For The Game

Lamia are more discriminated against than goblins, kobolds and verminkin. I think they've been in the game long enough to relax a few of these.

The following roles have all races allowed EXCEPT Lamia (and sometimes harpy):
Innkeep, Tailor, Soilson, Jester, Archivist, Druid, Guildmaster, Village Chief

Jester I don't especially want to mess with, the jester meta is alien to me
Archivist and Guildmaster are actually lowkey power classes that I understand wanting to restrict (even if goblins and kobolds can be them)
Druid is pretty tied with Dendor, until they add an Abyssor Druid variant, big whatever.
Village Chief is explicitly a town elder that the townies and even keep RP as having known for a long time, lamias being relatively recent arrivals, sure, why not. There are old goblins and kobolds for this role but no old lamia, whatever.

Innkeep is a retired adventurer, doesn't make much sense they'd be flat out denied the ability to own a business when goblins and kobolds can hold the same position. 
Tailor and soilson are just JOBS someone skilled at the task can have.
Servants again have the whole 'well goblins and kobolds can do it' argument. I don't know what makes lamias specifically the ultimate bottom of the barrel scum of the earth.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/peasants/soilson.dm
+++ b/code/modules/jobs/job_types/roguetown/peasants/soilson.dm
@@ -10,9 +10,6 @@
 	selection_color = JCOLOR_PEASANT
 	allowed_sexes = list(MALE, FEMALE)
 	allowed_races = RACES_ALL_KINDS
-	disallowed_races = list(
-		/datum/species/lamia,
-	)
 	cmode_music = 'sound/music/combat_soilson.ogg'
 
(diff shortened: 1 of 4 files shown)


PoltavskaPraca:
My initial intent was straight up not letting them be ANY city roles with the exception of:
templar
acolyte
mage apprentice
apothecary
bathhouse roles
mercenary

main reasoning for this was/is: they aren't supposed to be locals, they aren't even supposed to own an estate save for what has been provided to them by their superiors
there are no balance concerns, actually, whatsoever. only those concerning le Roleplay
as for the goblins/kobolds having access to those roles? yeah, that's been done a while ago and not by me and on AP (i think).

PR #1199, opened 2025-10-17
Flailing my way downtown 

Nik3321:
## About The Pull Request

Added more dmg to the flail and reduced the drain by half. Also added more force to the necran unique flail and the silver trait.

## Testing Evidence

25 force was feeling very underwhelming for the weapon that has 0 defense, and it was a ABUSRD stamina drain for anyone to use it.

## Why It's Good For The Game

Blunt weapons have been in the limbo for a good while now. These changes will perhaps make it more reliable, with the blunt changes coming it will aid in the work on draining armor integrity, it is a FULLY offensive weapon so should be alright. Gave also silver trait to the holy necran weapon since the goddess whole gimmick is fighting the undead/lickers so it would make sense, also gave it a bit more force than the normal flail.

Not quite sure if I should keep the drain at 2 or 1, testing in my own run 1 seemed fine but I still want to hear some opinions.

Changes based on AP numbers btw

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/flail.dm
+++ b/code/game/objects/items/rogueweapons/melee/flail.dm
@@ -52,11 +52,11 @@
 /datum/intent/flail/strike/smash
 	name = "smash"
 	chargetime = 5
-	chargedrain = 2
+	chargedrain = 1
 	no_early_release = TRUE
 	penfactor = BLUNT_DEFAULT_PENFACTOR
 	recovery = 10
-	damfactor = 1.6
+	damfactor = 1.9
 	chargedloop = /datum/looping_sound/flailswing
 	keep_looping = TRUE
 	icon_state = "insmash"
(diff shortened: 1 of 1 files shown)


Nik3321:
> flail used to be a god weapon (like THE meta weapon, like katar is now) and I think it got nerfed to dirt as a result. +2 for buffing it again

> Okay hang on. 35*1.8 dam factor * 1.4 STR modifier assuming our templars are playing at the strength cap... is 88 damage? can someone check my math? that seems a lil high for the necra flail


Numbers were based on Azure's picks, imo seemed ok for templars since they are supposed to be antag hunters and that weapon has 0 defense, there is a charge time of 5 seconds to get that kind of dmg that can be absorbed by shield, weapon or armor since it has no pen as well. Perhaps tone it down to 32 if it seems too much in game?

Nik3321:
image
 Amount of hits taken to eat through a half-plate + hauberk with the necran flail

PR #1200, opened 2025-10-17
Allows mages to gain Arcane EXP with Spellcasting

GrayRachnid:
## Testing Evidence
image

## Why It's Good For The Game
Mages should be allowed to organically improve their skill through casting.
Making it only through arcane crafting pushes the grinding meta further.


ophaq:
This isn't a bad idea if spell power doesn't scale with arcane level. If I recall right, it's only a 5% spell cast time reduction which is a 30% reduction at legendary.

Xaytan:
There are a couple of spells that scale duration, like invisibility and darkvision, but that's about it.

GrayRachnid:
> There are a couple of spells that scale duration, like invisibility and darkvision, but that's about it.

The alternative is killing 10 imps 10 sprites and 10 crawlers and turning all of it to arcane melds to grind the skill level
I think organically allowing it to go up with spell casting is good.

ShadowDeath6:
> The alternative is killing 10 imps 10 sprites and 10 crawlers and turning all of it to arcane melds to grind the skill level I think organically allowing it to go up with spell casting is good.

it's easier to just get a bunch of rocks and spam out amethysts

PR #1202, opened 2025-10-17
Auto-select subclass when there's only one.

ConstantineII:
## About The Pull Request
https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/3886
Fixes the issue of having to pick a subclass when there's only a single subclass for a job.
## Testing Evidence

https://github.com/user-attachments/assets/514bf006-a2ab-450b-bf0e-b9770286df0b

## Why It's Good For The Game
Having to select a subclass when there is only one is extremely annoying.


CVSthePharmacy:
I for one like to see the tutorial text

PR #1203, opened 2025-10-18
Mastermind wretch

Fatherless1Xx:
The Mastermind is a deserter subclass that's not meant to fight directly. They’re a frail, non-combatant type who excels at in supporting allies with orders, buffs, coordinating attacks, and weakening enemies with debuffs. They also have awesome crafting skills to let them be useful to people. 

Their gimmick is being able to choose any adventurer loadout for equipment, making them particularly hard to sniff out:

image
image
image
image
image

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/wretch/deserter.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/wretch/deserter.dm
@@ -1098,10 +1098,10 @@
 	H.mind.AddSpell(new /obj/effect/proc_holder/spell/invoked/hawks_eyes)
 	H.mind.AddSpell(new /obj/effect/proc_holder/spell/invoked/giants_strength)
 	H.mind.AddSpell(new /obj/effect/proc_holder/spell/invoked/guidance)
-	H.mind.AddSpell(new /obj/effect/proc_holder/spell/invoked/bonechill)
 	H.mind.AddSpell(new /obj/effect/proc_holder/spell/invoked/haste)
 	H.mind.AddSpell(new /obj/effect/proc_holder/spell/self/message)
 	H.mind.AddSpell(new /obj/effect/proc_holder/spell/invoked/mindlink)
+	H.mind.AddSpell(new /obj/effect/proc_holder/spell/self/call_to_slaughter)
 
 	H.set_patron(/datum/patron/inhumen/zizo) // Always valid, must hide their true power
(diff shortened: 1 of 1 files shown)


benj8560:
firstly I entirely like the concept here and would like it to be made to work in some degree. Not saying its a bad PR..but these issues come to mind.

Firstly....theres several people who do the malumite heretic smith gimick. This role pretty much blows them out of the water. They already have all the desirable smithing skills by far. Could I suggest no more then journeyman on those? That's still enough to forge everything but stuff like halberds and full plate. They should be a jack of all trades master of none. 

My second worry is rite of rituos. Wont they just...use that ritual to get attack spells then self buff with all these ''ally'' buffing spells? It's pretty much a no brainer to do that given they have high arcane already. 

Lastly why don't they have any int? This is a SUPPORT crafter mage class. They should have good int.

maker-of-nightmares:
I think apprentice polearms would be a nice addition to them so they can run away faster(all polearms give longstrider) and have SOME means to defend themselves from being rushed down immediately. It's nothing you can't get with 10 minutes of mindless clicking at a dummy either. At present they'd get slaughtered by even NPCs with zero difficulty

Fatherless1Xx:
> I think apprentice polearms would be a nice addition to them so they can run away faster(all polearms give longstrider) and have SOME means to defend themselves from being rushed down immediately. It's nothing you can't get with 10 minutes of mindless clicking at a dummy either. At present they'd get slaughtered by even NPCs with zero difficulty

Like you said, you can grind up to 2, what I don't want is for them to take a virtue to start with 3.

code/modules/jobs/job_types/roguetown/adventurer/types/wretch/deserter.dm
```diff
@@ -240,6 +240,880 @@
...
+					if("Rapier")
+						beltr = /obj/item/rogueweapon/sword/rapier
+					if("Whip")
+						beltr = /obj/item/rogueweapon/whip
+					if("Billhook")
+						r_hand = /obj/item/rogueweapon/spear/billhook
+
+
+	H.adjust_blindness(-3)
+	H.change_stat("speed", 3) /// Not a fighter, even with a virtue, built to be able to escape from fights.
```
dawsonkeyes: ditto - you have given them an absurd skill block and dodge expert, and three speed and constitution

dawsonkeyes (changes requested):

PR #1208, opened 2025-10-18
[TM first] perfomance npc+lava/acid animation tweaks

DelineFortune:
## About The Pull Request

turns off idle afk from most ctrl+c ctrl+v simple mobs around in dungeon (no pathfinder plz)
they can go agro if they only see you 
turns off acid/lava animation 
180 players, man

## Testing Evidence

image

+ dev channel

### DIFF
--- a/code/datums/ai/controllers/rous.dm
+++ b/code/datums/ai/controllers/rous.dm
@@ -21,4 +21,4 @@
 	
 	)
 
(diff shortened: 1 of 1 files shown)


KiteAE:
This made lava/acid have no animations, i would prefer them to have animations.

PR #1210, opened 2025-10-19
[TESTMERGE ONLY] Horrible mob life() optimizations

Ephemeralis:
## About The Pull Request

Trying (probably vainly) to bring mob subsystem times down to help deal with lag during load.

- discovered DeadLife() was calling update_sneak_invis on every dead mob several times a second, which was fetching lumcount each time. i *presume* get_lumcount is an expensive proc because it involves multi-tile lighting operations and as such, have rejiggled logic around so that it is only called when expressly necessary. unsure of the actual impact, need to see in round
- there was a bunch of psydon trait shit on the main life proc for *non-carbon* mobs meaning they weren't doing anything because they were being overwritten by the carbon life overload. i have attempted a good-faith reimplementation of this functionality as it was. please do not scream at me if they suck/are overpowered
- reduced two WHOLE WOUND LIST iterations down to 1 on base mob/living, which existed to apparently facilitate NPC regeneration? or something? i don't know, it had comments in russian. i re-implemented this as a (probably weaker) heal_wounds() call
- heal_wounds() now ends iteration early if it can perform no more new healing

## Testing Evidence

Compiled and ran around in-game for a bit. Sneaking seems to mostly work (though it's as buggy as ever due to our lighting changes)

## Why It's Good For The Game

shit run gooderer (i hope)

### DIFF
--- a/code/modules/mob/living/blood.dm
+++ b/code/modules/mob/living/blood.dm
@@ -164,6 +164,9 @@
 
 			if(blood_volume <= BLOOD_VOLUME_BAD)
 				adjustOxyLoss(blood_volume <= BLOOD_VOLUME_SURVIVE ? 3 : 1)
+			else if((blood_volume > BLOOD_VOLUME_SURVIVE) || HAS_TRAIT(src, TRAIT_BLOODLOSS_IMMUNE))
+				if(getOxyLoss())
+					adjustOxyLoss(-1.6)
 
 	//Bleeding out
 	bleed_rate = get_bleed_rate() // expensive proc, but we zero it on bled-out mobs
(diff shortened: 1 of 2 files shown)


Ephemeralis:
sneak opts look good so far, added a few more experimental ones:

- blood handling now checks to see if you actually have blood volume to work with, and if not, skips several expensive iterations and ends early.
- blood handling now only applies stress events if you're actually alive
- stress handling now uses 1 less iteration (includes a very tiny refactor as well to ensure stress-affecting traits always influence stress events)
- bandage expiry is now only checked if a limb is actually bandaged, using 1 less iteration

bare minimum roguetest wandering testing done for this, it is possible they may break things or alter expected behavior

Ephemeralis:
wrote a whole stupid-ass refactor of the bleeding system to avoid more get_wounds calls inside blood handling. i don't know if this is much good but the tick profiling i was doing inside life() seemed to suggest a modest improvement, so we'll see how it goes.

while i was in there i also changed how the bleeding status effects are called, so they're no longer trying to add/remove up to 3 status effects every single tick.

stress processing now only happens if you're actually conscious as well

none of this should really have any obvious player-facing influence (except you can, if very unlucky, end up unconscious for longer periods of time at really low blood (i removed a somewhat-costly unconsciousness check). please scream if anything fucky related to bleeding happens

PR #1165, opened 2025-10-12
Church game loop

DelineFortune:
## About The Pull Request

New organs:
Stomach - T1: ROT_EATER; T2: ROT_EATER, WILD_EATER; T3: NASTY_EATER, ORGAN_EATER, maxHealth x2. 
Lungs - T1: WATERBREATHING; T2: BREADY, ZJUMP, WATERBREATHING; T3: BREADY, LEAPER, ZJUMP, NOBREATH, LONGSTRIDER. 
Liver - T1: SEA_DRINKER, toxTolerance 15; T2: TOXIMMUNE, ZOMBIE_IMMUNE, SEA_DRINKER, toxTolerance 35, toxLethality 0.008; T3: TOXIMMUNE, ZOMBIE_IMMUNE, SEA_DRINKER, NOHUNGER, toxTolerance 50, toxLethality 0.005, maxHealth x2. 
Heart - T1: SHOCKIMMUNE; T2: +KNEESTINGER_IMMUNITY; T3: CRITICAL_RESISTANCE, SHOCKIMMUNE, KNEESTINGER_IMMUNITY, HEAVYARMOR, maxHealth x2
Eyes - T1: low-luminosity glow; T2: SEE_MOBS, darksight 5; T3: SEE_MOBS|SEE_OBJS|SEE_TURFS, darksight 10, high lighting visibility.

Artifacts & Tools:
Malum’s Tool - instant 1-bar forge + on-spot smelt 
Necra’s Censer -  AoE cleaning.
Dendor’s Endless Hose - ±100 water/nutrition per click, auto-bless, growth modes: none/mature/kill (hidden produce coz shit is a bit op), config UI.
Noc’s Phylactery - bind by blood (30s), show target coords + distance.
Eora’s Heart - shows unique partners this round (self/target).
Ravox Trace Lens - reveals species traces from blood/prints/reagents nearby.
Abyssor’s Rod - baitless fishing, fish-only loot
Xylix Gloves - theft chance/range boost, XP multiplier while worn.
Pestra Surgical Multitool - 7 modes (scalpel/saw/hemostat/retractor/bonesetter/suture/cautery) quick staples on incisions.
Astrata’s Star - ritual to let a target choose a patron.

**Quests**  
Quest Tokens (10 types) 

Antag check
Use on a frllow player
If target is an antagonist, owner gets +250 Favor + bonus: +500 (Vampire Lord/Lich), +250 (Werewolf/Lesser Vampire Lord).
[...]

PR #1211, opened 2025-10-19
Church remap.

DelineFortune:
## About The Pull Request

My https://github.com/Scarlet-Reach/Scarlet-Reach/pull/980 but with fixed issues. Its preparation to add https://github.com/Scarlet-Reach/Scarlet-Reach/pull/1165 (thats been tested but required minor changes)

## Testing Evidence

Church
see https://github.com/Scarlet-Reach/Scarlet-Reach/pull/980

Old church's p;ace is replaced with garden
image
image
image

North lands:
Dendor's tree 
image
image

Lighthouse
image
image
image

### DIFF
--- a/code/controllers/subsystem/rogue/devotion.dm
+++ b/code/controllers/subsystem/rogue/devotion.dm
@@ -150,22 +150,9 @@
 	H.verbs += list(/mob/living/carbon/human/proc/devotionreport, /mob/living/carbon/human/proc/clericpray)
 
 	if(HAS_TRAIT(H, TRAIT_CLERGY))
-		var/exclude = FALSE
-		var/title = ""
-		if(H.job)
-			title = "[H.job]"
-		else if(H.mind?.assigned_role)
-			title = "[H.mind.assigned_role]"
-
-		if(length(title))
-			var/ltitle = lowertext(title)
-			if(ltitle == "priest" || ltitle == "martyr")
-				exclude = TRUE
-
-		if(!exclude)
-			if(!H.mind.has_spell(/obj/effect/proc_holder/spell/self/learnmiracle))
-				var/obj/effect/proc_holder/spell/self/learnmiracle/L = new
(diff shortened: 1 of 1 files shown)


DelineFortune:
> split the warden changes off into a separate PR, we'll need to discuss those

lets test it a bit to let people get ideas or inspiration because im not going to remap it again

GrayRachnid:
I don't like the church just becoming a concentrated 40 bushes in town you can grind for resources now :sob:

Plum-but-gayer:
we should replace the public park with a giant dungeoneer prison

<---- running for office btw

coffeecoffeecoffee:
i'm sorry, i am just so not a fan in the slightest of moving the church far out of town

chemistrymain3:
please god merge this

zeratino:
I NEED IT!!!

Wintermote:
I'm assuming this is the church map that just went live, here are initial thoughts and feedback:

It's a massive improvement on the old church.

Problems I've run into:
The kitchen is too small (It would be nice for 2 people to have the room to prep at once), a well or cistern would be nice to have for water supply (rather than grabbing water from the library... pool...).

The library... probably shouldn't have water. Moisture + paper = mold and rotted books. It *looks* nice, but you'd really want a very dry and warm library to preserve them. Just an aesthetic issue.

The infirmary/medical area is very small, it would be nice if it were a little larger or perhaps if all the beds were moved down the stairwell. My biggest problem though, is that I'm wishing for space for a chest to hold all the assorted equipment that gets left behind by people, and a washbin both for cleaning and for putting leeches into (some people seem to have a thing for spawning dozens upon dozens of leeches - which I toss all into the washbin). I can drag them there and re-arrange things, it's just a bit time consuming, especially since the church is much larger now.

Altogether this new church makes the church actually feel important and soulful. If nothing at all changed I'd still be happy with it.

PR #1214, opened 2025-10-19
Rogue mages can buy the chalk and magebags keep thralls start with

Fatherless1Xx:
## About The Pull Request

Let's bandit mages buy the chalk/bag that keep thralls begin the round with.

## Testing Evidence

image
image

## Why It's Good For The Game

I don't really see a reason why one of the mage's coolest toys and gameplay loops is locked exclusively to the keep?

### DIFF
--- a/code/modules/cargo/packsrogue/Mage.dm
+++ b/code/modules/cargo/packsrogue/Mage.dm
@@ -178,3 +178,11 @@
 	name = "Cinnabar Ore"
 	cost = 10
 	contains = list(/obj/item/rogueore/cinnabar = 1)
+
+/datum/supply_pack/rogue/Mage/Summonerkit
+	name = "Beginner's summoner kit"
+	cost = 30
+	contains = list(
+		/obj/item/storage/magebag/starter,
+		/obj/item/ritechalk,
+	)


ThinkArbies:
keep mages have this gameplay loop because otherwise they'd have nothing to do in the course of a 3 hour round.

is it really a good idea to encourage antag mages to do nothing all round except sit and grind alchemy and enchanting? they won't be contributing to the round or actually be doing any _antagonizing_ while they do these notoriously time consuming things, and the fruits of their labours only serve to give other antags a massive power spike once its done. so now not only do you have an iconoclast, but now have an iconoclast with enchanted boots to give them longstrider and a katar that's enchanted to slow you while they hit you, among other nasty possibilities.

knoxon1:
Fun fact, the Hordemaster already lets mages buy silver daggers. I've been using this to enchant bandits for a while now, giving bandits darkvision and longstrider isn't as OP as people may believe. But also, for the record, I think Bandit Mage is the best enchanter in the game, above even Court Mages, due to the infinite supply of cinnabar and scrolls in the hordemaster.

GrayRachnid:
None of these items are locked to the keep.

Summoning pouch:
2 hide 1 fiber, skincrafting. Can be done at a low level.

Ritual chalk:
1 cinnabar 5oz mana potion. Can be done at a low level.

I don't see the point in making these easier than they already are to get. I don't like the ideas of bandits speedbuying and speedenchanting their equipment before the keep has managed to finish the roundstart RP. 

I think it's better that you just make bandit mages spawn with it that make it purchasable, then they'd just buy 11 pouches and don't have to venture out for the resources.

Ephemeralis:
Concerns about mass oversupply seem quite valid here given just how much stuff is in those pouches, so we're probably not gonna do this one.

PR #1215, opened 2025-10-19
royal champ. int nerf + climbing skill reshuffle

zeratino:
## About The Pull Request

this pr reduces the statbuff from RC's intelligence from a +3 to a +1
it also sacrifices their master climbing in exchange for giving expert to every class

## Testing Evidence
image

## Why It's Good For The Game
image

lets take into account this build. ignoring the town buff cuz it doesnt change intelligence - all this does is just change the 16 int to 14.

this is a good change because it makes RC into the not-meta, and allows other - weaker knight classes (such as mounted and heavy) to shine with their intelligence boost

the climbing adjustment was a requested change 
image
- tldr: people only pick RC cuz its utility in climbing is way too good compared to other classes
- this still makes the climbing ult there - it just makes it so that other knights can (kinda) keep up

### DIFF
--- a/code/modules/jobs/job_types/roguetown/nobility/knight.dm
+++ b/code/modules/jobs/job_types/roguetown/nobility/knight.dm
@@ -188,7 +188,7 @@
 		/datum/skill/misc/riding = SKILL_LEVEL_APPRENTICE,	
 		/datum/skill/combat/wrestling = SKILL_LEVEL_EXPERT,
 		/datum/skill/combat/unarmed = SKILL_LEVEL_EXPERT,
-		/datum/skill/misc/climbing = SKILL_LEVEL_JOURNEYMAN,
+		/datum/skill/misc/climbing = SKILL_LEVEL_EXPERT,
 		/datum/skill/misc/reading = SKILL_LEVEL_APPRENTICE,
 		/datum/skill/misc/athletics = SKILL_LEVEL_JOURNEYMAN,
 		/datum/skill/combat/knives = SKILL_LEVEL_JOURNEYMAN,
@@ -280,7 +280,7 @@
 		/datum/skill/combat/bows = SKILL_LEVEL_EXPERT,
 		/datum/skill/combat/wrestling = SKILL_LEVEL_EXPERT,
 		/datum/skill/combat/unarmed = SKILL_LEVEL_EXPERT,
-		/datum/skill/misc/climbing = SKILL_LEVEL_JOURNEYMAN,
+		/datum/skill/misc/climbing = SKILL_LEVEL_EXPERT,
 		/datum/skill/misc/reading = SKILL_LEVEL_APPRENTICE,
 		/datum/skill/misc/athletics = SKILL_LEVEL_JOURNEYMAN,
 		/datum/skill/combat/knives = SKILL_LEVEL_JOURNEYMAN,
@@ -373,7 +373,7 @@
 	traits_applied = list(TRAIT_MEDIUMARMOR, TRAIT_DODGEEXPERT)
 	subclass_stats = list(
 		STATKEY_STR = 1,
-		STATKEY_INT = 3,
+		STATKEY_INT = 1,
 		STATKEY_END = 2,
 		STATKEY_SPD = 2)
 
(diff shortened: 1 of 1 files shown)


zeratino:
image

coffeecoffeecoffee:
> Why should knights who are fullplated fragbeasts get better climbing than most bandits?

I... Would have assumed??? Most bandits get some level of climbing skill. The issue though is moreso that royal champ is the ONLY knight class that gets good climbing skill, making it the only viable choice over the other knight classes rn. I don't know about the balance of them being able to climb whatsoever or not compared to _other_ classes, only that royal champ was disproportionately good due to literally having access to entire areas of the manor that other knights just can't defend or get to in the event of a break-in. You could alternatively just take it away from them and let no knights climb.

With the sheer frequency of non-stop break-ins I'm tempted to suspect most bandits or wretches get good climbing skill but idk

bunnyywoo:
I don't agree with the int change.
RC is supposed to be the best duelist class.
It's literally the only thing that makes the class good at fighting, besides feinting you aren't winning. It's a good class to duel in 1v1s, even 1v2s, but at mass fights it sucks, you don't do much compared to the other heavier knights.

Even if you're taking away 2 points of stats, atleast consider putting them somewhere else, no? You're just nerfing the class to the ground like this.

And sincerely I don't care about climbing, I think it is fucked up that the rest of the knights don't get it, I'm just annoyed at the int getting nerfed so much just because.

zeratino:
I'll consider putting them somewhere else upon maintainer request, however I'm holding off on that because of the specialization into swift weapons which carries an entirely different bag

The intent of this pr is to bring all knight classes closer together while letting individual archetypes shine in their field, not to compare against bandits 

Once again, the change was made so that mounted and heavy knights have their own niche.

I don't think that it ruins the class since you can still feint most enemies reliably, but have difficulty against VLs and bandits with the necklace - which is good cuz they're major antags

bunnyywoo:
Still, the royal champion is **too** weak without the feints, I get that they're annoying to deal with if you don't know how to counter that, but imo it will fuck up its gameplay and the class will shift out to be the duelist/ranged class to solely the ranged class with some melee speed.

Even then, Not adding the +2 stats lost to something else like con (which is pretty low) or end for example, will make them quite shitty.
I'm not asking you to add them to speed, but atleast consider other stats if you will go through with this.

Fatherless1Xx:
I think it’s important to understand how knights got to where they are now. At first, there was only one, and they were deliberately made absurdly powerful. They were the Baron's most trusted enforcer who was OOCLY expected to NEVER betray them. Then, for reasons I still don’t fully get, the number was raised to three. And now we’ve added a knight captain. So not only are there more knights than ever before, but the captain is stronger than any knight has been in the past.

The strange part is, knights don’t even have any responsibilities that set them far apart from men-at-arms. They’ve been powercrept into a weird spot where it's difficult to justify their inclusion, since there's no IC or OOC consequence for oath breaking. Still, this change feels like a step in the right direction.

Wotziz:
Good PR sire, although I'd redistribute the two points you took from INT into SPD and STR

ThinkArbies:
no, this kinda fucking rots, dude.

yeah, RC is better because of its easier access to climbing compared to other knight subclasses in a current era of climbing mobility, but why is the solution to that to nerf it?

it's also a pointless nerf as it turns out, too. journeyman climbing, which every other knight class gets (with the exception of KC who only gets apprentice for some weird reason), is enough to climb most walls in the keep if I'm not mistaken.

only RC, with its master climbing, is able to do all the new climbing shenanigans. expert does allow them to do most things, if I'm not mistaken, with some limitations.

HOWEVER, to actually do any of the new climbing stuff, you can only wear **light** armor, which is the RC's specialty. the other knight roles (and RC if you pick the medium armor choice) do not actually benefit in any conceivable way from having expert because they still can't engage with any of the new mechanics enables. expert **might** allow them to climb the sheer brick walls? but to my best recollection spent after many many rounds of monkeying around the keep, there are **no** 1-tile high brick walls that provide any meaningful boost to mobility in the keep, making that point moot.

and who honestly thinks it's a good idea to nerf **INT** on the fucking **dodge class?** how does that make any justifiable sense?

this entire PR's rationale is steeped in misinformation and half-truths, and doesn't actually do anything to settle the issue with climbing/how much nicer RC feels to play over every other knight role. the only equalization it does is making sure RC sucks like every other knight subclass.

vitoras228322-ai:
> I think it’s important to understand how knights got to where they are now. At first, there was only one, and they were deliberately made absurdly powerful. They were the Baron's most trusted enforcer who was OOCLY expected to NEVER betray them. Then, for reasons I still don’t fully get, the number was raised to three. And now we’ve added a knight captain. So not only are there more knights than ever before, but the captain is stronger than any knight has been in the past.
> 
> The strange part is, knights don’t even have any responsibilities that set them far apart from men-at-arms. They’ve been powercrept into a weird spot where it's difficult to justify their inclusion, since there's no IC or OOC consequence for oath breaking. Still, this change feels like a step in the right direction.

How does this even correlate to the royal champion nerf?
Anyhow - I believe that nerfing duelist capabilitie of a duel-oriented class which can really get pwned in one or two well-landed feints/baits iss... rather odd. At least - compensate for the lost int with a point in str or speed.

iMoeMoe:
slop balancejack 
-1

iMoeMoe:
Most of the time you're fighting those people the only equality you have are the int, it's a 50-50 chance because all three already have high INT(VL, Bandit with Neck or Sawbones, Liches just have high INT to begin with) if they are to fight. That's the only niche you can use to fight on EQUAL footings against people. Nerfing it is absolutely meaningless, this whole pr is just salt balancejacking no matter what way you look at it.

PR #1218, opened 2025-10-19
Loadout changes + crafting additions

NuIIpunkt:
## About The Pull Request

Recently I came across an item called the "feldsher collar" wich made me wonder and code dive. I found the feldsher hood and robes but noticed that the robes use the physiker robe sprites and the hood wasnt even craftable
I changed this and did some suggestions from players while I was at it

Following changes to the loadout options:

-Feldsher collar was removed (It was the only one out of 3, why was it there to begin with??)
-Apron has been added
-Leather Belt has been added
-Added the Tome of Justice

Crafting recipes:
-Both physiker hood and robes are now properly craftable (the latter was craftable as the "feldsher robes" why?? i dont know)
-Made the Feldsher clothes finally optainable (They are just red versions of the physiker robes and their sprites already exist, why why why????)
-Plague mask is now craftable (We are now ready for the disease updateTM)

## Why It's Good For The Game

Its more options for clothes that already were in the code but unavailable otherwise (give it purpose or remove it entirely)

The tome of justice is also an important item that you should have easy access to, bookshelf spawns are random so tough luck trying to get them otherwise

Crafting wise, everything was inconsistent, I made it a little more consistent


NuIIpunkt:
> If you are adding the apron, can you add the cook apron as well please? You can't even sew a cook's apron... you have to just find it lmao.

Let me see if the apron is craftable if it isnt ill add it to since its just a cooking apron basically

NuIIpunkt:
Im not sure if I can add all of it but ill see wich pieces arent armoured, if I add artificer gear it will be the last addition to prevent making this PR bigger than intended

Lord-Preston:
Would it be possible to add the black leather belt as an alternative option too? I love the black leather belt but the brown is the default on most classes.

PR #1221, opened 2025-10-19
Add Arcane Warden subclass

Sniperjolly:
## About The Pull Request

Added a third warden subclass. Every change is noted in the code but notably

- Uses ranger as a base including armor
- gets hand-advisor level magic, with T2 Arcayne, 15 spell points and novice magic skill
- Trades 1 speed for 2 int
- combat skills are the worse of either subclass except knives (it has journeyman like ranger instead of apprentice like forester)
- has a wood staff and apprentice polearms
- has mana potion and alchemy book
- one more medicine than standard ranger
- apprentice alchemy
- expert literacy

Current bug: Warden helm/hood selection happens twice on every subclass now, but doesn't provide a second hood or helm so it's just annoying. Absolutely no idea how to fix.

## Testing Evidence

Screenshot 2025-10-19 140257
Screenshot 2025-10-19 140424
Screenshot 2025-10-19 140435

## Why It's Good For The Game

In response to a code bounty. Gives players more flexibility. More diversity in the Warden clubhouse.

Reworking the class is possible, maybe lowering some of the non-combat skills so they don't have expert tracking, swimming, climbing and athletics since mages are usually some kind of bookworm. Swapping journeyman knives for journeyman polearms? Lightening up their armor even further, maybe giving them a basic gambeson instead of a padded one. Going down in spellpoints is possible but maybe a bit painful since they don't have any other way of defending themselves. 

Potential issue is even apprentice alchemy getting a bit out of control surrounded by the forest and wardens, but they have to make a mortar/pestle and cauldron from scratch. Even if they did, their potion stash would be vulnerable to getting raided by antags and robbed off fallen wardens so it could create additional conflict.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/warden.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/warden.dm
@@ -206,7 +206,7 @@
 /datum/advclass/warden/arcane
 	name = "Arcane Warden"
 	tutorial = "A magically talented individual who volunteered for the wardens. You might have been a foreigner or a witch looking to get in the town's good graces, or an apprentice who liked practical work more than study."
-	outfit = /datum/outfit/job/roguetown/warden/ranger
+	outfit = /datum/outfit/job/roguetown/warden/arcane
 	category_tags = list(CTAG_WARDEN)
 
 	traits_applied = list(TRAIT_MAGEARMOR, TRAIT_ARCYNE_T2)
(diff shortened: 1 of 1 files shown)


Lioneliuss:
a good way to differentiate them could be using the blackhorn bow as a base and making a simple yet thematic staff for them

AdvilPMSevere:
I really don't get this. Not just because it seems out of place for a warden to have magic in the first place, but also it feels like magic is way too common already. It would be cooler for rp potential if magic were a bit rarer overall imo

Sniperjolly:
Just fixed the bug with loadouts. I hear and understand the problems with adding more magical subclasses (though I think 'keep/garrison' roles are just mage, apprentice mage, prince and hand right now?) but I'm largely doing this as coding practice in response to a bounty. I'm working on spriting up a 'blackwood staff' right now... time to figure out how to integrate sprites.

I think slashing the free access spellpoints and giving them a few built in flavorful spells is a good idea, I'll look into druids in a bit (I've literally never played one)

PR #1224, opened 2025-10-21
doll mini update and jak

ShadowDeath6:
## About The Pull Request

flavortext now correctly shows -2 STR
adds some skin colors
dolls can now install skill exhibitors in themselves or other golems

## Testing Evidence

image

## Why It's Good For The Game

spruces the race up a bit, hopefully makes it more appealing to pick
also addresses a rather noticeable pain point with the races while hopefully not deflavoring golems


coffeecoffeecoffee:
i think crit weakness as a doll enjoyer is actively flavorful and fun and part of the enjoyment of the race
personally i'd be in favor of them simply getting something else cool in exchange (because yeah, they have basically a LOT of downsides right now) rather than just removing a downside that's extremely fun rn

Ephemeralis:
Crit weakness is kino and pretty thematic to dolls, so I doubt we'll be removing it. Other stuff looks good, though.

Ephemeralis:
Come to think of it, their speech marker being "crackles" has always felt a little odd to me. Makes sense for golems, maybe.

coffeecoffeecoffee:
> > personally i'd be in favor of them simply getting something else cool in exchange (because yeah, they have basically a LOT of downsides right now) rather than just removing a downside that's extremely fun rn
> 
> I'm open to suggestions, I just think crit weakness is too much of a detriment (being stabbed in the head with a dagger with crit weakness can randomly skullcrack and instakill you)

could let them level up on their own. i think most people consider that vastly more 'just annoyance' feature and would prefer that lifted over the crit weakness, which is integral to their identity of fragile porcelain dolls

ShadowDeath6:
image

new

PR #1225, opened 2025-10-21
Fixes the descriptions of Wild-Kin and Doll races to properly reflect their actual stat changes

Sniperjolly:
## About The Pull Request

Wildkin have +1 PER and +1 CON in game. Their description says they have +1 CON and +1 END. This fixes that.

Dolls have -2 STR, +1 SPD and +2 INT in game, but their description says they have **-1 STR,** +1 SPD and +2 INT

## Testing Evidence

None needed

## Why It's Good For The Game

People building characters now accurately know what stats they're getting when picking a species.

I'm fine with the opposite being done, changing wildkin to have END instead of PER and giving the dolls a smaller strength malus, but 
1) Wildkin having perception makes more sense than them having endurance, they have on average better smell, hearing or both over humans.
2) I didn't want to balancejak Dolls. If someone wants to have them buffed just let me know.


Sniperjolly:
It now just changes Wild-Kin since Doll PR got merged

PR #1227, opened 2025-10-21
re-restricts psydon

chemistrymain3:
lorewise you aren't allowed to be an open psydonite in any position of authority or nobility, as they're relics- a tolerated heresy that's considered harmless

this was undone, for some reason. i'm fixing it now.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/courtier/physician.dm
+++ b/code/modules/jobs/job_types/roguetown/courtier/physician.dm
@@ -6,7 +6,6 @@
 	total_positions = 1
 	spawn_positions = 1
 	allowed_races = RACES_APPOINTED_OUTCASTS_UP
-	allowed_patrons = list(/datum/patron/divine/pestra)
 	allowed_sexes = list(MALE, FEMALE)
 	display_order = JDO_PHYSICIAN
 	tutorial = "You were a child born into good wealth--but poor health. \


chaoko99:
the difference here is that people don't pick psydon to do something interesting.

They overthrow the church, or pick psydon because the trait is good.

Samman166:
Probably shouldn't also restrict a bunch of non noble roles along with it. Vet/sgt/courtphys/etc.
If you wanna stop the high nobility being psydonite sure I don't see the need myself but I get the reasoning. But this is going beyond that and over correcting

Redrover1760:
Lorewise the inqusitior is a psydonite so I don't understand how the lore is supposed to work here.

chemistrymain3:
> Good that you reverted councilor, but I think jester, court phys, dungeoneer, veteran, hand, and mage apprentice are too far imo.

Resolved. Mage apprentice, jester unrestricted. Court phys locked to pestra, as the description indicates. Hand I am not adjusting.

iMoeMoe:
peak, +1

Blargar:
Could you also allow Psydon knights ? It just feels right to play a selfless, all-loving protector and be a Psydon worshiper. With the majority of the court being tennite you will also have reasons to keep your faith to yourself or be  chastised. I'll give a thumbs-up !

chemistrymain3:
> > Unimportant to me.
> 
> If you are removing Psydonite dukes you have to remake the influence screen and system so it doesn't impact it. Don't leave deadweight around dude

i do not care

PR #1232, opened 2025-10-22
Crimson Curse Virtue

EvenInDeathIStillServe:
## About The Pull Request

adds a virtue that turns you into a stray vampire, a stunted vampire spawn.
their regenerate ability doesn't restore limbs, and has a 5 minute cooldown
replaces the disguise and regenerate verbs with spells
locked from all church and inquisition roles, along with duke

removed critical weakness from them, and added a stat debuff for vampires caught out in the sunlight

this virtue will be PQ locked to 20 once i add functionality for that

## Testing Evidence

image
image
image

## Why It's Good For The Game

will fill this out later

### DIFF
--- a/code/modules/spells/roguetown/vampire.dm
+++ b/code/modules/spells/roguetown/vampire.dm
@@ -86,14 +86,16 @@
 	. = ..()
 	if(!ishuman(user))
 		return
-	for(var/datum/status_effect/debuff/silver_curse/silver_curse in user.status_effects)
-		to_chat(src, span_warning("My BANE is not letting me REGENERATE!"))
-		revert_cast()
+	if(user.has_status_effect(/datum/status_effect/fire_handler/fire_stacks/sunder) || user.has_status_effect(/datum/status_effect/fire_handler/fire_stacks/sunder/blessed))
+		if(prob(50))
+			to_chat(user, span_warning("I cannot regenerate while engulfed in holy fire!"))
+		else
+			to_chat(user, span_warning("Holy fire smothers my attempt to mend these wounds!"))
 		return
 	var/datum/antagonist/vampirelord/VD = user.mind.has_antag_datum(/datum/antagonist/vampirelord)
 	if(VD)
 		if(VD.disguised)
-			to_chat(src, span_warning("My curse is hidden."))
+			to_chat(user, span_warning("My curse is hidden."))
 			revert_cast()
 			return
 		if(VD.vitae < 600)


EvenInDeathIStillServe:
> A "virtue" that gives you critical weakness, makes you valid and trades free healing via sleeping for needing to attack people to drink their blood to heal. This should probably be a vice imho.

when i looked at the code, it seemed like they could heal by sleeping in coffins

GrayRachnid:
> A "virtue" that gives you critical weakness, makes you valid and trades free healing via sleeping for needing to attack people to drink their blood to heal. Ontop of all the other inherit flaws vampires have like sunlight/silver/fire/etc. This should probably be a vice imho.

Flavor doesn't need to be balanced. Let people eat shit for the sake of rp.

Plaguewalker:
> Should also probably be restricted from certain roles. Reasonably, people in positions of public power (Dukes, hands, marshalls, knights) and religious ones (acolytes, priest, martyr, etc.) probably wouldn't be able to hide it forever and thus would've been outed as vampires a long time ago.

Most people don't carry round to round interactions as canon, treating it all as procedural. The excuse of they'd be found out long ago is kind of a moot point given that. Duke is getting barred from having the virtue at least and church and inquisition is already barred as well.

JonathanTheII:
I do think so far this is pretty well balanced. Critical weakness isn't that rare and it makes for a more interesting drawback than the -4 to spd and str(which sounds like a gigantic debuff to me. Would it be anywhere or just if you're outside?). The fact that you have to undisguise to regenerate and there being a cooldown makes it pretty situational, which is good. Also of note is that they do get their Transfix ability, but it seems to not be working, most likely because they dont get any blood magic skill.

Way more risky and less useful to take than Hollow Lyfe, but it seems good like this.

Plaguewalker:
> I do think so far this is pretty well balanced. Critical weakness isn't that rare and it makes for a more interesting drawback than the -4 to spd and str(which sounds like a gigantic debuff to me. Would it be anywhere or just if you're outside?). The fact that you have to undisguise to regenerate and there being a cooldown makes it pretty situational, which is good. Also of note is that they do get their Transfix ability, but it seems to not be working, most likely because they dont get any blood magic skill.
> 
> Way more risky and less useful to take than Hollow Lyfe, but it seems good like this.

You don't need to undisguise to regen

Plaguewalker:
> > > Lock it out of duke too, imo. A vampire court magos makes sense, but vampire duke can go bad real fast when the VL shows up.
> > > Make sure they don't get the vampire skin armor either
> > 
> > 
> > already done
> 
> can u also give them the hollow lyfe -2 to con universally

How about no?

RyszardParys:
The virtue is not as good as people make it seem, sure you get no breathing and no eating, nopain (u cant see which limb of yours is damaged on the doll lol) ,you cannot level your skills (no sleep), you can't just farm vitae from mobs, your stats are nuked in the day and if you try to undisguise during the day you get a seizure as if you'd be hit with the psydonian crankbox. And even if you do undisguise you get uh... night vision? Zizo followers weep. 😂 
Regenerate uses 600 vitae and a normal person has around 500 in them. You'd either have to beg for blood like some **loser** on Bubberstation medbay (pls giv bloodbag :3) or just frag 2 people to get one heal back.
They do NOT need more nerfs.

Fatherless1Xx:
Consider giving them infinite stamina. This could be enough to push this into frail statpack but for virtues into genuinely worth considering territory.

NotGoldenglow:
>[...] you cannot level your skills (no sleep) [...]

To be fair, you can still sleep in coffins.

RyszardParys:
Not sure if it's a feature or a bug, but when the virtue user is put on fire while disguised during the day, their disguise breaks and you get astrata's big dong shoved down your hole (you get burned in the sun even more due to not being disguised).

Plaguewalker:
> Not sure if it's a feature or a bug, but when the virtue user is put on fire while disguised during the day, their disguise breaks and you get astrata's big dong shoved down your hole (you get burned in the sun even more due to not being disguised).

Part of that is due to how ass rendingly slow the fire pat out system is, because pretty much if you're not actively indoors you're fucked six ways to sunday

Xaytan:
Recommending a line in the virtue along the lines of:

_"The townsfolk consider you a monster. If discovered, you will be in grave danger."_

Right now people are apparently LOOCing at people that they aren't allowed to kill them for being a vampire because it's just a virtue and not a "real antag".

RyszardParys:
> I think this is what this trait needs, yes. Or some kind of mood debuff for witnessing vamp actions to nudge people in the direction of finding them offputting.

What kind of actions would that be anyway? Transfix shouldnt be noticable (if it even works bruh...) and a normal person can bite/drink too, some gamers use this to drain your blood in combat quickly without even being a vampire.

chemistrymain3:
PR's pretty good to be merged, imo.

PR #1238, opened 2025-10-22
Fix rend being mondo ultra whackshit post blunt AP changes

Ephemeralis:
## About The Pull Request

Direct all frag discussion about this change to my secretary, thanks.

## Testing Evidence

image

## Why It's Good For The Game

Maybe we shouldn't have an intent that deals 2.5x damage with no AP malus whatsoever?


roarbark:
this intent probably should go in general... it really exacerbates the issue of gamers owning people because they spent the first half hour of the game rping instead of farming a complete set of armour and full face protection, or just destroying non combat classes and so on and so forth. now there might be something speaking for keeping rend in the game but honestly i can't come up with an argument for it right now

PR #1248, opened 2025-10-23
[TESTMERGE ONLY] Re-enable static lighting only on actually static light sources

Ephemeralis:
## About The Pull Request

Chaoko's holy work of defuckulating practically all of the mobile light sources means we can provisionally attempt to bring static lights back on things like wall candles, sconces, etc. This should reset the general appearance of the game's lighting back to 90% what it was before the optimizations, with hopefully only a little bit of overhead in the process.

Will need to see what kind of load this produces.

## Testing Evidence

Without static lighting (current game post-optimizations):

dreamseeker_gy0ZZxeJwu

With static lighting (what this PR re-adds):

dreamseeker_ovECMK2uiJ

## Why It's Good For The Game

The performance improvements from the light update were *really* important, but we have sacrificed quite a lot of sensical lighting to get there. This gets us a bit closer back to how things were while hopefully keeping most of the performance benefits.

### DIFF
--- a/code/game/objects/lighting/_base_roguelight.dm
+++ b/code/game/objects/lighting/_base_roguelight.dm
@@ -5,6 +5,7 @@
 	fueluse = 60 MINUTES
 	bulb_colour = "#f9ad80"
 	bulb_power = 1
+	light_system = STATIC_LIGHT
 	var/datum/looping_sound/soundloop = null // = /datum/looping_sound/fireloop
 	pass_flags = LETPASSTHROW
 	flags_1 = NODECONSTRUCT_1


chaoko99:
This may have some small issues with torches however because the current system disables the ligthing of the sconces in favor of the slighting of the torches.

this has weird side effects. I would avise some code to replace the lighting of the torches in sconces by disabling light on for the torch while it is in there, and copying its light info to the sconce.

Ephemeralis:
Pretty much negligible lighting load from the first TM round so far. Will look at fixing the torch thing then this should be an easy add.

PR #1249, opened 2025-10-23
Heretic irremovable armor rites give mending

Fatherless1Xx:
## About The Pull Request

Rites that give  irremovable cursed Armor also gives the target the mending spell.

## Testing Evidence

image
## Why It's Good For The Game

Not being able to remove/repair your plate makes this intentionally overpowered antag armor useless for anything except a final, suicidal last stand. This change aims to make it viable beyond that by giving it some sustainability.

### DIFF
--- a/modular_azurepeak/code/game/objects/items/ritualcircles.dm
+++ b/modular_azurepeak/code/game/objects/items/ritualcircles.dm
@@ -535,7 +535,6 @@ var/forgerites = list("Ritual of Blessed Reforgance")
 		loc.visible_message(span_cult("Great hooks come from the rune, embedding into [target]'s ankles, pulling them onto the rune. Then, into their wrists. Their lux is torn from their chest, and reforms into armor. "))
 		target.mind?.AddSpell(new /obj/effect/proc_holder/spell/invoked/mending)
 		to_chat(target, span_notice("A new understanding fills my mind... I can mend that which breaks."))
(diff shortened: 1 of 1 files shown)


Fatherless1Xx:
> > > No. You repair it by performing the ritual again. This PR attempts to solve a problem that does not exist.
> > 
> > 
> > Forcing antags to hide because of arbitrary cooldowns hurts the flow of the round. There shouldn’t be any mechanic that pushes them into idling just for the sake of a timer. Especially in this case where non-hostile interaction is basically impossible because of your valid armor.
> 
> This isn't what you said in "why it's good for the game". 🤨 Did you know this and intentionally leave this opinion out? You said it was "useless for anything except a final, suicidal last stand. This change aims to make it viable beyond that by giving it some sustainability."

Nah, I didn’t know. But the supposed fix isn’t much better. Telling people to just wait a day to fix the armor I shredded in five clicks with peel intent is exactly why everyone plays heretic wanderer almost exclusively. Also part of the reason I didn't know about this. Why would anyone play heretic_vanillia when heretic_wanderer doesn't have to worry about this at all?

benj8560:
the fact that this cant be removed to be easily repaired is this armors only real downside...its the best armor in the game. Something needs to fill the gap in balance if you remove the ONLY DOWNSIDE it has.

PR #1251, opened 2025-10-23
Mending spell repairs coverage if integ is 100%

Fatherless1Xx:
## About The Pull Request

Mending spell will fix peeled coverage if the integ is already 100%.

## Testing Evidence
If below 100% mending will not fix coverage:
image
If it's at 100% it will:
image

## Why It's Good For The Game

It's what the spell is supposed to do.

### DIFF
--- a/code/modules/spells/spell_types/wizard/utility/mending.dm
+++ b/code/modules/spells/spell_types/wizard/utility/mending.dm
@@ -30,7 +30,6 @@
 			else
 				to_chat(user, span_info("[I] appears to be in perfect condition."))
 			return
-
 		var/repair_percent = 0.25
 		repair_percent *= I.max_integrity
 		I.obj_integrity = min(I.obj_integrity + repair_percent, I.max_integrity)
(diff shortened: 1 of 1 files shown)


Fatherless1Xx:
> Holy Based Batman now I don't have to worry about peel monsters with their rapiers ruining my armor instantly. I can also see a combat mending being useful to fix a peeled section and just totally fucking a peel monsters rhythm up as they do less damage to armor than mending fixes(Nominally about 20-30 damage to an armor piece which can be insta healed by mending)

chargetime = 5
recharge_time = 20 SECONDS
Even in the best-case scenario where their integ is 75% or higher, someone would have to charge and cast this twice to restore their coverage and they’d still need 20 full seconds of taking no armor damage for the cooldown between casts. Chill. Even 1% below full integ will make it not touch coverage.

maker-of-nightmares:
It still could fuck up their rhythm which isn't a bad thing just means its harder to be a peel monster which is ALREADY the best class in the game

PR #1253, opened 2025-10-23
Proper fix for roles not mouseoverable in player panel

ConstantineII:
## About The Pull Request
Proper fix for jobs not displaying mouseoverable in player panel. 
https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/4226

## Testing Evidence
503022506-e6304273-ed22-479d-bf58-1d5a273ad893

## Why It's Good For The Game
I don't really know why but it's probably better than my previous fix.


WeNeedMorePhoron:
You can probably revert the Magician's Associate and Magician's Apprentice name change now. 

The reason this fixes that bug is that by encoding it properly it stop those job names from breaking the onmouseover script.

PR #1263, opened 2025-10-25
Flailing my way downtown part 2

Fatherless1Xx:
## About The Pull Request

Done off the back of: https://github.com/Scarlet-Reach/Scarlet-Reach/pull/1199

## Testing Evidence

image

This is just a numberjak. There's no new code.

## Why It's Good For The Game

When the heretical armors and weapons were added, they were intentionally designed to be overpowered. But with the recent changes to swift journey the gilded flail is no longer the best flail. Swift Journey now has the same force and can be paired with a shield. The extra reach is of the gilded flail is nice but it’s not enough for it to not just be a side-grade.

This PR is meant to put the gilded flail back in it's rightful place as the objectively best flail.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/flail.dm
+++ b/code/game/objects/items/rogueweapons/melee/flail.dm
@@ -36,6 +36,7 @@
 
 /datum/intent/flail/strike/matthiosflail
 	reach = 2
+	damfactor = 1.9
 
 /datum/intent/flail/strikerange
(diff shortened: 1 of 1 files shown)


Fatherless1Xx:
It's been pointed out to me that the gilded flail has no charge time and I'm retarded for not realizing this. It's a mace larping as a flail.

PR #1269, opened 2025-10-25
Golem only weapon change proposal

Nik3321:
## About The Pull Request

This PR makes so the Golem Knuckles that are usually a drop from the zizo constructs becomes a craftable recipe for the smiths.

## Testing Evidence

image
image
image

## Why It's Good For The Game

Well I know this is a usual loot but it has a few issues I wanted to address:

Firstly it's only worth 3 mammons so it won't ever get bought back to the merchant.

Anyone that doesn't have bleed immunity won't even be able to carry it so only way would be using a sack UNLESS you are a golem.

Golems are a VERY unique race with their own way to level up, own way to heal so might as well used content already in the game to give them a different unique weapon. Surely it's not SUPER good and that's on purpose, it's just flavourful.

### DIFF
--- a/code/modules/roguetown/roguejobs/blacksmith/anvil_recipes/weapons.dm
+++ b/code/modules/roguetown/roguejobs/blacksmith/anvil_recipes/weapons.dm
@@ -765,7 +765,7 @@
 /datum/anvil_recipe/weapons/bronze/golemknuckle
 	name = "Golem Knuckle"
 	req_bar = /obj/item/ingot/bronze
-    additional_items = list(/obj/item/ingot/bronze, /obj/item/ingot/bronze, /obj/item/ingot/steel)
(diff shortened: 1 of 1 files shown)


Ephemeralis:
I don't think we're gonna do a 1 bronze ingot recipe for a 11 wdef (equal to a shield and a 50% improvement over base bronze knuckles) golem only weapon. The intent is clearly for them to be something you go out and adventure to get.

benj8560:
golems are pretty goated. They dont really...need buffs. They are easily the strongest race in PVP right now.

Nik3321:
> golems are pretty goated. They dont really...need buffs. They are easily the strongest race in PVP right now.

It's not a buff, the item is already in game, just making it so it's actually used.

ShadowDeath6:
would be cool actually seeing golems use the item every so often

PR #1273, opened 2025-10-26
Gives forester warden iron bracers

Plum-but-gayer:
## About The Pull Request

Gives wardens a slight buff as they felt a little too much like pushovers from both recent experience and what I've heard from others.

Forester:
1. Gets iron bracers, up from leather.

## Testing Evidence
Forester
image

## Why It's Good For The Game

The forester had a big issue where the chain vest didn't cover arms, meaning they were a non-dodge expert fighter with leather bracers as their only arm protection. They now have full 'medium' armor coverage.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/warden.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/warden.dm
@@ -58,7 +58,7 @@
 		/datum/skill/combat/crossbows = SKILL_LEVEL_JOURNEYMAN,
 		/datum/skill/combat/wrestling = SKILL_LEVEL_APPRENTICE,
 		/datum/skill/combat/unarmed = SKILL_LEVEL_APPRENTICE,
-		/datum/skill/combat/knives = SKILL_LEVEL_EXPERT,
+		/datum/skill/combat/knives = SKILL_LEVEL_JOURNEYMAN,
 		/datum/skill/combat/axes = SKILL_LEVEL_APPRENTICE,
 		/datum/skill/misc/athletics = SKILL_LEVEL_EXPERT,
 		/datum/skill/misc/climbing = SKILL_LEVEL_EXPERT,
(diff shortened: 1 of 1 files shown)


Plum-but-gayer:
we also still need an iron bracers sprite

roarbark:
wardens spawn with like 50 mammon in their meister - this might have been meant for them to upgrade their gear with. since that is now frontloaded you should probably reduce that to 10 so wardens are reliant on wages again, and all the character conflict and politics that comes with that

Teragenic:
Also, I will say Wardens SHOULD not be onpar/better than Wretches.

Wretches USUALLY work alone, they do not have the backing of the Keep and the entire industrial complex of the town.

The gear change just seems also not fitting for FORESTERS and RANGERS. I feel like heavy leather would be better, but now it matches what wretches get.

> the recent trend of wardens building up their outpost with walls can be indicative of them feeling like if they don't do that sort of defensive behavior they'll just get picked off by a wandering WW/LE/Zizite Knight while playing lute outside the warden tower.

I just feel like the actual warden tower itself is a HORRIBLE defensive position. You are literally surrounded by trees that most 'good' LEs and WWs will abuse.

Zizite Knights too with the Woodwalker Trait. 

HOWEVER. There is literally nothing stopping you from just... Not manning it or just not building much/not engaging/etc.

HOWEVER, your argument also of > picked off by a wandering WW/LE/Zizite Knight while playing lute outside the warden tower.

This is ANY role, this is not an argument. Any role can be picked off doing that, this is not JUST wardens.

Teragenic:
Wardens have decent roundstart gear that CAN carry them throughout the entire round, Bodkins being the only 'real' need for the archers. 

They already get natural Longstrider outside in the forest. Which most Wretches DO NOT (Lamia players...). They are meant to be a skirmisher class and a blocking force.

This also puts them now even BETTER against also just random adventurers. This is slowly becoming TDM based in my opinion. Directly comparing us vs them. You should tap into the cycle of hunting/etc or just other things, instead of just free gear + Skill buff.

Plum-but-gayer:
people apparently have strong opinions about this.

PR now just gives wardens iron bracers, so they have full armor coverage. That's all.

PR #1276, opened 2025-10-26
Campfire QoL improvements

Ephemeralis:
## About The Pull Request

Campfires now automatically keep cycling hand-warming on them after you first click, with an interval that should help ensure you're constantly benefitting from the passive healing they give you if you're doing nothing else that interrupts do_after actions.

The status effect that campfires give also no longer has the miraculous healing examine text.

## Testing Evidence

image

## Why It's Good For The Game

Makes a source of non-red healing a little more viable to actually use, and should hopefully encourage people to do a little bit of roleplay while healing up as well.


YakumoChen:
this kills magic campfire

Ephemeralis:
It should make magic campfire better, if anything? The healing behavior is base campfire stuff, not specifically magic.

PR #1285, opened 2025-10-27
Updates the armoury, brings back the acid pit.

KiteAE:
## About The Pull Request

Updates the armory for the Knight and adds one for the Squire (AP ports, will credit them when I find which PR added them) Updates the MAA armory to be more compact, adds in a summoning circle room beneath the keep.

Also removes the drawbridge and moves the Kybraxor further inside the keeps gate as i added an acid pit underneath. 

## Testing Evidence

image

image

## Why It's Good For The Game

Armouries have needed updated for a while now, summoning circle room is needed to contain powerful summons, acid pit is probably not good for the game but its funny and gives the gatemaster something more to do.

### DIFF
--- a/_maps/map_files/dun_world/dun_world.dmm
+++ b/_maps/map_files/dun_world/dun_world.dmm
@@ -4345,9 +4345,9 @@
 /area/rogue/indoors/town/garrison)
 "bil" = (
 /obj/structure/rack/rogue,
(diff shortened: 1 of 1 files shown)


Quink-Fix:
Can we get the squire armoury changed to an armoury key, since squires have them? Anyone with a manor key can kinda just go in there and take whatever they want.

PR #1293, opened 2025-10-29
The Old Ways: Witch subclasses & shapeshifting improvements

Ephemeralis:
## About The Pull Request

Witch is in a weird spot. On one hand, zad form is fucking incredible to the point of being abjectly annoying on anyone with half a brain, and that alone imparts a lot of power. On the other hand, they're locked to T1 spells (which are broadly abysmal for anything interesting save for perhaps invisibility) and are otherwise fairly mediocre alchemists. This aims to change that.

- **Flying witch shapeshift forms will now knock the witch unconscious for 15 seconds if their wild shape is killed.**
   - This does not apply to ground-based forms. Read below.
- Witch now has subclasses. You pick these by choosing the Witch role as a Towner, similar to how other specializations work.
- Old Magick represents the "default" witch experience, except that it now has T2 arcyne instead of T1, and 9 spellpoints instead of 6 (essentially 12 with arcyne potential).
   - Old Magick witches receive the Aerosolize and Guidance spells for free.
   - Witches still do not have mage armor and I am not going to give them it. Shapeshifting is their defensive option, and they're not intended to be a combat capable role as anything other than an especially mobile support.
- The Godsblood subclass has been added, which is essentially a witch with T2 miracles.
   - Godsblood witches have half the normal devotion for their tier (500 instead of 1000) and **cannot pray to regain devotion**. In exchange, they have double the major cleric devotion regen rate at 2 per tick instead of 1.
   - Godsblood witches are themed after stealing or siphoning away miracles via the Old Ways. What are the Old Ways? Who the fuck knows - make that shit up for yourself. They are not intended to be traditional clergy, because you know, you're a witch. Characters with strange relationships to their patron divinity may find this alluring.
- The Mystagogue subclass is a mixture of T1 arcyne and T1 miracles and is worse than either of the previous specializations at any specific type of magic, in exchange for having access to both.
   - Mystagogues have cratered devotion regen (0.5) and cannot pray, similarly to Godsblood witches.
   - They also only receive 6 spell points.
- All witches can now pick from 1 of four shapeshift "second skin" forms at round start, from a list of zad, cat, bat and lesser volf.
   - Lesser volf is a smaller, faster, runtier simplemob wolf with barely a third of the damage of a normal volf.
   - Before some of you shit your pants, it is in no way comparable to the druid volf form shift and even has a different description to denote this. You will **NOT** be combat capable if you pick lesser volf. It is for flavor and speed and that is IT.
   - If you pick bat, don't be surprised if the Inquisition starts asking questions. That one's on you.

Other QoL changes:
- Minor shapeshifting forms no longer cause gib explosions (SFX and visuals). This applies to the witch shapeshifting forms and also VL's bat form.
- Fixed a runtime on shapeshift spells that had no stored mob saved to their datum. Which was apparently all of them.
- Fixes a looped Destroy() call on shapeshift holders that fixing the above runtimes revealed.
- ALL shapeshifts now create a track on shapeshift commencement and when you shift back into human form. Both commencement and exit tracks are marked as belonging to your *human* mob and thus can be used to track you via marking.
   - This track is quite low difficulty and most characters will be able to spot it.
- Witch shapeshifts now apply speed bonuses to their forms. Flying shapeshifts have a 25% speed increase, ground-based forms have a 50% speed increase.
- Witch shapeshifts no longer show the player's true name.
- Added knockout functionality to the shapeshift holder, with visible_messaging to show what just happened (for both player and observers).

## Why It's Good For The Game

Bitches love witches, but witches are stitches and kind of britches to play. This should hopefully entice people into playing them more.

### DIFF
--- a/code/game/objects/shapeshift_holder.dm
+++ b/code/game/objects/shapeshift_holder.dm
@@ -56,14 +56,14 @@
 		shape.death()
 
 /obj/shapeshift_holder/proc/shapeDeath()
-	//Shape dies.
+	//Shape dies.		
 	if(source.die_with_shapeshifted_form)
 		if(source.revert_on_death)
 			restore(death=TRUE)
 	else
-		restore()
+		restore(knockout=source.knockout_on_death)
 
-/obj/shapeshift_holder/proc/restore(death=FALSE)
+/obj/shapeshift_holder/proc/restore(death=FALSE, knockout=0)
 	restoring = TRUE
 	qdel(slink)
 	if (stored)
@@ -76,7 +76,11 @@
 		the_evidence.track_type = "expanding animal tracks into humanoid footprints"
 		the_evidence.ambiguous_track_type = "curious footprints"
 		the_evidence.base_diff = 6
-		
+		if (knockout)
+			stored.Unconscious(knockout, TRUE, TRUE)
+			stored.visible_message(span_boldwarning("[stored] twists and shifts back into humen guise in a sickening lurch of flesh and bone, and promptly passes out!"), span_userdanger("I quickly flee the waning vitality of my former shape, but the strain is too much--"))
+			to_chat(stored, span_crit("...DARKNESS..."))
+
 	if(shape && shape.mind)
 		shape.mind?.transfer_to(stored)
 	if(death)
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/witch.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/witch.dm
@@ -116,17 +116,17 @@
 			ADD_TRAIT(H, TRAIT_HERESIARCH, TRAIT_GENERIC)
 
 /obj/effect/proc_holder/spell/targeted/shapeshift/crow/witch
-	die_with_shapeshifted_form = TRUE
-	convert_damage = TRUE
+	knockout_on_death = 15 SECONDS
 	shifted_speed_increase = 1.15
 	show_true_name = FALSE
+	die_with_shapeshifted_form = FALSE
 
 /obj/effect/proc_holder/spell/targeted/shapeshift/bat/witch
 	overlay_state = "bat_transform"
-	die_with_shapeshifted_form = TRUE
-	convert_damage = TRUE
+	knockout_on_death = 15 SECONDS
 	shifted_speed_increase = 1.15
 	show_true_name = FALSE
+	die_with_shapeshifted_form = FALSE
 
 /obj/effect/proc_holder/spell/targeted/shapeshift/cat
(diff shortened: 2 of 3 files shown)


Fatherless1Xx:
This is cool, but I’d point out that the whole reason we bumped Message up to Tier 2 was because people were using Witch to play like spydrones. No class that can fly should also have access to Message or Mindlink. I would personally suggest just giving them set spells rather than spell points.

Ephemeralis:
Don't TM this for the moment. Apparently the cat shapeshifts break the game in a hysterically convoluted way. Ugh.

Xaytan:
I think having zad form death instantly kill the witch is a little harsh with how fragile zad form is.

I agree it shouldn't be ablative armor, but instant death is an overcorrection I think and will probably lead to witches not bothering with it due to how easy to kill it makes them. Putting them in crit would be plenty IMO.

Ephemeralis:
> These changes make witch into a combat oriented class, which doesn't fit the towner role.
> 
> I personally think that, at the least, witch should be given limited slots if these changes are added. Like the thug towner class, which has the same niche of being a towner capable of combat. I'd say something like 4 slots.

What about these changes veers them towards being a combat class, in your view?

Ephemeralis:
Shapeshift holders store your old body (human form), so it should still be keeping tabs on that.

Ephemeralis:
I've expressly watched no less than 5 witches since this was TM'd attempt combat and eat shit for it. I think we're gonna be okay.

ChickyRooster:
Bug I have noticed, changing to a form makes you sometimes unable to read emotes after, your own or others, changing Z-level fixes this

GrayRachnid:
> Bug I have noticed, changing to a form makes you sometimes unable to read emotes after, your own or others, changing Z-level fixes this

I believe this bug is related specifically to dullahans transforming

Kroto-Ker:
> What about these changes veers them towards being a combat class, in your view?

Tier 2 arcyne allows access to combat spells, and most miracle sets are geared for use in combat. These witch changes make them roughly the same level as an adventurer class mage, in terms of effectiveness. 

But, magic combat itself is kind of shitty. Realistically they probably won't fuck up too much stuff just by having access to combat spells, but it does definitely make them a more combat oriented class, rather than entirely crafting focused like old witch.

Redrover1760:
My Witch bat form still has my witch's name on it.

Also for some reason it set my name to something that isn't what I set when I joined too.

Ephemeralis:
- Flying forms now knock you out for 15 seconds and no longer convert damage, since damage conversion was full-healing you (admin style). This also fixes the hunger thing mentioned above.

KiteAE:
full merging to cut down on open PR's

PR #1299, opened 2025-10-30
[TESTMERGE ONLY] more grungly optimizations

Ephemeralis:
## About The Pull Request

- disables crediticons subsystem. this nukes the end-of-round head scroll (which breaks more than half the time anyway) in exchange for removing a fairly large source of lag when lots of new characters join the round. crediticons is **expensive as FUCK** and its really low priority meant that it'd often immediately hit overrun, spend all of its budget, then pause either forever for the rest of the round or introduce high spikes of TIDI at a later date when the MC tried to catch it back up during lulls.

## Why It's Good For The Game

faster speed gooderer


RyszardParys:
Good idea

PR #1300, opened 2025-10-30
adds the humble scarf

zeroisthebiggay:
## About The Pull Request

adds an item which is literally just the pulled-down hijab, with no hood
it's called the scarf and does following, differentiating itself from the hijab

- it doesn't cover the mouth so you can drink water without having to pull down a whole ass hood
- it doesn't have a hood therefore there's nothing to toggle
- it doesn't provide any protection at all, it's literally just for drip; for people who like the pulled-down hijab sprite but don't like the other features it has
- it can be equipped on the mouth slot!!!!! so you can do things like mix and match with other hoods or wear it with a fucking coif!!!!!

also it's a loadout item and you can craft it for dirt cheap

## Testing Evidence

i promise i tested it

## Why It's Good For The Game

it's just harmless drip, and i think the mouth slot is sorely underutilized
people who like the hijab's looks but hate the functionality doing shit like preventing you from drinking will probably love this
idk the balance implications if any of allowing an item in the mouth slot but it's realistically no different from a zig
all i can think of is that it doesn't let you spit


zeroisthebiggay:
> provide a screenshot

don't mind the hijab sitting on the floor it was just apart of my testing loadout

image

PR #1302, opened 2025-10-30
Balancejacks Intellectual and Breaks All Game Balance

GrayRachnid:
## About The Pull Request
Removes the +1 INT stat bonus from intellectual. One line change.

## Testing Evidence
It's a one line change I apologize.....

## Why It's Good For The Game
This virtue has no downsides yet many, many, upsides. I considered replacing the +1 INT with the GOODWRITER trait that archivists get but then I looked and realized you're already getting so much. +3 to Reading skill(this takes you to legendary if you're at journeyman!) and 3 extra languages (Something most people don't even care about and probably use to pick Otavan to listen in on the inquisition only). Not to mention being able to QUICKLY assess your enemies from range and see their stats/skills. This virtue is too good in its current state.

TL;DR: Free +1 INT (most desired) with no downsides (And +3 reading!)

### DIFF
--- a/modular_azurepeak/virtues/utility.dm
+++ b/modular_azurepeak/virtues/utility.dm
@@ -125,7 +125,7 @@
 /datum/virtue/utility/linguist
 	name = "Intellectual"
 	desc = "I've spent my life surrounded by various books or sophisticated foreigners, be it through travel or other fortunes beset on my life. I've picked up several tongues and wits, and keep a journal closeby. I can tell people's exact prowess."
-	custom_text = "Maximizes Assess benefits with a bonus of the target's Stats. Allows the choice of 3 languages to learn upon joining. +1 INT."
+	custom_text = "Maximizes Assess benefits with a bonus of the target's Stats. Allows the choice of 3 languages to learn upon joining."
 	added_traits = list(TRAIT_INTELLECTUAL)
 	added_skills = list(list(/datum/skill/misc/reading, 3, 6))
 	added_stashed_items = list(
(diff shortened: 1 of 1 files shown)


coffeecoffeecoffee:
it's... it's a virtue, it's supposed to be an upside without a downside? that's the point of having a virtue?

ShadowDeath6:
every virtue is a free (thing) with no downsides aside from extremely game-changing virtues like hollow life
the main thing about intellectual is that it's competing with both arcyne potential and sentinel of wits on magic characters, this just makes it a never-pick option

GrayRachnid:
> every virtue is a free (thing) with no downsides aside from extremely game-changing virtues like hollow life the main thing about intellectual is that it's competing with both arcyne potential and sentinel of wits on magic characters, this just makes it a never-pick option

I disagree that it's a never-pick option. It's +3 reading, instant accurate asses and 3 languages. That's already a lot, the +1 INT is bloat.

That's my stance on it and I don't think it will change. If maintainers disagree they'll close this PR and I'll move on with my life.

ophaq:
The +1 INT is not a huge deal, and is basically the best part other than the language side of this virtue. Hell, the languages are just a bonus and I would rather have the int. Also, it's INTellectual so I honestly think it should stay, especially because it is one of the only virtues that boosts a negligible state other than statpacks. I typically take this virtue for the added crafting chance but I can see how it could affect combat feinting in a minor manner.

Fatherless1Xx:
Virtues shouldn't give stats. That's what statpacks are for.

PR #1306, opened 2025-10-30
Removes Middle Aged from Veteran

Kroto-Ker:
## About The Pull Request

Restricts veterans to old age only.

## Testing Evidence

image

## Why It's Good For The Game

A lot of people want it. Players go veteran as middle-aged to fight _with_ the garrison instead of training them.


A-Distinguished-Axolotl:
Haven't had trouble with this myself and it feels like an unnecessary restriction. A veteran being 40-50 years old sounds reasonable to me thematically. If people are being KC 2 that's more of a fail rp issue and should be QC'd instead of making a role change.

zeratino:
good change

PR #1307, opened 2025-10-30
Adds Otavan Armor & Kazengun Drip to Craftables!

ophaq:
## About The Pull Request
This PR basically just adds a ton of recipes that Kazengun have, or are in the loadout and are missing to both leatherworking's unique recipes and to sewing recipes. Also, it adds the Otavan helmet, plate armor, yoroku masks, and steel mempo to smithing recipes.

I focused on making these harder to make, such as the Kazengun recipes requiring a lot of skill and silk. They are EXPENSIVE on purpose because they are foreign designs that are complex.

Also, I fixed a few recipes like the geta and daisho not being able to be crafted.

## Testing Evidence
kazendripproof

## Why It's Good For The Game
I have been wanting smiths to make me Otavan helmets for ages, plus if you are an Otavan knight you can't upgrade your cool armor to better quality. Also, all of the cool Kazengun drip is locked behind certain classes, some of which are not played that often. More recipes for the tailor and hunters to make too, which gets them more business. I don't see why people would object to this PR, other than it removing some 'exclusivity' from certain classes. But then again, these are expensive and high skilled items not everyone is going to be able to get immediately!

### DIFF
--- a/code/modules/roguetown/roguejobs/blacksmith/anvil_recipes/armor.dm
+++ b/code/modules/roguetown/roguejobs/blacksmith/anvil_recipes/armor.dm
@@ -918,7 +918,7 @@
 	craftdiff = 3
 
 /datum/anvil_recipe/armor/steel/mempo
-	name = "steel mempo"
+	name = "Steel Mempo"
 	req_bar = /obj/item/ingot/steel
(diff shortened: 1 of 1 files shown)


awawaawawaawa:
It's peak sire.

PR #1309, opened 2025-10-31
Adds Bulk Bomb Recipes || Craftable Smokebombs

ophaq:
## About The Pull Request
Yeah, you heard me. Craftable bulk bottle bomb recipes and you can now make smoke bombs! It's the same recipe as the bottle bomb, but it's salt instead of coal. Ever wondered why that fog was white?

## Testing Evidence
boomproof

## Why It's Good For The Game
We love committing crimes against humenity. Also, we really needed a way to make smokebombs.


Ephemeralis:
Craftdiff 3 or 4 for the smokebombs is sorely needed. I don't think we need every kobold or greenskin player having ten smokebombs on them at all times.

PR #1320, opened 2025-11-01
Containers do not autoclose on move

Fatherless1Xx:
## About The Pull Request

This PR makes it so things like belts/satchels do not automatically close when you move.
This does have the downside that inventory containers like wood bins will not close either, but you still need to be next to it to get anything from it. I think the good outweighs the bad.
## Testing Evidence

image
## Why It's Good For The Game

Makes it easier to do inventory management. You can move while interacting with backpacks.

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -950,10 +950,6 @@
 			var/mob/living/L = pulledby
 			L.set_pull_offsets(src, pulledby.grab_state)
 
-//	if(active_storage && !(CanReach(active_storage.parent,view_only = TRUE)))
-	if(active_storage)
-		active_storage.close(src)
-
 	if(!(mobility_flags & MOBILITY_STAND) && !buckled && prob(getBruteLoss()*200/maxHealth))
 		makeTrail(newloc, T, old_direction)
 	//Hearthstone port - track creation hook.


Xaytan:
You could do a check to see if the parent item of the storage is in the player's inventory.

code/modules/mob/living/living.dm
```diff
@@ -950,10 +950,6 @@
 			var/mob/living/L = pulledby
 			L.set_pull_offsets(src, pulledby.grab_state)
 
-//	if(active_storage && !(CanReach(active_storage.parent,view_only = TRUE)))
-	if(active_storage)
```
chaoko99: add an override var for bins rather than deleting this line. This functionality could be used elsewhere.

Fatherless1Xx:
> > It’s kind of baffling to reject quality-of-life changes in the name of balance. If this somehow makes potions too strong, we can just adjust their numbers afterward or add a CD before you can benefit from it again. Balance shouldn’t come at the cost of convenience.
> 
> This is not qol, this is balance. This is a default feature of ss13 code bases that was deliberately disabled for balance.

We can add cooldowns to potions, we can make weapons take up more inventory space, what do you think this could possibly break beyond repair? And even if it is a balance issue (I don't agree it is but let's say it is for the sake of argument) having it auto-close off is objectively worse for quality of life.

ConstantineII:
This is like this so MAA gamers don't go around chugging red at lightspeed while stabbing you. No.

PR #1322, opened 2025-11-02
Adds Basic Purchasable Tools to Merchantry

ophaq:
## About The Pull Request
You can buy from silverface / goldface now:
- Chisel
- Iron scissors
- Hunting Knife
- Tongs

## Testing Evidence
Scissorproof

## Why It's Good For The Game
Found it odd we had other tools there but not these. Scissors seem to be a hot item people want, and I have wanted on several occasions. It's just iron ones, not the fancy steel ones... go to the smith for that! But if you want to skip that long ass smith line for iron ones, here you go.

Let me know if the prices are too low. I made it the same as the hoe's modifier. I added the handsaw and chisel too since it was lacking those.

### DIFF
--- a/code/modules/cargo/packsrogue/merchant/tools.dm
+++ b/code/modules/cargo/packsrogue/merchant/tools.dm
@@ -119,11 +119,6 @@
 	cost = 10
 	contains = list(/obj/item/rogueweapon/huntingknife/scissors,)
 
-/datum/supply_pack/rogue/tools/handsaw
-	name = "Handsaw"
-	cost = 10
-	contains = list(/obj/item/rogueweapon/handsaw = 1,)
-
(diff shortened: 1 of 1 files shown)


SMoto95:
Don't the merchantry already have scissors? Could've sworn I saw it somewhere... We do need chisels sold there for the builders though!

ophaq:
> Don't the merchantry already have scissors? Could've sworn I saw it somewhere... We do need chisels sold there for the builders though!

I can do this. Also, they don't I looked.

benj8560:
they also dont have tongs last I checked. unless I was just being blind

ophaq:
Just realized handsaw was in there already so I removed my addition. Also, added tongs.

PR #1326, opened 2025-11-02
Allows middle aged to be squire, any age to be clerk and shophand

coffeecoffeecoffee:
## About The Pull Request
In the wake of the whole discussion surrounding middle-aged veterans, a friend and I talked about middle-aged squires, and found out there's a couple of other odd age restrictions that feel out-of-place now. This allows any age character to play clerk or shophand (did you know [the oldest clerk was 85 and died in 1902 after working for 57 years](https://www.nytimes.com/1902/12/01/archives/oldest-pension-clerk-dead.html)?) and allows middle-aged characters to squire. I think it'd be too far / unsuitable for _old_ characters to squire, but given we allow unholy creations of the gods like dark elves to do it, having a squire that's past their prime and stuck in their position sounds alright.

## Testing Evidence
ia8z4qXYSo

## Why It's Good For The Game
There are people who prefer to play older characters but do not feel they have the experience or seniority to play central roles in some of the areas they're restricted to (having to play merchant if they want to work in the shop, having to play steward if they want to help the stewardry, etc.) and other roles don't work this way (knight, the blacksmiths, the inn, etc.)

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/squire.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/squire.dm
@@ -10,7 +10,7 @@
 		/datum/species/lamia,
 	)
 	allowed_sexes = list(MALE, FEMALE)
-	allowed_ages = list(AGE_ADULT)
+	allowed_ages = list(AGE_ADULT, AGE_MIDDLEAGED)
 	advclass_cat_rolls = list(CTAG_SQUIRE = 20)
 	job_traits = list(TRAIT_SQUIRE_REPAIR)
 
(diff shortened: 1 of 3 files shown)


coffeecoffeecoffee:
> > having a squire that's past their prime and stuck in their position sounds alright.
> 
> No, we don't want middle aged or old squires as it's meant to be an aspiring young role.

Why? This isn't necessarily how squires operated in real life and the "aspiring" part is already out the window with disfavored races being allowed to squire-for-life. I have met a number of people now playing grizzled old squires.
I did not add "old" to it, and that would probably be too much, yes. Only middle-aged, fyi.

Nik3321:
Brother, squires are literally described as very young and early in training, there is no world where they would train a man on his 40s to be a knight. Literally goes against the role.

PR #1327, opened 2025-11-02
Sentimental Instrumentality: Absolver buffs & Stigmata adventurer subclass

Ephemeralis:
## About The Pull Request

Buckle in. We're *gaming*.

- Adds the PSYDON-locked Stigmata cleric subclass, essentially a mini-Absolver. The Stigmata gets the wound transfer miracle that Absolver has, and a nerfed version of ABSOLVE titled as AMEND that **only** transfers injuries to you from targets alive or dead. It does not and cannot resurrect.
   - Stigmata is a pacifist with dodge expert, crit resistance and a very minor speed boost, as well as lesser bonuses to CON and END compared to Absolver. It also doesn't start with a suit of essentially blacksteel, either. Optimizing your Stigmata for combat survivability will require you to make sacrifices in CON and END due to how statpacks work, which means you'll be more susceptible to bleeding and the like. Choices, choices!
   - Stigmata is not painstun immune like Absolver is and thus managing high levels of wounding is going to require you to use other things to get close to their level of effectiveness.
   - To accommodate for AMEND not being able to rez (like, at ALL), they start with Expert medicine to get them direct access to the lux-revival surgery.
   - Note: as AMEND does not attempt resurrection, it makes Stigmata clerics very capable corpse-fixing helpers.
   - Stigmata start with a very limited selection of gear. This does not include any kind of surgeon's tooling or anything of the sort. You'll need to ENDURE to get the mammons for the things you need.
   - Stigmata do not have any innate inquisitorial affiliation and will need to seek that out like Naledi mercenaries do, if they so choose.
- ABSOLVE's logic sequence now works a little differently on dead mobs. It will now attempt to resurrect a dead mob as it would before, but instead of doing nothing if the person refused the rez (or was unable to accept it), it will transfer all of their wounds to you instead.
   - Functionally, this is a TREMENDOUS increase to the Absolver's capability to peel people off the floor. Healing your wounds/damage is significantly easier than spending 5 minutes in surgery fix wounds hell. This change kind of elevates them to a tiny step below Astratan clerics in their raw ability to prepare people for resurrection, albeit at considerable risk to themselves.
   - NOTE: Absolve will attempt to warn you with an alert prompt if you're going to absolve a corpse with enough oxyloss damage to instantly kill you. To successfully prepare a corpse in this way, get its oxyloss damage down.
- ABSOLVE's resurrection no longer kills you (the caster) on the first use. Instead, it devitalizes you with a debuff that lasts for 10 minutes and functionally eliminates the CON bonus you receive from the class during this timeframe.
   - If you use ABSOLVE again while devitalized in this way, you can attempt a second resurrection at the cost of your lyfe, as before.
   - For all intents and purposes, this is essentially an expedited lux surgery revival without actually requiring lux.
   - This means that the Absolver is now functionally capable of peeling up to 2 people off the floor in the heat of combat every 10 minutes. They'll probably die in the process, though.
- ABSOLVE's resurrection is no longer an admin full-heal (too strong for the frequency it can be done at). You can however, do a rough equivalent by spending two casts of ABSOLVE - one to prepare the corpse (by taking its wounds onto yourself), and another to bring them back to lyfe.
- Absolver and Stigmata both receive the ritualist trait, and a stick of rite chalk at round start. This is so they can do the currently unimplemented PSYDON rite for sovl. Thank the suggester below.

## Testing Evidence

image

image

## Why It's Good For The Game

Absolver is STUPENDOUSLY kino. Possibly the most class of all time. Stigmata lets people get a tiny bit in on that kind of stuff without having to necessarily war with the other absolver mains for the ONE slot.

### DIFF
--- a/modular_scarletreach/code/modules/jobs/job_types/roguetown/adventurer/stigmata.dm
+++ b/modular_scarletreach/code/modules/jobs/job_types/roguetown/adventurer/stigmata.dm
@@ -111,7 +111,7 @@
 	var/oxy_transfer = H.getOxyLoss()
 	var/clone_transfer = H.getCloneLoss()
 
-	if (oxy_transfer >= 180)
+	if (oxy_transfer >= 150)
 		if (alert(user, "THEY ARE ASHEN WITH STILLED BREATH. AMENDMENT MAY INSTANTLY KILL YOU, STIGMATA. PROCEED?", "SELF-PRESERVATION", "YES", "NO") != "YES")
 			revert_cast()
 			return
--- a/code/modules/spells/roguetown/oldgod.dm
+++ b/code/modules/spells/roguetown/oldgod.dm
@@ -399,7 +399,7 @@
 	var/oxy_transfer = H.getOxyLoss()
 	var/clone_transfer = H.getCloneLoss()
 
-	if (oxy_transfer >= 180)
+	if (oxy_transfer >= 150)
 		if (alert(user, "THEY ARE ASHEN WITH STILLED BREATH. ABSOLUTION MAY INSTANTLY KILL YOU, LAMB. PROCEED?", "SELF-PRESERVATION", "YES", "NO") != "YES")
 			revert_cast()
 			return


Sgt-Solid01:
This is good! But one fat issue stands, TIME DILATION, the ten minutes revival timer will likely be much longer when it's high-pop, is this accounted for?

modular_scarletreach/code/modules/jobs/job_types/roguetown/adventurer/stigmata.dm
```diff
@@ -0,0 +1,141 @@
...
+	outfit = /datum/outfit/job/roguetown/adventurer/stigmata
+	class_select_category = CLASS_CAT_CLERIC
+	category_tags = list(CTAG_ADVENTURER, CTAG_COURTAGENT)
+
+	subclass_stats = list(
+		STATKEY_CON = 5,
+		STATKEY_END = 3,
+		STATKEY_SPD = 1,
+		STATKEY_STR = -2,
+		STATKEY_INT = -2,
```
benj8560: why negative int on a casting class...int effects spell cooldowns

benj8560: ALSO a dodging class. Already just begging to be feinted

Ephemeralis:
> could absolver get their unarmed bumped up to expert, if we're talking buffs? unlike the wrestling they don't have, they can't attack at all with unarmed, and novice is utterly dwarfed by their usual attackers when it comes to defending

Half of why Absolver is so kino is that you're forced to huddle behind the rest of the inquisition since you're both super vital to them and hellishly vulnerable on your own. Outside of the fact that you're decked out in darksteel on round start.

If you are surrounded by the enemy alone, you've already lost. I don't really want to change that. I might look at giving them a bit more wrestling in a later PR, though.

modular_scarletreach/code/modules/jobs/job_types/roguetown/adventurer/stigmata.dm
Ephemeralis: int malus was an unintentional holdover from this being based off flagellant, have removed it

moltosotto:
You should also make Absolve scale so that it applies a 

> > could absolver get their unarmed bumped up to expert, if we're talking buffs? unlike the wrestling they don't have, they can't attack at all with unarmed, and novice is utterly dwarfed by their usual attackers when it comes to defending
> 
> Half of why Absolver is so kino is that you're forced to huddle behind the rest of the inquisition since you're both super vital to them and hellishly vulnerable on your own. Outside of the fact that you're decked out in darksteel on round start.
> 
> If you are surrounded by the enemy alone, you've already lost. I don't really want to change that. I might look at giving them a bit more wrestling in a later PR, though.

Wrestling would be based, as you're literally forced to tackle deadites once in a while.

Also, Absolver armor is light armor aside from the crown of throrns and blacksteel thorns (bracers.) The breastplate is especially weaksauce.

benj8560:
totally pointless sovl suggestion. Could we give the Absolver and this new class ritualist and some chalk? (absolver has no spare bag space so that might be a issue...) Just so they can create the (MECHANICALY USELESS) psycross ritual circles. The lack of psydonites with ritualist is sad. I think the only possible class is witch for it currently.

Ephemeralis:
> totally pointless sovl suggestion. Could we give the Absolver and this new class ritualist and some chalk? (absolver has no spare bag space so that might be a issue...) Just so they can create the (MECHANICALY USELESS) psycross ritual circles. The lack of psydonites with ritualist is sad. I think the only possible class is witch for it currently.

I hear you, and it shall be done.

Ephemeralis:
Okay, so, this has cropped up a few times now: **if you absolve or amend a corpse that has maxed-out oxyloss (aka has probably bled out or drowned, which is most cases), you WILL instantly die by taking their wounds.** This will not rez them. It will not rez you. It will just kill you and fix their now-healed corpse.

I have added warnings to allow you to back out of a corpse absolve at 150 oxyloss if you want to. You still may die even below this threshold depending on how your oxyloss damage is.

Please don't adminhelp about this. The spell is working as intended.

moltosotto:
@Ephemeralis 

Can you jak expert medicine to Absolver as well? It's funny that Stigmata gets it but Absolver, the dedicated healer doesn't.

Ephemeralis:
I don't want to jak absolver to expert medicine because they should be rezzing with ABSOLVE, not with lux surgery. Helping out with fellow Psydonite resurrections at the church or on the road is how you should be getting it to expert.

PR #1333, opened 2025-11-03
Adds Weak, Carved Oni & Kitsune Mask to Loadouts + Crafting

ophaq:
## About The Pull Request
See title. It takes apprentice crafting to make, along with a log, cloth and fiber.

## Testing Evidence
maskproof

## Why It's Good For The Game
Just making a weakened version for people who want the aesthetic but don't want the hassle to get the steel variant ingredients, especially because a smith isn't always around. Also, wooden masks are cool! I also changed the descriptions of the original ones to more accurately reflect they're made of steel and therefore way better.


kittysmooch:
these are for ninjas. if you want one you're supposed to play the ninja role.

PR #1334, opened 2025-11-03
Servant/seneschal vomitorium stipend trait

Ephemeralis:
## About The Pull Request

Simple, probably controversial change: servants and butlers now have a trait entirely unique to their role that allows them to pull whatever they like from the vomitorium with the cost taken directly from the treasury funds (if any are available).

A notification in the vomitorium alerts them of this fact, and every withdrawal is also logged to the steward's records.

Yes. this means servants can potentially embezzle large amounts of money from the treasury. One could say this is a new steward gameplay loop! How shocking.

## Testing Evidence

5BceFlVpzi

## Why It's Good For The Game

I wrote a paragraph about this originally but figured nobody will read it. Long story short, embezzlement is funny, and servants being able to do their role without wanting to commit die on bad keep rounds is pretty good.


Xaytan:
Suggestion: Make this only work for buying from local stock. That would dramatically improve the perceived importance of farming and other resource acquisition jobs.

Redrover1760:
Could this be added to the steward as well? And maybe the Duke?

maker-of-nightmares:
This is peak I love it :D

awawaawawaawa:
its peak sire

Cleric2145:
> Could this be added to the steward as well? And maybe the Duke?

IMO I think it's best left for the servants. It's their whole job, and it's a huge boon for them to have. Giving it to more people will just negate the benefit. Maybe Duke, since it's literally his realm and money, but he really shouldn't be faffing about with food tbh. Servants should be a major boom to have instead of just being the people with higher cooking skill (which takes so little time to grind with no real consequence unlike with medicine or combat.)

EvenInDeathIStillServe:
There's no reason for money to be taken from the treasury funds, though? The stockpile is owned by the duchy.

Ephemeralis:
> There's no reason for money to be taken from the treasury funds, though? The stockpile is owned by the duchy.
> 
> And maybe have a limit on the max value of an item a servant can withdraw. There's no reason for them to be able to pull out gold from the vomitarium.

You know, that's actually a good point. It sort of doesn't need to take money from the treasury, does it?

PR #1346, opened 2025-11-04
Surgery PR 2: Speeeeeeeed boooooooooost

Xaytan:
## About The Pull Request

Adjusts surgery logic and numbers in response to player feedback. Fixes a bug where surgery speed was WORSE the more favorable the conditions.

- Removes 20% increased failure chance when performing surgery without anasthesia. Instead, anasthesia increases surgery speed by 20%

- Rebalances surgery times based on conditions. Beds and operating tables are the fastest, standing is the slowest.

- Adjusted situational success probability modifiers.

- Corrected an issue where improving surgery conditions caused surgery to become slower, rather than faster.

## Testing Evidence

image

## Why It's Good For The Game

Rounds off some rough edges from the last surgery PR. Makes things behave closer to how they're intended.

### DIFF
--- a/code/modules/surgery/_surgery_step.dm
+++ b/code/modules/surgery/_surgery_step.dm
@@ -382,8 +382,11 @@
 		var/implement_type = tool_check(user, tool)
 		if(implement_type)
 			speed_mod *= implements_speed[implement_type] || 1
-	speed_mod *= get_location_modifier(target, user)
-
+	speed_mod *= get_location_speed_modifier(target, user)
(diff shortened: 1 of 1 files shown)


Ephemeralis:
One thing from the previous PR: the inquisition keeps encountering situations where to sign their arrival slips, they need to perform an incision on their hand. We've had no less than **5** ahelps about people nearly killing themselves off major infections doing this.

Self-incisions on the hand probably need to either always succeed or not immediately cause major infections.

PR #1351, opened 2025-11-04
single line change i made on my phone that buffs fireball

chemistrymain3:
## About The Pull Request
it now sets you on a tiny bit of fire
## Testing Evidence

i didnt test it its on my phone
## Why It's Good For The Game

fireball should probably set you on fire a little


chemistrymain3:
> I thought it already did set you on fire I specifically remember being set on fire by the infernal watcher meatball guy's fireball

It is more likely with gfb because it has 3 tiles and walking through those fire tiles sets you aflame but sometimes you get hit by the fireball and walk OFF the flaming tile at the same time and don’t ignite

so now it sets you very minor on fire (1)

PR #1364, opened 2025-11-06
Three New Druid Forms! || Fox, Bear, Cabbit

ophaq:
## About The Pull Request
Adds the following:
- Vernard (fox) druid form! A mix between cat and volf... roguish!
- Direbear form! A more tankier, slow melee form in comparison to the volf.
- Cabbit form! Be cute, cuddly, and slightly worse than the cat! At least you are a tad faster with some luck on your side.
- Adjusts spider form's END from 6 to 11, STR from 12 to 10.

## Testing Evidence
- Removed reach native, ignore that in the pic.
Vernardproof
Cabbitproof
Direbearproof

To battle my steed!
direride

## Why It's Good For The Game
More forms to turn into is cool, and I have been wanting to have a fox form for a long time. Also, bear form is a druid staple and I'm surprised we didn't have it. Cabbit is in there because they're cute and people like them.

Volf Form:
I made it a much weaker version of the volf form while focusing on speed. It's your rogue form, but not like a cat. You get DE and 16 speed, less con than the volf but more than the cat, and weaker durability on your hide armor. You also get less STR and less wrestling/unarmed so your attacks don't do that much damage. 

Tried fighting a few skeletons, and tested things out. Felt good, not overpowered. Still got hit quite a few times and DE drains stamina fast so it didn't feel overpowered. 

Bear Form:
Very slow, about as strong as a volf in terms of claw damage and hide armor. However, you can grapple if you're smart enough with more strength in comparison. Plus, you get swimming so consider this your swimming form as well if you're not being a saiga.

Also, yes- you can ride this but it might look a bit janky. I am considering removing it.

Cabbit Form:
It's the cat form, but with more luck, 20 speed and less skills. Is that too much? Perhaps, the cat has 18 but you also are worse in terms of skills and this form is just to get you around and probably fragged.

Spider Form:
I was told that due to recent climbing updates, the -climbing- form doesn't have enough endurance to do its job properly. Thus, I bumped its END up, and lowered its STR in exchange since beespiders are kinda weak.

Should be ready to TM. Let me know what needs changing.

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/wildshape/bear.dm
+++ b/code/modules/mob/living/carbon/human/species_types/wildshape/bear.dm
@@ -19,7 +19,7 @@
 		src.adjust_skillrank(/datum/skill/misc/athletics, 3, TRUE)
 
 		src.STASTR = 14 //Might be too high, but then again you're a bear, and you gotta wrestle
-		src.STACON = 14 // One higher than saiga
+		src.STACON = 12 // One less than saiga
 		src.STAEND = 12
 		src.STAPER = 10
 		src.STASPD = 6 // You are a hulking mass of muscle, and this is for balance reasons


ophaq:
> Druids already don't trigger ambushes, they have ambushable set to false like bandits do.

Really? Okay, could have sworn I triggered them before. I'll take out reach native then. I am also trying to figure out how to let cat form wear bell collars but that might be another PR.

ChickyRooster:
Looks fun! Really hoping we get minor werevolf or something for druids one day.

Murderfoxen:
Woooooooooh!!

Ephemeralis:
I've un-testmerged this after watching a direbear survive over a minute of diamond formation from the garrison. We're not doing that, no way.

zeratino:
druids try not to immediately abuse the tools given to them challenge: impossible

A-Distinguished-Axolotl:
Spider probably doesn't need the str nerf but that's just my take

PR #1368, opened 2025-11-06
Collaring prisoners/MASK OFF

Sniperjolly:
## About The Pull Request

Changes Castifico cursed mask to a cursed collar. It's distinct from the normal cursed collar in game files to not fuck with that item. In addition to being pacified and being magiclocked it gives -2 strength and -2 endurance. This also adds an on-examine effect saying that they're collared to make it slightly more obvious. 

## Testing Evidence

Screenshot 2025-11-06 104830
Effects trigger, text is correct, sprite shows up on model (though not in inventory for some reason)
Screenshot 2025-11-06 104853
Removes properly. Text is correct here as well.

## Why It's Good For The Game

People hate the mask so much OOC they completely disengage with the entire system and hold OOC grudges against players who DO engage with it. It's 100% just because it ruins their drip. I think they're crazy but I'm not attached to the mask either.

One thing I wish I could fix is that the item icon in the equipped items panel is blank instead of the fancy sprite the actual cursed collar has but I've no idea how to do that.

I'm not actually sure about the strength and endurance debuff, since it makes it harder to use them as slave labor. I do like the idea of it debuffing but I'm not sure what the debuff should be.


Cleric2145:
Peak

Sniperjolly:
I wanted to avoid speed because I didn’t want to slow the process of taking them to court. I guess if that’s the issue you can just grab them. 

When I’m at my computer I’ll look at the code (maybe the scruffing code?) to see if making grabs succeed automatically is possible. Being pacified will make it pretty easy to deal with them regardless. I guess the current -2 strength helps there. 

Yes the collar is leashable.

ophaq:
Finally the masked bandits can blow me and be leashed.

coffeecoffeecoffee:
Finally, I can be leashed when I blow up the retinue.

KiteAE:
Havent heard anything since my last comment, will close for now. Feel free to re-open in the future.

PR #1383, opened 2025-11-08
Permits foreigners to use the Meister with extra tax

JustSomeGalSS13:
## About The Pull Request

Comments out the code making foreigners unable to use the Meister. Now they can, but if they aren't a noble or part of the Inquisition they get taxed at 150% the rate of normal people (e.g. if the tax rate is 20% they are charged 30%)

## Testing Evidence

See later comment for test of current code.

## Why It's Good For The Game

Prohibiting foreigners from using the Meister doesn't really do anything positive for the game. It locks them out of being able to fill the stockpile for money, requires them to find a place to stash larger amounts of cash (walking to and from your money box is not good gameplay), and is generally a nuisance. Furthermore, it doesn't make much sense in-character, as the Duke can only collect taxes when people use the Meister. This means that foreigners just don't get taxed, any tax-enforcement laws can't really apply to them (can't demand people put their money in the Meister if they're literally incapable). This change makes it so that they can use it, but there's still a penalty to being foreign in the form of an extra tax, currently at 5%.

### DIFF
--- a/code/controllers/subsystem/rogue/treasury.dm
+++ b/code/controllers/subsystem/rogue/treasury.dm
@@ -1,7 +1,7 @@
 #define RURAL_TAX 50 // Free money. A small safety pool for lowpop mostly
 #define TREASURY_TICK_AMOUNT 6 MINUTES
 #define EXPORT_ANNOUNCE_THRESHOLD 100
-#define FOREIGNER_TAX_SURCHARGE 0.05 //Amount extra that non-noble foreigners are charged as tax
+#define FOREIGNER_TAX_MULTIPLIER 1.5 //Amount that the tax rate is multiplied by for foreigners
 
 /proc/send_ooc_note(msg, name, job)
 	var/list/names_to = list()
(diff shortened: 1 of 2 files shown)


DarknessWithin994:
> Yes please, I had been saying for the longest time we need this back but just at a higher tax rate or something. Though I might suggest the +5% be variable and possibly able to set by the duke and/or steward?

More code work than I wanted to do (I'd have to fiddle with the throne or nervemaster or something), but not a bad idea if someone else wants to code it after.

Teragenic:
You may want to add a tax to actually like... Putting stuff into the stockpile? I.E. Treasures and more.

DarknessWithin994:
> You may want to add a tax to actually like... Putting stuff into the stockpile? I.E. Treasures and more.

That's a good idea but out of scope for this PR. I might make something for it in the future, since it would also encourage direct sales (which involve RP) rather than just dumping everything into the stockpile mindlessly.

DarknessWithin994:
Changed the extra to "50% of the general tax rate" so that if the tax rate is 20% foreigners get 30%, if it's 30% foreigners get 45% etc. (I did update the text on the Meister to reflect the exact amount after the screenshot was taken)
image

PR #1386, opened 2025-11-08
Massive overhaul of combat, armor, health, etc.

FrownTown:
## About The Pull Request

# Armor and Combat System Rework

### Peel 

- Peel has been completely removed from the game, as it is no longer needed. All references to it have been ripped out of the codebase, including comments, so no one will foolishly try to put it back in the game again. Many weapons that originally had peel, such as rapiers, can now be used to stab through the gaps in armor instead.

- Many weapons that had peel intent but lacked a strike intent have had a strike added in its place. A couple axes were given bash. This gives almost everyone an alternative blunt damage option when needed.

### Armor Penetration System

- Armor penetration now scales with character stats based on weapon type:
  - Heavy weapons (WBALANCE_HEAVY): +5 AP per STR point above 10
  - Normal weapons (WBALANCE_NORMAL): Average of STR and PER contributions
  - Swift weapons (WBALANCE_SWIFT): +3 AP per PER point above 10
  - Punches, kicks, and stomps now grant +5 STR-based AP per point above 10
  - Blunt Weapons: Base penetration now varies between armor types (-25, +20, +35) and +50 against metal helmets.

### Precision Strike System

- Swift weapons (daggers, rapiers, etc.) using stabbing attacks with aimed intent can bypass heavy armor on vulnerable zones (arms, legs, neck, eyes, groin). Success is based on weapon skill, attacker PER, defender SPD/CON, and AP vs. armor rating. Underlayers of armor are still checked, so the damage can still be blocked (e.g., by a chain shirt under plate armor).

- High skill and perception is necessary to stab through gaps in armor using a sword in a standing fight. It is easier with shorter weapons, and MUCH easier while grappling or being grappled by the opponent.

### Armor Damage Conversion

- Edged attacks (cut/chop/stab/pierce/pick) convert to blunt damage when dealing less than 15/20/25 damage through armor (depending on armor class). Blunted attacks deal 50% of their already-reduced damage

### Armor Refactor

- Armor has been heavily refactored. It will last far longer, but will often reduce damage significantly than stop it entirely now.

- "Armor softens the blow" will show in chat when armor reduces but doesn't stop damage; it will also show "the attack was blunted by armor" when edged attacks are converted to blunt ones. The amount of damage below which a blow will be blunted varies based on armor class.

- Completely blocked attacks drain stamina now. The amount scales with the amount of damage actually being absorbed by the armor, and with the armor class. Light armor drains far less than heavy armor.

- Armor integrity is now far more consistent, and will tend to last much longer unless sundered deliberately with specific attacks. Destroy soft armor by cutting and hacking it, and destroy heavy armor by smashing it. Destroy medium armor by picking and stabbing at it.

- Armor damage absorption decreases as armor becomes damaged. The maximum decrease scales with the armor class (60% reduction for light armor, 50% for medium, 40% for heavy).

- Armor absorption dramatically reduces dismemberment chance using an exponential formula.

- Dodging and parrying now cost more stamina in heavy armor and less stamina in light armor.

- Wearing heavy armor now imposes a penalty to dodge rolls (reduced significantly by heavy armor training), while wearing light armor imposes a bonus (increased slightly by expert dodger, but capped based on the existing bonus. If you have more than about 12/13 speed, this won't make you dodge any better than before).

- When an attack's damage is stopped entirely by armor, the wearer loses a small amount of stamina based on the amount of damage absorbed.

### Blunt Weapon Knockback

- Blunt weapons can knock targets back when hitting hard enough in Smash Mode, especially if you're a Big Guy.

### Combat Maneuver: Tackle

- Sprint-jump into someone with grab intent in combat mode to tackle them down to the ground, causing both of you to drop what you're holding and initiating a grapple automatically. This is easiest to do from behind, much harder from the front, and is based on STR/CON/wrestling skill and armor weight. If you fail a tackle, you get immobilized for a bit, fall down and lose stamina, so this is dangerous to attempt in the middle of a fight, especially with multiple opponents. It is best used for ambushes and chases.

### Defense Modifiers

- Previously, attacking someone with a swift weapon had a bonus to hitting them while dodging, and a heavy weapon had a penalty. Now, this is only the case when using those weapons on swift or strong intents, respectively. This helps to make dodging more consistent.

- Parrying heavier weapons with lighter ones now consumes more stamina, and parrying lighter ones with heavy weapons consumes less. The existing stamina drain increase from the attacker's strength when using a heavy weapon now applies to all parries, but is lower with weapons lighter than heavy weapons.

- Running out of stamina when parrying has a chance to cause your weapon to be flung out of your hands. This chance is 0% if you have the same strength as your opponent and are wielding equally-sized weapons. It only happens if someone has an advantage, and the chance it occurs is low unless the advantage is significant.

- The maximum chance to dodge or parry an attack has been raised from 90% to 95%, as being wounded in combat now has a higher chance to be crippling or even fatal.

### Critical Hits

- Removed the "90% damage" protection from critical hits and replaced it with a probability calculation based on the ratio of current to max damage on the limb, modified by CON and LUC. Additionally, certain types of critical hits will also have a chance to cause rapid or instant death, so you don't have to cut someone's head off or bleed them like a pig to kill them. Again: this is a *chance* of death, not a guarantee, and dying from a critical hit is less likely than the critical hit occurring to begin with.

- Removed the random +10 damage to wounds from LUC, as this would cause wounds and HP to 
desync, which is a problem I aim to fix. Instead, the attacker's LUC has a chance to 
drastically increase the chance of a critical hit *being allowed to roll*. I want to reiterate that this is the chance that the critical hit *is allowed to roll*, not whether it actually occurs. You can think of it like confirming a critical hit in D&D 3.5e

- There are a number of new critical hit effects that can cause crippling injuries and occasionally rapid death, but only if the target already has a large amount of damage on the limb or the attack deals an absurdly high amount of damage in a single blow. These effects are all survivable, but may occasionally kill you. This is mitigated heavily by CON and critical hit resistance. Using strong intent will make smashing skulls and such more likely, while aimed intent will make skewering hearts or livers etc. more likely.

- These new, potentially fatal critical hit effects have a minimum damage requirement before they are allowed to occur: the limb must have reached 75% of its maximum brute damage.

- Critical hit prevention from armor is still in place, although I am considering tweaking this to remove the crit prevention once the armor is sufficiently damaged (or perhaps destroyed). It depends on how things play out.

- Bear in mind that death from critical hits is generally the realm of piercing and blunt weapons, as slashing weapons can already instantly kill you by cutting your head off, and will make you bleed out fastest.

### Combat Triangle

- Combat now follows a rock-paper-scissors formula: blunt weapons defeat heavy armor, cutting weapons defeat light armor, and piercing weapons defeat medium armor. Overall, heavy armor is still much more protective, but has a few drawbacks to make up for it (detailed above). Conversely, light armor is still much less protective, but has a few advantages.

### Wound damage

- Calculation for wound damage was incorrect before, and resulted in most wounds being worse than they should have been, as well as the inability for armor to actually reduce damage. This has been fixed, and wound damage should now properly sync up with the mob's overall brute damage again. Wounds in general have also been rebalanced to allow for less severe wounds to be more common by scaling their blood loss and pain effects with damage from a lower base number.

- Removed armor capping the size of wounds, as this was no longer necessary, doesn't make sense, and encouraged cutting every limb once or twice, rather than focusing on dealing more damage to a single limb.

### Survival

- Bleed rate now slows dramatically as a character runs out of blood, to simulate loss of blood pressure and extend pre-death time. The soft crit threshold now scales with CON, and the rate of oxygen loss while unconscious from blood loss also scales with CON. In general, bleeding to death will be much slower, and you will remain unconscious peacefully for a much longer time before you actually pass away.

- Rebalanced unconsciousness threshold. In general, you will be unconscious for between 90-130 seconds before death with "moderate" injuries (this is as opposed to about 15-20 seconds before)

- Falling unconscious from blood loss will happen in roughly the same amount of time as before with 10 constitution, but will take more or less time based on your stats now.

- Bleed rate from large, devastating wounds will generally be greater now due to damage scaling, while bleed rate from multiple minor wounds will generally be lesser due to a lower base amount.

### Pain Value Adjustments

- Pain is no longer based solely on wound type, but scales with actual damage inflicted, from a lower base number. In addition, there has been significant rebalancing of pain values for different wound types:

	- Puncture Wounds and Gouge Wounds now cause roughly 3x more pain
	- Bruises cause roughly 40% as much pain as before, which is still higher than anything else besides being whipped

- The end result is that punctures now properly hurt and bruises are no longer near-instantly incapacitating.

### Organ Damage

- Taking damage to organs from critical hits does not instantly kill you most of the time, but it may kill you over time or just ruin your week unless you get it treated. In the spirit of fun, I've reinstated the brain trauma system since brains can take damage again. However, I stripped out all but a few of the traumas from the allowed list to keep things "gritty" and also "not extremely annoying and/or silly/immersion breaking."

- For the time being, both organ damage and brain traumas can be cured with wound miracle or healing surgery applied to the external limb the organ belongs to. They cannot, however, be cured by potions.

### Balance Changes

- Heavy armor blunt protection increased, because blunt damage now has 35 AP against it by default. This isn't enough to defeat armor unless you have around 13 strength, or are hitting with strong or smash intent. The point of this change is to make blunt weapons more likely to cause damage through heavy armor without increasing their raw damage amount so much that they destroy it instantly or kill unarmored people in a small number of hits. Blunt attacks will get +50 AP against hard helmets.

- Leather armors had blunt damage reduction reduced because blunt attacks now have -25 AP against it by default. They will tend to reduce damage far more often than they stop it completely now. These changes were intended to allow blunt weapons to remain effective against heavy armor without making them instantly lethal against unarmored opponents (higher damage) or making them easily penetrate other types of armor (higher AP). 

- Blunt damage has +20 AP against chain coifs and the like by default. It is a relatively low amount, but necessary to prevent chain coifs from stopping warhammers when a visored sallet won't.

- Studded leather had a slight reduction in slashing protection (still in line with hardened leather) in exchange for keeping high blunt protection and will stop many blunt attacks entirely.

### Additional Stuff

- Added Knight and Man-at-Arms NPCs for testing purposes (these don't spawn anywhere, but admins could spawn them). They are hostile NPCs that do not have infinite stamina, and which spawn with role-appropriate stats and equipment.

- Made the beast claws actually turn punches into claw attacks while worn, but only if the wearer has expert pugilist. Also improved their description.

- Fixed some weirdly chosen sound effects

- Fixed misuse of the "adjustStaminaLoss" proc in roguetown code, as it is vestigial and doesn't actually do anything. Replaced instances that looked like they were part of code that's actually executed with stamina_add. Everyone should be using that in the future.

## Testing Evidence

Tested on local vs. a variety of mobs and NPCs and it feels pretty good to me, but it's impossible to test changes this broad without a lot more live data, so... Here we are lads.

## Why it's good for the game

Fixes many longstanding problems. Keeps balance roughly where it is, but closes the distance between different roles by making variable tactics more effective. Fights will tend to last for a shorter duration and be more dangerous and unpredictable, but good armor and stats will still keep you on top... mostly.

A test merge is necessary to get any real data on how this turned out, but I think people will be pleased.


CaydenT:
This all sounds kind of awesome! I am concerned about tackle but if its as risky as it sounds it should _hopefully_ not be too bad. Does defense intent stop it like charging already does?

KiteAE:
Expect a lot of "fix conflicts" in the coming week or two.

code/modules/mob/living/blood.dm
```diff
@@ -162,9 +162,26 @@
...
+			// CON-based soft crit threshold: higher CON = survive at lower blood levels
+			// Base is 122 at 10 CON, -10 blood per CON point above 10
+			// This means high CON characters enter soft crit later (at lower blood volume)
+			var/soft_crit_threshold = BLOOD_VOLUME_SURVIVE - ((STACON - 10) * 10)
+			soft_crit_threshold = clamp(soft_crit_threshold, 40, 200)
+
+			// Progressive oxygen damage based on blood loss severity
+			// Much lower rates for extended bleedout time (120-160 seconds)
+			if(blood_volume <= 40) // Death threshold - ~7% blood
+				adjustOxyLoss(2) // Rapid death
```
Ephemeralis: pre-build the oxyloss value amount in a distinct variable and apply it with 1 ``adjustOxyLoss`` call where possible to minimize health updates, big contributors to lag

code/modules/mob/living/blood.dm
```diff
@@ -194,6 +211,17 @@
...
+
+	// Blood pressure system: less blood = lower blood pressure = slower bleeding
+	// Uses exponential curve for more dramatic slowdown near death
+	if(bleed_rate > 0 && blood_volume < BLOOD_VOLUME_NORMAL)
+		var/blood_pressure = blood_volume / BLOOD_VOLUME_NORMAL
+		// Square the pressure ratio for exponential slowdown
+		// At 50% blood: 0.5^2 = 25% bleed rate
+		// At 25% blood: 0.25^2 = 6.25% bleed rate
+		// At 10% blood: 0.1^2 = 1% bleed rate
+		bleed_rate *= (blood_pressure ** 2)
```
Ephemeralis: food for thought: since this is so much dramatically slower to hit 0 blood volume now, we can possibly consider **massively** increasing the amount of oxyloss damage dealt at this point to expedite things along, or even call death() directly?

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -55,10 +146,44 @@
...
+		// Armor degradation multipliers based on damage type vs armor class
+		var/degradation_mult = 1.0
+		if(istype(used, /obj/item/clothing))
+			var/obj/item/clothing/armor_piece = used
+			switch(blade_dulling)
+				if(BCLASS_BLUNT, BCLASS_SMASH)
+					// Blunt damage: lowest vs light, highest vs heavy
+					switch(armor_piece.armor_class)
+						if(ARMOR_CLASS_LIGHT)
+							degradation_mult = 0.15
```
Ephemeralis: enormous nitpick but these values may be best to have in defines, potentially?

Ephemeralis:
Honestly - this looks very good on paper and well thought out. It does sort of make dodge even more amazing than it was before, though. Tackle is going to be *very* funny with Xylixians.

The blood pressure change is simple genius. Great stuff.

code/modules/mob/living/blood.dm
FrownTown: I thought about making it so that reaching 0 blood volume just makes you die, but wasn't sure. That is probably fine, I think, but since killing strokes are now possible without having to behead somebody, I wanted to see how it would play it first.

FrownTown:
> This all sounds kind of awesome! I am concerned about tackle but if its as risky as it sounds it should _hopefully_ not be too bad. Does defense intent stop it like charging already does?

Good idea. It does now.

RyszardParys:
This just sounds like heavy armor users will run out of stamina really quickly with how many "user loses stamina" i've read in this PR, let's not forget they're also swinging their weapon in a fight.
Meanwhile light armor dodge masters will keep being the meta, since now you can cause damage even through heavy armor anyway. 😕

WeNeedMorePhoron:
This is not directly related to your PR.

Yesterday I ported over Ephemeralis's optimization to AP and got a bug report everyone's wounds were clotting quickly. I saw that in `code/modules/mob/living/life.dm` after Line 38 that Ephemeralis optimized some life code in Russian that gave everyone heal_wounds(1) and set it to heal_wounds(1) or (1.6).

When I checked the git blame it looked like during a runtime fix it was mistabbed and the **passive healing meant for simple animal were applied to everyone**.

tl;dr Your codebase unknowingly gave everyone Psydonite level of healing equal to having a miracle (without healing brute loss / fire loss) active at all time on everyone, which was originally only meant for simple animals.

I'd suggest fixing, removing, or taking into account this unintended feature / bug before any major tweak to the system. I genuinely did not understand why you guys thought omniwound were insufficiently deadly and doubled bleed rate on slash and puncture until I found out about this bug yesterday.

Code_H7hRP0WizM

Ephemeralis:
> This is not directly related to your PR.
> 
> Yesterday I ported over Ephemeralis's optimization to AP and got a bug report everyone's wounds were clotting quickly. I saw that in `code/modules/mob/living/life.dm` after Line 38 that Ephemeralis optimized some life code in Russian that gave everyone heal_wounds(1) and set it to heal_wounds(1) or (1.6).
> 
> When I checked the git blame it looked like during a runtime fix it was mistabbed and the **passive healing meant for simple animal were applied to everyone**.
> 
> tl;dr Your codebase unknowingly gave everyone Psydonite level of healing equal to having a miracle (without healing brute loss / fire loss) active at all time on everyone, which was originally only meant for simple animals.
> 

Holy smokes.

code/modules/mob/living/blood.dm
FrownTown: Decided to implement this after further testing. Update incoming.

FrownTown:
For the people complaining about dodging, this was why the tackling mechanic was added, and why wearing heavy armor gives you a bonus to the tackle roll. A tackle cannot be roll-dodged, only resisted or guarded (if you have good reflexes IRL).

You can also switch to a smaller weapon and use swift intent to reduce their chances of dodging.

Furthermore, heavy armor can be defeated more easily than before, but the difference in survivability is also larger than before. In testing, a fight between a man at arms in heavy armor and any mercenary class with "expert dodger" always ended in victory for the man at arms for a few reasons.

1) Heavy armor stops damage completely far more often than light armor
2) Heavy armor prevents almost all types of critical hits from occurring until the armor is completely destroyed

This meant that it only took a few square hits from a spear or similar weapon to cause crippling damage against the person wearing lighter armor, while the man at arms would take mostly minor wounds that were easily survivable.

Raising the cap on parrying and dodging to 95% only slightly mitigated this disparity. I highly encourage downloading the branch and testing it for yourself if you're worried about it.

For antagonists, fighting one of the keep's knights is still going to feel like a boss battle.



> I don't think parrying in heavy armor should take extra stamina. Heavy armor is already starting to sound like noob bait with both 95% dodge/parry chances and running out of stamina super fast, blocking damage fully resulting in stamina damage, getting kicked over, and immediately getting fully bypassed, compared to light armor and medium armor that can just out stamina people. Irl wise heavy armor isn't that stamina draining anyways.
> 
> Also dodging is already rather strong. If someone can roll 95% dodge checks against you due to speed maxing you basically cant do much outside of stamina draining them, charging, and now I guess switching weapons.

Heard regarding parrying draining differing amounts of stamina based on armor worn. I've removed that feature.



> This is not directly related to your PR.
> 
> Yesterday I ported over Ephemeralis's optimization to AP and got a bug report everyone's wounds were clotting quickly. I saw that in `code/modules/mob/living/life.dm` after Line 38 that Ephemeralis optimized some life code in Russian that gave everyone heal_wounds(1) and set it to heal_wounds(1) or (1.6).
> 
> When I checked the git blame it looked like during a runtime fix it was mistabbed and the **passive healing meant for simple animal were applied to everyone**.
> 
> tl;dr Your codebase unknowingly gave everyone Psydonite level of healing equal to having a miracle (without healing brute loss / fire loss) active at all time on everyone, which was originally only meant for simple animals.
> 
> I'd suggest fixing, removing, or taking into account this unintended feature / bug before any major tweak to the system. I genuinely did not understand why you guys thought omniwound were insufficiently deadly and doubled bleed rate on slash and puncture until I found out about this bug yesterday.
> 
> Code_H7hRP0WizM

Fixed.

FrownTown:
Update: 

- Applied the degradation multipliers for heavy armor to bucklers and metal shields to make them last longer unless deliberately targeted with blunt weapons, which will smash them apart. Wooden shields, specifically, will also take extra damage from CHOP intent.
- Shields now use less stamina when blocking than parrying with a weapon does, since they prevent you from using weapons in two hands, which is a tremendous decrease in damage.
- Attacking through the gaps in armor is now possible with a few, select weapons that have a balance of 0 (longswords, estocs, greatswords and zweihanders) which were historically gap-finders. These weapons must be wielded in two-hands while attacking for this to work, do not get a bonus to the attempt while grappling/being grappled, and receive a slight penalty to the success chance.

kiwedespars:
> For the people complaining about dodging, this was why the tackling mechanic was added, and why wearing heavy armor gives you a bonus to the tackle roll. A tackle cannot be roll-dodged, only resisted or guarded (if you have good reflexes IRL).

the solution to overtuned dodge is not tackling, even if the chances on landing it are really good.

- miss and hit a wall? you're stunned and paralyzed for 2 seconds and knocked down.
- miss? you're now stuck in place and off-balanced, letting them jump into you or kick you to knock you down

this also makes fighting multiple people even more hopeless because if your solution to fighting one whose a dodge expert is to tackle them and put both of you on the ground, you're basically dead

> You can also switch to a smaller weapon and use swift intent to reduce their chances of dodging.

this also isn't a solution currently and in the pr because using swift weighted weapons w/ swift intent only reduces the currently absurdly high chance to dodge based on your speed vs their speed unless i'm mistaken.

image

outspeed them and your reward is reducing their dodge chance by 10 per level of speed difference

if you don't outspeed them you're just giving them extra dodge chance

FrownTown:
> > For the people complaining about dodging, this was why the tackling mechanic was added, and why wearing heavy armor gives you a bonus to the tackle roll. A tackle cannot be roll-dodged, only resisted or guarded (if you have good reflexes IRL).
> 
> the solution to overtuned dodge is not tackling, even if the chances on landing it are really good.
> 
> * miss and hit a wall? you're stunned and paralyzed for 2 seconds and knocked down.
> * miss? you're now stuck in place and off-balanced, letting them jump into you or kick you to knock you down
> 
> this also makes fighting multiple people even more hopeless because if your solution to fighting one whose a dodge expert is to tackle them and put both of you on the ground, you're basically dead
> 
> > You can also switch to a smaller weapon and use swift intent to reduce their chances of dodging.
> 
> this also isn't a solution currently and in the pr because using swift weighted weapons w/ swift intent only reduces the currently absurdly high chance to dodge based on your speed vs their speed unless i'm mistaken.
> 
> image
> outspeed them and your reward is reducing their dodge chance by 10 per level of speed difference
> 
> if you don't outspeed them you're just giving them extra dodge chance

Go download the branch and try it for yourself. If you experience something like you're worried about, screenshot it and post it here. Until you do that, I'd appreciate if you could stop whining in the comments on my PR, please.

kiwedespars:
> It does not apply anything if you are not faster than they are. They do not get "extra dodge chance."

thank you for clarifying, since i was glancing over it prior and didn't pick that part up, but i still don't see it as a solution to dodge because the values are so low. -10 per speed diff is slightly helpful but doesn't amount to much since you need to fullsend into speed to make use of it

>Almost certainly, any instance where a player has maximum dodge chance against another combatant of similar skill is a result of the other player using a heavy weapon

this is where the misunderstanding is, since max dodge is easy to hit due to dodge expert. you can reach maximum dodge chance with minimal investment since it's speed * 15 vs their speed * 10 

10 speed (DE) vs 10 speed (DE or not) turns into 150 - 100. dodge baseline starts at 50%, then adds on the difference, so that'd 100% dodge chance, and you'd have to outspeed them by 2 just to make them not hit the current cap of 90.

having dodge expert as the attacker doesn't do anything and there's no way to influence this outside of lowering their speed somehow. it will always be your speed * 10, so if they invest 2 points, it's 130% dodge chance, and you need 5 points invested into speed to make sure they don't sit at the cap

> These changes do not make dodging "more overturned" - they explicitly make it worse by halving the bonus from speed in many cases.

i wasn't stating the changes to intents were what made it overtuned, i was more going on about how

> dodge is looking really overtuned with dodging stamcost going from 10 -> 6.5 and cap being raised from 90 to 95

if i didn't come across that way, that's probably my bad

FrownTown:
The dodge baseline starting at 50% 

> > It does not apply anything if you are not faster than they are. They do not get "extra dodge chance."
> 
> thank you for clarifying, since i was glancing over it prior and didn't pick that part up, but i still don't see it as a solution to dodge because the values are so low. -10 per speed diff is slightly helpful but doesn't amount to much since you need to fullsend into speed to make use of it
> 
> > Almost certainly, any instance where a player has maximum dodge chance against another combatant of similar skill is a result of the other player using a heavy weapon
> 
> this is where the misunderstanding is, since max dodge is easy to hit due to dodge expert. you can reach maximum dodge chance with minimal investment since it's speed * 15 vs their speed * 10
> 
> 10 speed (DE) vs 10 speed (DE or not) turns into 150 - 100. dodge baseline starts at 50%, then adds on the difference, so that'd 100% dodge chance, and you'd have to outspeed them by 2 just to make them not hit the current cap of 90.
> 
> having dodge expert as the attacker doesn't do anything and there's no way to influence this outside of lowering their speed somehow. it will always be your speed * 10, so if they invest 2 points, it's 130% dodge chance, and you need 5 points invested into speed to make sure they don't sit at the cap
> 
> > These changes do not make dodging "more overturned" - they explicitly make it worse by halving the bonus from speed in many cases.
> 
> i wasn't stating the changes to intents were what made it overtuned, i was more going on about how
> 
> > dodge is looking really overtuned with dodging stamcost going from 10 -> 6.5 and cap being raised from 90 to 95
> 
> if i didn't come across that way, that's probably my bad

OK. I hear you. Here's what I did.

Wearing light armor now gives anyone a +5 dodge bonus.

Dodge experts get a bonus of prob2defend += max(5, 20 - ((H.STASPD - 10) * 5)). This means they get a bonus of up to 20%, which diminishes by 5% for each point of speed they have above 10.

At the same time, I reduced the multiplier from dodge expert from x15 to x13.

The result is:

Defender Speed | Light Armor Bonus | Base Calc | Total | Attacker Speed Needed | Difference
---------------|-------------------|-----------|-------|----------------------|------------
10                  | +20                    | 130         | 200   | 11 (prevent 95%)     | +1
11                  | +15                    | 143         | 208   | 12                            | +1
12                  | +10                    | 156         | 216   | 13                            | +1
13                  | +5                      | 169         | 224   | 13                            | 0
14                  | +5                      | 182         | 237   | 15                            | +1
15                  | +5                      | 195         | 250   | 16                            | +1
16                  | +5                      | 208         | 263   | 17                            | +1
17                  | +5                      | 221         | 276   | 19                            | +2
18                  | +5                      | 234         | 289   | 20                            | +2
19                  | +5                      | 247         | 302   | 21                            | +2
20                  | +5                      | 260         | 315   | 22                            | +2


At 13 speed there's a weird spot where the defender will have less than the cap dodge rate against opponents with the same stats.

From 10-16, other than 13, it's a linear increase, basically a 1:1 ratio where the attacker only needs 1 more speed to reduce the dodge chance below the cap.

At 17 or above speed, the ratio becomes 2:1. This is still far lower than you've identified.

This should solve the problem without making dodge expert useless at lower stat levels.

FrownTown:
Currently going through and adding a snowflake hardcoded variable to every clothing item in the game to allow the damage system to determine what material they should be considered to be made of without actually giving them an armor class, since armor class has so many other effects (don't want someone to be unable to sprint while wearing a pair of plate gloves for example)

May take a while, recommend not TMing this until I finish later this afternoon.

FrownTown:
UPDATE: Completed addition of a snowflake var to every clothing item with an armor value in the game that uses ARMOR_CLASS_NONE that will allow them to be considered as LIGHT, MEDIUM or HEAVY solely for the purposes of interacting with the new blunt armor penetration and armor durability damage mechanics. Things will now behave as you'd expect them to.

Teragenic:
image
So, uh... Are fists meant to pass thru full plate and stuff? 

image
Same as bites

Dustssires:
Fist should not be able to do that

DarknessWithin994:
The -25 blunt AP against leather makes it more or less impossible to kill the ambush bandits that spawn when you step wrong at night if you're a mace user. My Mace Otavan Routier basically couldn't kill one after wailing on it. Leather should be better against blunt than most, but this was overtuning.

TempleOperatingSystem:
As with #4000, my main critique here is the rock paper scissors system. Classifying weaknesses/resistances based on light/medium/heavy is almost completely arbitrary. There's light maille and medium plate-and-leather. There's medium cuirasses (all cuirasses made out of metal are medium, iirc.) How can you justify a cuirass's weakness being piercing or a maille shirt's being slash.

kerillians-netizen:
> As with #4000, my main critique here is the rock paper scissors system. Classifying weaknesses/resistances based on light/medium/heavy is almost completely arbitrary. There's light maille and medium plate-and-leather. There's medium cuirasses (all cuirasses made out of metal are medium, iirc.) How can you justify a cuirass's weakness being piercing or a maille shirt's being slash. I really like the rest of this PR! My complaints about every weapon getting AP are muted by the "armor damage conversion" bit. I just dislike the concept of rapiers and sabres being able to hurt you through heavy armour and similar silliness. Perhaps certain weapons/attacks should have a cap on their AP against different armor classes? E.G fists cap out at 20% AP against plate armour? IDK, just a thought, I digress. I still think that the rock-paper-scissors combat system is the main concern... it makes 0 sense and destroys the uniqueness of different armours. Again, I get the INTENTION, but light armours aren't always "soft armours" like the PR claims, and medium armours... I really can't understand the justification for medium armours being weak to stabbing in the slightest.

Hmm, i hear your worrying about rapiers and stuff hitting throught armor but lets not forget this : You cant get crits nor substantial damage while having armors (Yes you find gaps but from what i understand your vitals are still protected) And for that to happens you need to be in aimed intent ( reduce your dps by 20percent something) And the light bleeding you get should impact you way lower compared to what is it now. It looks goods on the paper, but on the field it might not so much

Again i hope i got it all good and i maybe wrote some misstake here.

FrownTown:
> As with #4000, my main critique here is the rock paper scissors system. Classifying weaknesses/resistances based on light/medium/heavy is almost completely arbitrary. There's light maille and medium plate-and-leather. There's medium cuirasses (all cuirasses made out of metal are medium, iirc.) How can you justify a cuirass's weakness being piercing or a maille shirt's being slash. I really like the rest of this PR! My complaints about every weapon getting AP are muted by the "armor damage conversion" bit. I just dislike the concept of rapiers and sabres being able to hurt you through heavy armour and similar silliness. Perhaps certain weapons/attacks should have a cap on their AP against different armor classes? E.G fists cap out at 20% AP against plate armour? IDK, just a thought, I digress. I still think that the rock-paper-scissors combat system is the main concern... it makes 0 sense and destroys the uniqueness of different armours. Again, I get the INTENTION, but light armours aren't always "soft armours" like the PR claims, and medium armours... I really can't understand the justification for medium armours being weak to stabbing in the slightest.

Well the idea was basically that light armor is cloth and leather, medium is chain (which gets pushed apart by a thrust) and other flexible metal armor, and heavy is rigid metal armor. If you give me more specific examples, I can always make exceptions. Like a cuirass probably should be treated like plate would be, and that is easy enough to do using the armor integrity mod override var I wrote for non-torso/leg armor. I just felt it didn't make sense to go to the effort of writing a material system for it.

Also, just to clarify, rapiers cannot pierce straight through plate. There is a new mechanic that allows using certain weapons to stab into the gaps of heavy armor, which then is affected by underarmor. The mechanic is based on a stat/skill roll. If it fails you hit the armor and almost certainly get the hit soaked.

Based on feedback from the test, I have some changes in the pipe to numbers for a lot of things. Unarmed attacks are going to only gain an AP bonus if the user has expert pugilist. I will also be introducing a soft cap for AP scaling, widening the range at which armor blunts attacks, making some adjustments to the rate of armor degradation and so on.

PR #1388, opened 2025-11-08
Fixes crimson curse silver weakness

UntoldTactics:
## About The Pull Request
Adds the silver weakness trait to stray vampires (crimson curse vampires)
Maintainer also added the trait to damned vice. Yippeee.

## Testing Evidence
image

## Why It's Good For The Game
During the silver rework, crimson curse vampires were missed when the silver weakness trait was added to everything it was intended to effect, including every vampire except stray vampires (crimson curse vampires)


coffeecoffeecoffee:
This probably should've just been the case for the original crimson curse PR and I think I voiced this thought. Not only is it a balance issue, it's also just cool and having crimson curse characters have to avoid silver and be able to be found out is just neat.

Ephemeralis:
I've co-opted your PR to add TRAIT_SILVER_WEAK to the Damned vice also, since it apparently needs it.

PR #1389, opened 2025-11-08
Fix global wound regeneration

Ephemeralis:
## About The Pull Request

[This comment covers this best.](https://github.com/Scarlet-Reach/Scarlet-Reach/pull/1386#issuecomment-3506601603)

TLDR: simple mob wound regeneration was applying to everyone when it shouldn't have (including human mobs). Wounds should be significantly more dangerous now over long periods of time.

## Testing Evidence

image

## Why It's Good For The Game

People probably shouldn't be regenerating like Psydonites all the time by default.

### DIFF
--- a/code/modules/mob/living/life.dm
+++ b/code/modules/mob/living/life.dm
@@ -36,12 +36,12 @@
 		handle_blood()
 		//passively heal even wounds with no passive healing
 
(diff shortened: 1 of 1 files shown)


benj8560:
believe it was also pointed out are bleed rates where balanced around this behavior unknowingly and likely also need to be adjusted with this PR. Regardless based PR. No more magically recovering from broken bones. Likely just as simple as checking what number AP uses without this issue and going from there.

PR #1393, opened 2025-11-09
removes the knockdown/stun ability from churn undead

zeratino:
## About The Pull Request
see title

## Testing Evidence
508772464-419a08e9-ba80-4e1d-96fb-239544002378

## Why It's Good For The Game
let's not give necrans a one-button press that instantly stuns and knocks down every deadite on screen


chemistrymain3:
it isn’t supposed to be doing this to skeletons but does so anyway for some reason

zeratino:
> "Lets not have the class type expressly existing to deal with undead, be capable of dealing with the undead."
> 
> 
> 
> I could understand *changing this to be effective in a different way*, because I can see screenwide-knockdown on 10 people being a problem, but just outright gutting the spell like this? What is even the point of having this spell exist now?

You see, it already blinds the undead and gives them a noticeable stat debuff when all of their stats are 10, making them 8

Which is fine, but the knockdown and stun is applied ontop of this

Which is not fine, so that's what im removing

Fatherless1Xx:
> "Lets not have the class type expressly existing to deal with undead, be capable of dealing with the undead."
> 
> I could understand _changing this to be effective in a different way_, because I can see screenwide-knockdown on 10 people being a problem, but just outright gutting the spell like this? What is even the point of having this spell exist now?

This ontop of Necra's vow, Necrans definitely have enough tools to handle the undead without a free stun.

TheEternalFlame:
I think removing the stun is fine, I don't think it should lose the knockdown though

PR #1398, opened 2025-11-09
Blacksmith / Artificer Apprentice Virtue Changes

ophaq:
## About The Pull Request
- Engineering to Blacksmith apprentice.
- Blacksmith to Artificer apprentice.

## Testing Evidence
Trust me bro it's a 2 line change. Also, it compiled.

## Why It's Good For The Game
Someone pointed this out on the discord and I answered. Blacksmiths already get engineering, but the virtue doesn't give it to you. And you need it to make things like locks.

Apparently having basic blacksmith would be nice on artificer too. This does not include armor/weapon smithing. If you want that, choose blacksmith apprentice.

### DIFF
--- a/modular_azurepeak/virtues/utility.dm
+++ b/modular_azurepeak/virtues/utility.dm
@@ -210,6 +210,7 @@
 	added_skills = list(list(/datum/skill/craft/crafting, 2, 2),
 						list(/datum/skill/craft/carpentry, 2, 2),
 						list(/datum/skill/craft/masonry, 2, 2),
+						list(/datum/skill/craft/blacksmithing, 2, 2),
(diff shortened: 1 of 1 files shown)


ophaq:
> Can you add blacksmithing to artificer apprentice as well

It is kinda the whole point of blacksmith vs artificer is to choose one side. Unless you mean just blacksmithing itself and not armor/weapon smithing. If people want that enough I'll do it. What is artificer apprentice going to do with that, though?

Ephemeralis:
The apprentice virtues already have a comparatively huge amount of value for what they offer, and roundstart engineering directly affects golems & dolls in terms of combat longevity and recovery.

It is possible that blacksmith apprentice may not survive the month as it stands. We'll see.

PR #1358, opened 2025-11-05
Squashes a Few Bugs: No More 30000 Paralytic Spiders Everywhere

Plum-but-gayer:
## About The Pull Request

Removes the dozen or some spiders that can paralyze you from the main roads of dunworld. They were added in the halloween temporary changes and are still in for some reason. I assume this is an oversight.

## Testing Evidence

I am no longer covered in spider bites.

## Why It's Good For The Game

Having a bunch of enemies that can one hit paralyze you on the main low-level roads is probably not very fun for a lot of people. I also have a bunch of homies who hate spiders and I seeing them get jumpscared every round is making me sad. RIP Halloween, see you next year.

PR #1399, opened 2025-11-09
Gets rid of weird tiling in the church + Fixes most of the mapping issues that are easy to fix

KiteAE:
## About The Pull Request

The church has hundreds of different tiles for whatever reason and no uniformity, stone tiles + decorated stone tiles + red moss + blue moss + craftstone all in the same room. This makes it more uniform and less of an eyesore. Also corners didn't line up for some reason? Idk its inconsistent mapping. 

Gives the church vault a single extra tile of space because it was literally 1 tile tall and you had to climb over a chest to get to another chest. Also gets rid of the hole that leads to the wardens that people just patch up every round anyway. 

Also fixes most of the glaring issues with icon stacking (where you stack multiple things on top of each other when mapping) as that contributes to lag and drops frames, especially in the rooms with 3+ plants on one tile x 14 creating over 40+ plant tiles in the same room. 

ALSO ALSO gets rid of the prisons in the basement as im not having a 4 tile square for EPIC PRISON ROLEPLAY of being left in a cell alone the whole round with no ability to escape because you are surrounded by rock walls. 

The are a lot more issues but I do not feel like touching them currently. 

Also fixes all the small mapping bugs that I know of, adds a loadout crate to the boat for latejoins, some stuff in the councillors room that were bugged, a river outside the church, oh and removes spiders on the road thanks to https://github.com/Scarlet-Reach/Scarlet-Reach/pull/1358 for reminding me about that. 

## Testing Evidence

image

image

image

image

## Why It's Good For The Game

People are gonna complain at me for touching the church so it being good for the game is subjective, reduces lag and fixes map bugs though.

### DIFF
--- a/code/controllers/subsystem/nightshift.dm
+++ b/code/controllers/subsystem/nightshift.dm
@@ -84,6 +84,8 @@ SUBSYSTEM_DEF(nightshift)
 		if(!cmode)
 			SSdroning.play_area_sound(areal, src.client)
 		SSdroning.play_loop(areal, src.client)
+	if(mode != NPC_AI_OFF)
+		return
 	if(todd == "day")
 		if(HAS_TRAIT(src, TRAIT_VAMP_DREAMS))
 			apply_status_effect(/datum/status_effect/debuff/vamp_dreams)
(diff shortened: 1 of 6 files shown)


ophaq:
It would be nice if we had stairs like RW's map did right next to the boat exit so people can get to the zone without climbing up/down the cliffside. People do this constantly, it's a high traffic zone. Also, we need a statue around spawn too.

PR #1401, opened 2025-11-09
running out of blood now kills you

zeratino:
## About The Pull Request
when you reach 0 blood volume, the death() proc is called - which kills you lol

## Testing Evidence
yeah so i forgot to take a screenshot but i took a minotaur, had it cut me up - and then died when i bled out

## Why It's Good For The Game
i think that people resorting to decapitations because it takes too long to bleed out is silly
if you've reached 0 blood, you've definitely - without a doubt - lost the fight you were in

### DIFF
--- a/code/modules/mob/living/blood.dm
+++ b/code/modules/mob/living/blood.dm
@@ -215,7 +215,8 @@
 	var/old_volume = blood_volume
 	blood_volume = max(blood_volume - amt, 0)
 	if (old_volume > 0 && !blood_volume) // it looks like we've just bled out. bummer.
-		to_chat(src, span_userdanger("The last of your lyfeblood ebbs from your ravaged body and soaks the cold earth below..."))
+		to_chat(src, span_userdanger("The last of your lyfeblood ebbs from your ravaged body and soaks the cold earth below... Woe, you are perished!"))
+		death(src)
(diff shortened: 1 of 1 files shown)


Xaytan:
Hard disagree with this PR, people bleed out alarmingly fast in this game. You hit 0 blood very quickly. People who are out of blood usually don't stay standing much long after, and as far as combat is concerned the guy in crit with no blood in his body isn't relevant anymore.

People go for decapitate not because bleeding out takes too long, but because half the server are immune to death by bloodloss due to virtues. But people would still decap either way because of the deadite timer and just plain old sadism/desire to grief.

Ephemeralis:
This is already implemented in the combat reworks proposed in https://github.com/Scarlet-Reach/Scarlet-Reach/pull/1386 but with major differences regarding how quickly 0 blood is reached.

I had this functionality enabled during life-opts testing and it regularly resulted in fights ending in less than 15 seconds, which didn't seem very fun for anyone. The entire current gameloop is balanced around hitting 0 blood volume then taking like 3 minutes to 200 oxyloss out.

Going to close this for the moment. We can revisit 0 blood death as a concept if #1386's implementation of it doesn't turn out to feel very good.

PR #1404, opened 2025-11-10
FATAL INSOMNIA: Sleeping while crit no longer heals you

Ephemeralis:
## About The Pull Request

Possibly the most heinous balancejak of all time (and likely to be one of the least popular, too).

- Sleeping while crit no longer heals you. This has been a longstanding (and probably partly intentional) oversight since forever.
- You can no longer initiate sleep while dying.
- If you do somehow manage to initiate sleep while dying, the sleeping buff will clear itself on the next tick.
- If you start to die in your sleep, you will no longer keep healing.

Sleeping used to provide a steady source of blood regeneration and significant wound healing which you could initiate at any time via the hidden sleep verb (which appears to be a vestigial holdover from /tg/station). You could also usually just sleep off most poisons that had any real chance of killing you unless their values were cranked through the fucking roof. Not anymore!

This should also technically have the slight side effect of dying mobs no longer calling `handle_sleep()` which is a fairly hot life proc. Optimizations, yay!

The importance of this change is pretty wide-spanning and hard to articulate outright. Bear with my basic summary here: **crit sleeping currently invalidates a LOT of mechanics present in the game**. Why bother with leeches, antidotes or other anti-tox when you can just sleep off an otherwise lethal poisoning? Why bother with many forms of less-than-ideal healing when you can just sleep through crit? Why bother with bandages (outside of pre-bandaging to frag, which will also be fixed in a forthcoming patch) when you can just sleep and juggle anything less than 2 arterial wounds with the passive blood regen that sleeping gives?

NOTE: This **does not** disable the healing sleep normally gives. It just stops it from happening while you're dying.

Bring friends with you. Bandage your wounds. Be smart. You're going to need it, considering that between this and the all-wounds regen fix made in the last two days, the innate regeneration available to players has been reduced by something like 90%.

## Testing Evidence

image

## Why It's Good For The Game

People should probably die when they are killed. We're returning to grimshart, baby.

### DIFF
--- a/code/datums/status_effects/debuffs.dm
+++ b/code/datums/status_effects/debuffs.dm
@@ -106,7 +106,6 @@
 
 /datum/status_effect/incapacitating/sleeping/tick()
 	if(owner.health < owner.crit_threshold) // no sleep-healing while we're dying.
-		qdel(src)
 		return
 
 	if(owner.maxHealth)


Redrover1760:
Good change

Fn-OS:
Could we get like
Medicine shit that we can use to heal that ISN'T Red if this is getting kept? Like an actual medicine update?
Miracles, Red, and sleep being the ONLY way to heal was already painful for exploring, especially when people don't really do large-person adventuring parties unless it's been planned out OOCly. Adding this gives really NO way to recover if something fucks up and you WOULD HAVE had a chance to recover.
Things like bandages that slowly heal whatever limb they're placed on akin to patches, poultices that heal burns, shit like that would make this acceptable, but right now it only really punishes players, and alot of people are seeming to dislike it.

TLDR add healing bandages and shit that people can craft that isnt fucking Red so people have a chance.

Fn-OS:
> Could we get like Medicine shit that we can use to heal that ISN'T Red if this is getting kept? Like an actual medicine update? Miracles, Red, and sleep being the ONLY way to heal was already painful for exploring, especially when people don't really do large-person adventuring parties unless it's been planned out OOCly. Adding this gives really NO way to recover if something fucks up and you WOULD HAVE had a chance to recover. Things like bandages that slowly heal whatever limb they're placed on akin to patches, poultices that heal burns, shit like that would make this acceptable, but right now it only really punishes players, and alot of people are seeming to dislike it.
> 
> TLDR add healing bandages and shit that people can craft that isnt fucking Red so people have a chance.

Like I wouldn't have a problem with this if there were more ways to heal damage BEFORE going into crit, so you at least have a chance.

Neo-moss:
I do not mean to be disrespectful when I say this, but honestly, this PR _really, really, really sucks._

It severily limits the playstyle of wanderers, and limits a good chunk of the playerbase considering most of us play **alone** and, if not playing a Keep or Town role, have to actively explore and move out in order to get resources.
I absolutely detest the idea of simply removing a core mechanic because you think 'it should happen, after all, that's how it works!', not in this game, I find! This feels arbitrary and unnecessary. Why are you punishing your players by making the gameplay loop expel them, rather than challenge them? This isn't fun to deal with, it feels like the game actively does not want us to play.

This PR does not at all adress issues such as burn damage, and does not add a single good thing that would balance this change out. If it were to provide ANYTHING at all to help deal with damage and improve the gameplay loop, maybe adding something into the most boring Red + Surgery (That is barely used to heal) + Miracles set that LITTLE TO NO CLASS THAT ACTUALLY EXPLORES THE MAP has access to easily. Red is easy to come by- but it depends on players and again, you are most likely playing alone! How are you to acquire the resources to make or buy red as an adventurer? 

If this PR proposed ANYTHING benefitial, instead of simply taking the mechanic away, I would be less displeased. Why not add new means to heal yourself and prevent crit in order to balance it out? Maybe add actual bandages? Ointments for burns? Oils? Herbs? Poultices? Healing Incense? ANYTHING to make the game more interesting to explore!

PR #1408, opened 2025-11-11
Lobotomizes Sawbones&Enforces the 1 slot comment

NotGoldenglow:
## About The Pull Request

As the title says,
For some reason, in the Sawbones slot list, there's two while the comment demands one as,
`	maximum_possible_slots = 2 // We only want one of these because of master in swords. `
So I'm enforcing that by removing the additional slot and turning that into a 1.

Also reduces Sawbones INT from 4 to 2, and gives them 1 CON/END. They retain master swords since they're a 1-slot job. 
A human sawbones with intellectual starts with a whopping 16 int before statpacks. They can still feint, but they're not the '0% for thee, 90% for me' anymore. They'll still be menaces after the necklace, but that' their little treat.

## Testing Evidence

it's a one line change and it doesn't explode the game.

## Why It's Good For The Game

I LOVE BALANCEJAKKING BANDITS
I LOVE BALANCEJAKKING BANDITS
I LOVE BALANCEJAKKING BANDITS

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/antag/sawbones.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/antag/sawbones.dm
@@ -16,7 +16,7 @@
 		TRAIT_DECEIVING_MEEKNESS
 	)
 	subclass_stats = list(
-		STATKEY_INT = 2,
+		STATKEY_INT = 3,
 		STATKEY_SPD = 3,
 		STATKEY_LCK = 3,
 		STATKEY_CON = 1,


NotGoldenglow:
> Bandit class with 5 weapon skill vs Mercenary class with 5 weapon skill:
> 
> image image image
> If anything, the bandits need a buff. Mercenaries have been power-crept to the moon and the church now gets access to every miracle for insane combos like kneestingers+call to arms. PRs like this are exactly how you end up with mercenaries pretending to be bandits because the actual bandits don’t have the skills required to do banditry anymore.

Have you considered that maybe a Mercenary like that shouldn't be allowed in the first place, and the best way to combat powercreep is to prevent it in the first place?

Why can this one random mercenary suddenly have 15+ weighed points and no painstun, a unique weapon with insane force only they can use AND an ability that further increases their damage output?

I don't think no armor/no parry (when active) is enough of a tradeoff for a guy who has old WW-equivalent chops-your-leg-in-two-swings type combat prowess.

Like-- I wouldn't equate a 4 INT Sawbones to that, even if they had Legendary swords as one-slotter. They could have Legendary and +10 to it and they still wouldn't beat that Slayer. It's much easier to tone down the absurd than to accept it and bring everyone up to it's level.

Fatherless1Xx:
> I don't feel like comparing bandits to the dwarf slayer merc is a good comparison because one the mercenary is incapable like completely incapable of wearing armor because of their trait and also they are a purely melee combat class, like you wouldn't probably want to go melee combat with a ravoxian dwarf slayer that has a giant fucking axe, it's easier and safer to use ranged options like crossbows, bows or magic to completely counter them, and that's why you also work as a team not as lone wolves.

The reason this matters is because mercenaries are the bandits’ main enemy. There are two different merc classes with weapon skill 5, meaning in theory all eight merc slots could have weapon skill 5. It’s absurd that the town-sided class can field an army of weapon skill 5 fighters while the antag side is restricted to just one.

And that’s before even factoring in all the other advantages mercenaries get. like not being outlaws, being free to walk in town, and having the guards and adventurers on their side.

Dustssires:
Oh wait, my sleep deprived mind just noticed they have axe 5 that's. Really not great for game balance. Holy shit why the fuck is it master, okay yeah no I see why it's overtuned, they either need to make it expert or restrict it to one.

Fatherless1Xx:
> At one hand here I can see a reason why they need master skill for the axe but on the other hand 8 trollslayer is uh.... Kinda insane.

Between freifechters and trollslayers, there's nothing stopping a full stack of 8 mercenaries with weapon skills 5. The reason we don't see this hapeen is pretty funny. The other classes that don't have weapon skill 5 are more powerful. Mercenary players turn down what we're now trying to restrict bandits from having.

Plum-but-gayer:
I personally never understood why sawbones has master swords? It always seemed more like an expert-dagger guy. Imo we could just make a 1slot 'swordmaster' bandit and let sawbones brew potions and saw bones in peace.

PR #1409, opened 2025-11-11
Church gameloop issues fixes

DelineFortune:
## About The Pull Request

Current:
1. Collect Coins has been renamed to Field Burials. It can no longer be used on player's corpses
2. Beast Form now retains all currently active miracles and no longer grants your patron''s T4 miracles
3. Fixed Crafter’s Covenant exploit: the best plate (Ornate Fluted, +1 CON, +1 END) is no longer obtainable for ~40 mammons. Its 500 now 

## Testing Evidence

image

image

### DIFF
--- a/code/modules/clothing/rogueclothes/armor.dm
+++ b/code/modules/clothing/rogueclothes/armor.dm
@@ -1067,7 +1067,6 @@
 	icon_state = "ornateplate"
 
 	max_integrity = 500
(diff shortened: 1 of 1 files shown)


Redrover1760:
When it asks are you sure if you want to summon an artifact with flavor it uses item/artifact/"name" or whatever

Pestran's artifact is called "medical multitool"

PR #1410, opened 2025-11-11
Hitman Reborn: Total tabard name removal (Sussy imposter edition)

Jackofashepard:
## About The Pull Request

Simple change. I've removed the code that adds names to guards and knights tabards/jupons/whatever.

## Why It's Good For The Game

So I've noticed its near impossible to infiltrate the ranks of the guard/knights even when they are all running around with masks like gang goons because by some magic or loving effort by their mammas their names are sewn onto their tabards boldly for all to see (looking fancy knight jr!). This change simply removes that and gives people the opportunity to infiltrate the guard with some luck unless they get mask off pants offed! I think it will be a interesting new dynamic between antags and keep roles aside from the usual climb over the wall and sneak around/total frag entrance loop.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/manorguard.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/manorguard.dm
@@ -38,19 +38,6 @@
 /datum/outfit/job/roguetown/manorguard
 	job_bitflag = BITFLAG_GARRISON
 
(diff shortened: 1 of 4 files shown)


Jackofashepard:
> We probably shouldn't blanket remove this. EIDISS has covered the correct course you should take with this.

Ah well feel free to close then if you dont want this solution, was just a quick and fast PR I did because it seemed like a pretty decent idea.

Jackofashepard:
image
its fully doable to rename tabards currently btw

PR #1413, opened 2025-11-12
Adds wooden dildo to loadout

JustSomeGalSS13:
## About The Pull Request

Exactly what it says on the tin. It adds the wooden dildo to the loudout options, just below the collars and leashes

## Testing Evidence

image

## Why It's Good For The Game

It's an RP item that most people *can* craft, but it takes time to do so. Let the girlies have their fun without cutting down their own tree.


Brabodie:
dude its novice crafting, no reason to even take it but having it as a choice don't hurt.

PR #1425, opened 2025-11-12
Crafty Villains: Enter The WRETCHED TOILER!!! 

Plum-but-gayer:
LEVER

## About The Pull Request

EDIT: Ready to TM, let's see what people do with it.

TL;DR: I really liked the concept of #1203 and made character concept for it but then that PR died. I made my own heavily inspired by/using some of that code. 

Main differences with old mastermind is that the massive stack of abilities are split between two subclasses: a mastermind guy who gets more magic, orders, and a crystal ball, and a servant guy who gets a big stack of skills, buffs, and a cursed collar on his neck. The other big difference is they don't get their adventurer disguise. I think it's a cool concept but would be VERY maintenance heavy bc every time adventurer loadouts are changed someone would have to go in and update mastermind too.

Class is split into 2 subclasses: the mastermind, who gets commands like sergeant does, gets some message spells, a CRYSTAL BALL(!!) and an evil eyepatch. The other role is the servant, who gets a bunch of crafting skills and a pile of buffs to cheer on their master with. Both get t2 miracles AND t2 spells - master gets 12 spellpoints, servant gets 6. They already have a lot of buttons, so these spellpoints are the first thing on the chopping block if they're abused.

Aside from that... it's a badguy omnicrafter. Hopefully with this we see more creative dungeons/forts/traps etc created by bad guy for the good guys to assault, instead of just doing  5 man hail mary dives into the vault. 

## Testing Evidence
I said machinations TWICE FUCK!! I'll come up with a better desc soon promise (edit; I fixed this but forgot to take a screenie)
image
Subclasses: Mastermind and Servant
image
Mastermind Fit
image
Servant Fit
image

## Why It's Good For The Game

Sometimes people wanna be bad guys and not frag out. Sometimes people wanna RP as a drug dealing backalley surgeon and their only options are a single-slot role that inexplicably has master swords. Sometimes people wanna make giant evil ass castles. In general, this fills a bad guy niche we don't currently have. 

I think there's an understandable 'issue' with the potential for this to be abused by people picking this class and then not taking purpletext and just acting like towner++. I call upon all my strongest validhunters to unveil their ravox scrolls and utterly murk anyone using this class to churn out armor and potions for the keep. I also threw a limit on how many of these guys can be running around in a round to prevent wretch from becoming the Crafting Role.

Balancejaks may comment but note that you are standing on a trap door that leads to a snake pit. This is a -2 con -1 str t2 caster with no DE. Your move.

(I'm kinda on the fence about giving them spellpoints honestly. lmk what you think)

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/wretch/wretchedtoiler.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/wretch/wretchedtoiler.dm
@@ -62,7 +62,12 @@
 	//you get miracles, yay! not as peak as a dedicated miracle class, though
 	var/datum/devotion/C = new /datum/devotion(H, H.patron)
 	C.grant_miracles(H, cleric_tier = CLERIC_T2, passive_gain = CLERIC_REGEN_DEVOTEE, devotion_limit = CLERIC_REQ_2)
-
+	//give minion orders if they're a zizite
+	if (istype (H.patron, /datum/patron/inhumen/zizo))
+		if(H.mind)
+			H.mind.AddSpell(new /obj/effect/proc_holder/spell/invoked/minion_order)
+			H.mind.AddSpell(new /obj/effect/proc_holder/spell/invoked/gravemark)
+			H.mind.current.faction += "[H.name]_faction"
 	var/classes = list("MALICIOUS Mastermind","SNIVELLING servant")
 	var/classchoice = input("Choose your archetypes", "Available archetypes") as anything in classes
 	switch(classchoice)


code/modules/jobs/job_types/roguetown/adventurer/types/wretch/wretchedtoiler.dm
```diff
@@ -0,0 +1,128 @@
...
+	outfit = /datum/outfit/job/roguetown/wretch/wretchedtoiler
+	category_tags = list(CTAG_WRETCH)
+	traits_applied = list(TRAIT_RITUALIST, TRAIT_ARCYNE_T2)
+	cmode_music = 'sound/music/combat_bandit_mage.ogg'
+	extra_context = "Choose between 2 options: being an EVIL mastermind or a WRETCHED servant" //choose between master and servant
+
+	// balance isn't real i picked these bc they're funny
+	subclass_stats = list(
+		STATKEY_STR = -1, //YOU ARE WRETCHED!!!
+		STATKEY_CON = -2, //AND YOU WILL +TOIL+!!!!!!!!!!!
```
zeroisthebiggay: i kinda like the idea of being a wretched supergoon who smiths for the other wretches but smithing stamina loss is super affected by con
maybe the dedicated toiler could flatten their con and maybe get an extra point of endurance at the cost of losing some int to better create the identity of "big strong loyal goon" rather than mastermind with all the brains

UntoldTactics:
I am forever supportive of support stuff. Mmngh wretched TOIL...

kerillians-netizen:
it looks fun and goofy as fuck, i love the idea, maybe instaure a slot limit to avoid all the slot filled with toilers ? ( In case it happens)

code/modules/jobs/job_types/roguetown/adventurer/types/wretch/wretchedtoiler.dm
benj8560: you could also just give them the malum smithing trait. I'd say thats dumb but you know druid wildforms get graggerite blessings and so on so whatever.

For context that trait just makes it so hammering a anvil uses effectively 0 stamina.

it's TRAIT_FORGEBLESSED

Plum-but-gayer:
I think this is ready to TM now. Forgive the kind of lacking description, been busy lately.

PR #1438, opened 2025-11-14
Makes Fast Reflexes (TRAIT_DODGEEXPERT) more reasonable with its calculations

GrayRachnid:
### Changes:
- Softcaps speed at 15, for every point after they only gain 50% of their base value. This amounts to 25% probability at 20 speed, about as equal to guidance.
- Fast Reflexes is now a 1.2 multiplier, rather than the 1.5 it was before
- Fast Reflexes no longer ignores weapon skills. You will get owned by someone who has better weapon skills, this also means that if you don't have your best weapon out while dodging, it will use your unarmed skill(pretty sure it did this before too).
- Fixed some math where heavy weapons and lower speed actually gave you a higher chance to hit as the attacker, lmao.
- Re-enabled OffBalance dodging knocking you on your ass, very few things give the debuff anyways and it sounds funny.

## Testing Evidence
image

## Why It's Good For The Game
This will allow skirmishers to have Dodge Expert again without being the meta pick.

Some math I ran earlier, NOT taking into account the swift/heavy checks:
Defender 13 spd, attacker 8 spd:
DE: 76%
NDE: 50%

Defender 11 spd, attacker 11spd, same weaponskill.
DE: 22%
NDE: 5%

Defender 12 spd, attacker 10 spd, defender 1 weaponskill lower:
DE:34%
NDE:10%

Sawbones scenario: defender 15 spd, attacker 11 spd, defender 1 weaponskill higher:
DE: 80%
NDE: 50%

### DIFF
--- a/code/modules/mob/living/roguetownprocs.dm
+++ b/code/modules/mob/living/roguetownprocs.dm
@@ -712,7 +712,6 @@
 	dodgecd = TRUE
 	playsound(src, 'sound/combat/dodge.ogg', 100, FALSE)
 	throw_at(turfy, 1, 2, src, FALSE)
-	
 	if(drained > 0)
 		src.visible_message(span_warning("<b>[src]</b> dodges [user]'s attack!"))
 	else


GrayRachnid:
I'll revisit this if the combat rework doesn't go through

PR #1440, opened 2025-11-14
[COMM] forces (most) naledians to wear their masks

zeratino:
## About The Pull Request
you can no longer cast vizier/pontifex spells without wearing the naledian mask

## Testing Evidence
508772464-419a08e9-ba80-4e1d-96fb-239544002378

## Why It's Good For The Game

https://discord.com/channels/1389349752700928050/1434887538228592742/1434887538228592742

### DIFF
--- a/code/modules/spells/roguetown/acolyte/general.dm
+++ b/code/modules/spells/roguetown/acolyte/general.dm
@@ -318,6 +318,7 @@
 	recharge_time = 10 SECONDS
 	miracle = TRUE
 	devotion_cost = 10
+	req_items = list(/obj/item/clothing/mask/rogue/lordmask/naledi)
 
 /obj/effect/proc_holder/spell/invoked/regression/proc/get_most_damaged_limb(mob/living/carbon/C)
 	var/obj/item/bodypart/most_damaged_limb = null
@@ -376,7 +377,7 @@
 	movement_interrupt = FALSE
 //	chargedloop = /datum/looping_sound/invokeholy
 	chargedloop = null
-	req_items = list(/obj/item/clothing/neck/roguetown/psicross)
+	req_items = list(/obj/item/clothing/neck/roguetown/psicross, /obj/item/clothing/mask/rogue/lordmask/naledi)
 	sound = list('sound/magic/convergence1.ogg','sound/magic/convergence2.ogg','sound/magic/convergence3.ogg','sound/magic/convergence4.ogg')
 	invocation_type = "none"
 	associated_skill = /datum/skill/magic/holy
@@ -428,6 +429,7 @@
 	var/blood = 0
 	miracle = TRUE
 	devotion_cost = 30
+	req_items = list(/obj/item/clothing/mask/rogue/lordmask/naledi)
 
 /obj/effect/proc_holder/spell/invoked/stasis/cast(list/targets, mob/user = usr)
 	if(isliving(targets[1]))
--- a/modular_azurepeak/code/modules/spells/warscholar.dm
+++ b/modular_azurepeak/code/modules/spells/warscholar.dm
@@ -13,6 +13,8 @@
 	chargedloop = /datum/looping_sound/invokegen
 	associated_skill = /datum/skill/magic/arcane
 	hand_path = /obj/item/melee/touch_attack/rogueweapon/bladeofpsydon
+	req_items = list(/obj/item/clothing/mask/rogue/lordmask/naledi)
+
 
 /obj/item/melee/touch_attack/rogueweapon/bladeofpsydon
 	name = "\improper arcyne push dagger"


RandomGooberCoder:
NGL it should be made indestructible then, I think?
There's no need to ban bad faith actors if the code prevents them from acting in bad faith
(Don't mix indestructible with unbreakable tho)

zeratino:
> NGL it should be made indestructible then, I think?
> 
> There's no need to ban bad faith actors if the code prevents them from acting in bad faith
> 
> (Don't mix indestructible with unbreakable tho)

Wish I could, but that doesn't really exist

Fatherless1Xx:
The intention makes sense, but why hand Naledians a clean, ahelp-free way to get round-removed? 
Having your Naledian mask stolen is almost as bad as being trapped in a cursed mask for the rest of the round for roles whose entire kit is based around casting.
These masks cannot be crafted/replaced either, that's by design.

zeratino:
> this PR was asked by someone on discord, I know the author of the PR just do what he is paid for, and i do not blame you for this idea. I will just explain why this is a bad idea, and why a simple mooddebuff is enought.
> 
> This is the most toxic possible outcome solution, that encourage toxic behavior ( stealing mask, and we already know that will just increase ahelp ticket, and crying in OCC, all of that without bringing much flavor) wich won't impact much the one who paid for this PR( she play pontifex ) but will fuck with all the rest of the naledians player ( mage role who wont be able to do shit without their mask, aka RR)
> 
> It tackle an almost inexistant problem ( most of naledians player already keep their mask, or find an alternative to hide their face, only a minimal substrat that i never seen actually go face uncovered)
> 
> Lore wise ,there is no flavorfull reason to make their mask tied to magic. In the lore we also have non native warscholar, that probably just fought a few time with them, but earn their title and gear and combatstyle ( and thus are not anchored as deeply in their traditions, so technically a foreign warscholar can exist, and if he go back to his homeland, since there is no djinn here, he could forsaken their mask traditions that he just adapted to for his stay) And what the fuck you fight djinn, things that probably could make your mask fly away, and suddenly you are utterly dogshit in battle? Come on..
> 
> "Protect your face traditions" is tied to djinn that are in great numbers in naledi, it's where their traditions was born from. so a mood debuff would be more ok, and make more sense . And what even, if you want to play a native naledians, that want to forsaken their mask for x-y reasons because there is no djinn here. No more reason to hide itself.
> 
> But anyway. This idea is bad, it reduce RP possibility, and force you to play a hard naledian native.
> 
> The only things that can make sense, and that respect the possibility of RP Background you could possibility choose to incarnate ( native warsholar, foreigner...) Is a simple mooddebuff ( The one who paid for this agreed that a mood debuff could work too)
> 
> You are a warscholar, you fought great and dangerous battle with them. You carry their arts, weapons, knowledge, and even gear. And the mask is a part of it, a very important one. By removing it, and still using their arts, you are insulting the memory and the one you fought with.
> 
> A possible mood debuff message ( That would respect both foreigner,native naledians, and the other who want to forsaken their way) could be ( To be modified, this is just a first draft)
> 
> I have removed my mask.. An insult to the one i fought side by side with..
> 
> EDIT: And come on, if you see another warscholar who go face naked on the street, and that bother you. BECAUSE YOU, find its an insult to WHO YOU ARE. Take the adequate action to fix it. THIS CREATE RP POSSIBILITY, i don't understand why you are trying to reduce those funs possible interraction.
> 
> Tell them you wont allow someone using your arts and gears and stuff desecrate your tradition, there is actually so many possibility. And if they are dickhead no rper, make a hitsquad and end his lyfe if you wishes nothing stop you

yeah so here's the thing
you're not supposed to be a war scholar if you're not a hard naledian native. its literally in the name and background when you select the text. a scholar is someone who commits their life to a specific field of study, and they're fleeing naledi because it's a demon-infested hellhole

you can still roleplay and do shit without the mask, you just have to wear it while you're in combat. this is so that people are encouraged to keep to their mercenary slot's identity instead of - ala kashira - dropping all of their unique armor in favor for leather armor 

also, ad hominen/slippery slope

benj8560:
I think I'd much rather a mechanical BOON rather then a huge fuck you for taking the mask off. Mainly due to as others have pointed out people will game this and steal the masks.

Perhaps the most simple boon would be axing a stat off each class and adding a plus 1 to the mask. Would need a alt version for pontifex. This would also mean a impact on hierophant. Mask still worth stealing and something to be careful not to lose. But your round could go on if it happened.


There's also the only valid argument I've heard for them taking their mask off: lack of anthro alt

Smuggytheone:
W PR, keep cooking twin.

pyrzal:
Honestly a bad change, and pretty controversial too. This railroads Warscholar drip for no practical reason. Forcing people to wear a specific mask (that in my opinion doesn't look very good) isn't going to improve their roleplay, and it certainly isn't going to force them to respect the role's lore. Only jobbans will do that. We have to recognize that there's a reason people are ditching the mask, and I don't think "They don't want to RP the role" is the reason for that. Plus, even in these cases, a mechanical change isn't going to force them to RP differently. It's just going to make their sprite look a bit different.

I would much rather see Naledis be forced to cover their faces period in order to cast spells. This respects their lore of covering their faces while also allowing some degree of customization in terms of their outfits. I'd be happy to make a separate PR that expands the allowed list of masks to the parent /obj/item/clothing/mask/ if people agree, since I find this PR pointlessly restrictive.

PR #1444, opened 2025-11-15
Court Chaplain addition as a role in the Keep

BadAtThisGame302:
## About The Pull Request

Adds the Court Chaplain as a role, which I've always wanted for a while, considering Astrata is the one that deems nobles worthy to be, well, nobles by our lore, it makes sense to have a Court Chaplain inside the Keep, acting right below the Priest as his spokeperson in the Keep and keeping the Pantheon influence alive in the court.

**Also would be good if anyone actually mapped a little area for the role in the Keep, since this PR's on hold until something like that happens. Unless I map it myself.**

Also renames some files to their appropriate name, like:
puritan.dm -> inquisitor.dm
butler.dm -> seneschal.dm
manorguard.dm -> maa.dm
bailiff.dm -> marshal.dm
knavewench -> tapster.dm
orphan.dm -> vagabond.dm

## Testing Evidence

Showing me this when I try to local host, idk why!
image

## Why It's Good For The Game

I think the Councillors and Keep have been severely lacking in creating the feel of a Keep alive and always scheming, not to mention the quite frequent Crown vs Church conflicts that'll, with this addition maybe have the ability to ease the tensions.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/courtier/chaplain.dm
+++ b/code/modules/jobs/job_types/roguetown/courtier/chaplain.dm
@@ -85,3 +85,6 @@
 		/obj/item/ritechalk = 1,
 		/obj/item/rogueweapon/huntingknife/idagger/steel = 1
 	)
+
+	var/datum/devotion/C = new /datum/devotion(H, H.patron) // This creates the cleric holder used for devotion spells
+	C.grant_miracles(H, cleric_tier = CLERIC_T4, passive_gain = CLERIC_REGEN_MAJOR, start_maxed = TRUE)	//Starts off maxed out.


GrayRachnid:
It's called be an Astratan Acolyte and walk into the keep, I've played an Astratan Acolyte in the keep plenty of times. I've sent Astratan Acolytes to the keep plenty of times as the priest. Acolytes are not MEANT to stay in the church, they can be anywhere on the map.

ConstantineII:
I mean I'm really not against the idea but this needs more than just a role, It would need unique sprites and an area mapped out for it.  Also changing the file names for certain roles is probably a bad idea for the future as it will make it more annoying to bring over content in the future.

PR #1457, opened 2025-11-16
Allows Wildkin to be from Gronn

Plum-but-gayer:
## About The Pull Request
Allows wildkin to be from Gronn, which they were not allowed to be from for some reason.

## Testing Evidence

image

## Why It's Good For The Game

Lets me play my wildkin who I have been roleplaying as from Gronn for over a year correctly without sinking 3 triumphs a round. 
Wildkin are pretty ubiquitous and are more 'wild' than halfkin are so it's not a huge stretch that they'd be in the cold north war country.

### DIFF
--- a/modular_azurepeak/virtues/origin.dm
+++ b/modular_azurepeak/virtues/origin.dm
@@ -66,7 +66,7 @@
 
 /datum/virtue/origin/racial/gronn
 	name = "Gronnic"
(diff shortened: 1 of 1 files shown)


Kroto-Ker (commented):
I'm fine with this, Gronn is so restrictive in the first place because I was told it was mainly the home of orcs. Might need to ask the lore maint?

Besides that please make sure to fix the typo caused in the origin description.

Plum-but-gayer:
image

PR #1463, opened 2025-11-16
Adds Customizable Species Names

Kroto-Ker:
## About The Pull Request

Adds a new 'Race Title' option to the character menu that pops up when you select Wild-Kin, Half-Kin, or Verminvolk.
This allows you to choose a custom species name reflecting what animal they are based off of. It is optional! You can select 'None' to use the default race name.

image

image

image

image

All the names use a similar naming scheme to the base race name, so it should be obvious what the character's actual race is.

## Testing Evidence

image

Tested spawning in with race title, spawning in with none, and spawning in as a race without access to titles. All worked as expected.

## Why It's Good For The Game

Allows players to more easily communicate what species their character is based on.
I decided to implement this system instead of using subraces, simply because there are way too many species options to cover even with the new subrace category.

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/species/anthromorph.dm
+++ b/code/modules/mob/living/carbon/human/species_types/species/anthromorph.dm
@@ -6,7 +6,7 @@
 	id = "anthromorph"
 	base_name = "Beastvolk"
 	use_titles = TRUE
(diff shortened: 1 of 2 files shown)


Cleric2145:
As long as we don't let people add a custom name. I guarantee we'll have all sorts of goofy sparkledog races and try hards abusing it to name their half-kin "Humen" or something.

Kroto-Ker:
> As long as we don't let people add a custom name. I guarantee we'll have all sorts of goofy sparkledog races and try hards abusing it to name their half-kin "Humen" or something.

No, this PR does not allow that. It pulls from a set list.

Plum-but-gayer:
Request to rename 'Wolf-kin' to 'Volf-kin' for lore consistency

UntoldTactics:
half-rabbit should be half-cabbit

WhatTheFlip339:
Could you change the placement on the UI for race titles to be under origin for continuity of the person looking at it?

PR #1468, opened 2025-11-17
Blocks Dark Elves from Being part of the Inquisition

GrayRachnid:
## About The Pull Request
Otava is racist

## Testing Evidence
image
image

## Why It's Good For The Game
I don't know Frok asked me to do it

### DIFF
--- a/code/modules/jobs/job_types/roguetown/Inquisition/absolutionist.dm
+++ b/code/modules/jobs/job_types/roguetown/Inquisition/absolutionist.dm
@@ -9,6 +9,7 @@
 	disallowed_races = list(
 		/datum/species/lamia,
 		/datum/species/harpy,
+		/datum/species/elf/dark,
 	)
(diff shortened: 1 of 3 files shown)


Vizendel:
Otava is also one of the few places where Drow can be found, ostensibly welcomed?

coffeecoffeecoffee:
...but why would _they_ of all people be racist to _dark elves_
this is like the one role where they thrive quite well and the stigma against them makes the inquisition's shtick more interesting
if frok can explain that would be neat idk fucking anything about otava lore i thought drow could be _from_ otava

saltprjak:
We really don't need any more racelocks (ones that detract from roleplay more than they add to it) on a faction so unpopular both to play as and to play with. Speaking as both Inq and Absolver main.

I already struggle getting more than one orthodoxist before the 2hr mark outside of peak pop. Inquisitors themselves are comparatively rarely played, and almost never played in a way that makes people want to keep playing with them (boatloads of looc salt). I'd go as far as to suggest expanding Inquisitor to dark elves to interest more people to try it out, it's in dire need of fresh roleplay.

PR #1471, opened 2025-11-17
Ports Azure's Weapon Specials

FreeStylaLT:
## About The Pull Request

This PR ports the Special framework. Dunno if you guys want that. It lets you make custom grids with various delays per tile. The code isn't really performant, but I'm not sure how it could be without offloading the coords to a txt somewhere else or something.

I copied it over 1:1 as best as I could, but there are various codebase differences that made it difficult to keep parity with the peripheral changes.

-- Original PR Desc Below ---

These are Specials that are unique to a given weapon (or weapon type, in this PR's case).
They are generally going to be distributed to weapons that **cannot pen** or **do not have exceptional damage**. This may change in the future, obviously.

They are initiated by using Strong stance Rclick. There is a global cooldown for all of them.
They are described on examine, if applicable:
dreamseeker_dD1b8CiifP

### Distracting Swipe.
Longswords and derivatives. Not Estoc or Greatswords, though.
Swings along the right flank. Anyone caught in it will be Exposed for 5 seconds.

https://github.com/user-attachments/assets/52f43b20-293b-4ee7-a4b9-52830acb166f

The user **can move around** after using this, but if they fail to be adjacent to the effect before it actually hits, it will fail.
dreamseeker_RoXzGIO5K1

Cooldown: 18 seconds.
### Shin Prod
Two tiles from the attacker. Anyone caught in it will be slowed down.

https://github.com/user-attachments/assets/1763c738-5532-46f2-b148-2ef33615b30a

The user **can ALSO move around during this one.** Same rule. Just have to be adjacent to the effect when it "pops". This will be an important distinction in just a bit.

dreamseeker_KvjVlA0lg3
Cooldown: 15 Seconds.

## FOR MACES
Not Grand Mace.
### Ground Slam
Three straight tiles from the user in a wave pattern. Getting struck by it will slow the target down and offbalance them. The user is also offbalanced for a short while during the chargeup time.
If the target is struck while ALREADY offbalanced, they will be knocked down to the ground instead.

https://github.com/user-attachments/assets/d8aadded-2a5a-47fb-b2f9-0536ac5dcabd

Note the "do after" charging timer. (the little dots above the guy's head)
**The user CANNOT move while charging this effect without cancelling it.** Though, once the wave starts, they're free to move again.

Cooldown: 20 seconds.
## FOR FLAILS
All of them, as far as I can tell.

### FLAIL SWEEP
Swings in a circle, knocking people back. The more mobs are hit, the stronger the effect applied to all of them.

https://github.com/user-attachments/assets/eaed9e38-c9a1-4ef8-a0aa-17c7ecf8dfb3

The user **cannot move** while channeling this. Not that they'd want to, I imagine. None of these effects have any IFF against the user themselves or anyone else.

It scales with people struck. Just 1 person won't do much besides slow them down:

https://github.com/user-attachments/assets/772cedd1-546a-4c60-9473-4a1407b7f2ee

You saw two, it applied an Exposed effect as well.

At 3 or more we have knockdowns:

https://github.com/user-attachments/assets/05c62a3a-6981-4285-8203-e94c033054b6

And from there it keeps scaling all the way up to 8 victims, but I imagine it doesn't matter much to the average gamer past the knockdown.

https://github.com/user-attachments/assets/46446ec2-22aa-433c-ad88-17351d5eaa5a

Cooldown: 20 seconds.
## FOR AXES
Not Greataxes, though Graggar Axe could get something else entirely tbh.

### Hefty Swing
Getting struck by this will immobilize and expose the target for 3 and 5 seconds respectively.

https://github.com/user-attachments/assets/67bb8b69-2f79-4654-a00d-136d890bd852

Cooldown: 20 seconds.

## Why It's Good For The Game
The framework I made opened up **so** many opportunities for various battle manoeuvres as well as (probably for later) unique spells, I decided to capitalize on that and give non-penetrative weapons some added utility.

All sound credit to Artninja, literally all of the new SFX is from them:
freesound.org/people/Artninja/


ConstantineII:
If there's anything that we're missing that you need added do let me know and I'll add it.

Plum-but-gayer:
Cool as fuck

PR #1478, opened 2025-11-18
New role: Bandit Hostage [ BOUNTY ]

ConstantineII:
## About The Pull Request
Adds a two slot role for bandits called "Hostage (Bandit)". that gives the bandits two hostages to do whatever they want with. 

## Testing Evidence
Screenshot 2025-11-17 192159

## Why It's Good For The Game
Hostage already existed in the code but wasn't used so I gave it some purpose.


Plum-but-gayer:
when's this one dropping? smile

knoxon1:
Actually kinda awful. Bandits have only seven members and having two prisoners each round forces them to dedicate one of their few numbers to watching them. Either the bandits have to ignore the prisoners, giving them a bad experience, or a bandit has to give up their gameplay loop of being a bandit to watch the prisoners, which can be what they want, but could also be what nobody wants because having hostages isn't opt in and you can't opt out, you just get saddled with them.

knoxon1:
Add a bandit dungeoneer equivalent role, either as a subclass or a separate class, and disallow the hostages spawning if one doesn't spawn. Or remove this entirely tbh

PR #940, opened 2025-09-12
removes conversion rite

chemistrymain3:
## About The Pull Request

image

Deliberate bad faith abuse of the conversion ritual turning wretch into a conversion antagonist (it is not one) to do on-the-spot on-the-street conversions of people, typically guards or known fraglords, who then sneak back in and then start fragging/betraying the keep the moment they're given the opportunity.

It was never meant to be used like this and is routinely abused. It was always meant to be a mechanical implementation of a roleplay solution- that you didn't have to ahelp after you successfully convinced someone to convert to Zizo or Baotha or whatever. You'd kidnap them, RP with them, break their will, and then you'd use the rune to convert them.

This is not what's happening, and it's not what the rite is meant for. Instead the rite is being used as a substitute for the roleplay because you can use it to skip that part entirely. Instead it's a two minute thing, rather than actually converting someone to your god.

It's gone now.
## Testing Evidence

It's just a deletion.

PR #414, opened 2025-07-31
Conversion ritual for the ascendants

PoltavskaPraca:
## About The Pull Request
reupload of https://github.com/Scarlet-Reach/Scarlet-Reach/pull/318

+ Gives heretics a ritual that lets them convert tennites, psydonites and ascendant worshippers of the other deities.
+ Graggar sprite, only base white one for now (i'm tired boss)
+ Convertees get buffs to certain non-combat skills AND a stat buff:
ZIZO: +2 reading, +1 medicine, +1 alchemy.
MATTHIOS: +1 climbing, +1 lockpicking, +1 stealing. 
GRAGGAR: +1 athletics, +1 butchering. 
BAOTHA: +1 athletics, +1 music, +1 riding.
+ It is entirely possible to convert anyone, even a priest, in its current iteration. Victims get the option to either accept or refuse. Upon refusal they won't die and just get hit for some damage. In short, you can't just LOLCONVERT people and force them to play along.
+ No cooldown on conversion rites in case of someone fatfingering a thing or two, besides, the rite itself is purely social, no need for a CD.
+ Acolytes get T4 miracles out of the gate, since they start with CLERIC_T4, whereas others get their miracle tiers assigned based on what their previous tier was. If you couldn't cast miracles before, you can't do so now either.

## Testing Evidence

https://github.com/user-attachments/assets/2b08442a-ad8a-4f27-9efe-3b782867f9e6

https://github.com/user-attachments/assets/87bb2f3c-3cde-40b1-9540-0cc6c6899397

image

## Why It's Good For The Game

I sure do love LARPing when there could be just a simple mechanic to make captives of heretics switch faiths. It also incentivizes people to SURRENDER and engage in PRISONER larp, while giving them a slight boost upon conversion.

PR #1491, opened 2025-11-19
Baothan rituals

pyrzal:
## About The Pull Request

Ports the **Fertility ritual** from https://github.com/Rotwood-Vale/Ratwood-2.0/pull/282 and re-adds the **Baothan Conversion** rite from #414 that was removed in #940.

The Conversion rite functions as you would expect, changing the target's patron to Baotha. The Fertility Rite has the following effects. If male: allows the target to be impregnated through anal sex actions. If female: toggles fertility from Sterile to Fertile (if the vagina organ is already fertile, it has no effect). Regardless of gender, the target will also be given a pink tattoo marking on their groin.

Both rituals can be resisted by the target via a popup, so there is no concern regarding prefbreak. You can simply refuse it.

This PR also allows Baothans to climax with no refractory period, and renames 'butt' to 'ass' in ERP verbs because it's more appropriate.

## Testing Evidence
**Conversion Rite**
image11
**Fertility Rite**
image12
**Male Impregnation**
image13
**Baothan Marking**
image

## Why It's Good For The Game

🫃


pyrzal:
> First rite is honestly useless and it would've been much better if it actually given an infinite addiction instead.

I'm not sure why you would think a ritual to convert someone's patron is useless. Even if you ignore the mechanical benefits of being a Baothan, the conversion rite is meant to be a tool for roleplay.
> Second rite is... Well, to me it's something that Eora should have instead, just take away the marking..

Nah.

PR #1497, opened 2025-11-20
Knockdowns no longer disarm, Repulse is a knockdown instead of a paralyze

anything86:
## About The Pull Request
Makes it so knockdowns only force you to drop your items if you don't have one of four 'martial' traits.
Makes Repulse's paralyze into a knockdown.

Mobs without a mind skip this check and are disarmed no matter what.

## Testing Evidence
image
Tested in-game by throwing myself off many balconies.

## Why It's Good For The Game

Effectively makes it so repulse no longer disarms you, but also so that you can't diamond formation the bad guy and just kick him into one of your buddies (or kick one of your buddies into him) to instantly disarm him, no matter how bad he is.

Repulse is an overtuned spell right now and a big part of it is just how it disarms you and throws you away from your weapon with no statcheck or saving throw. It's already good in other ways, it doesn't need the added disarm functionality.

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -61,6 +61,7 @@
 	if(!isgroundlessturf(T))
 		playsound(src.loc, 'sound/foley/zfall.ogg', 100, FALSE)
 		ZImpactDamage(T, levels)
+		drop_all_held_items()
 		record_round_statistic(STATS_MOAT_FALLERS)// Wouldn't this only count if you DIDN'T fall in the moat?
 	return ..()
 


Brabodie:
can we NOT nerf this into 2 second inconsequential lay down(nap) time which grants you 0 advantages because by the time you close the distance the receiver is already half way done with standing up and the very situational throw someone off a cliff, rendered useless by the fact 99% of fights happen nowhere near any zlevel changes?

anything86:
> can we NOT nerf this into 2 second inconsequential lay down(nap) time which grants you 0 advantages because by the time you close the distance the receiver is already half way done with standing up and the very situational throw someone off a cliff, rendered useless by the fact 99% of fights happen nowhere near any zlevel changes?

What do you mean "close the distance", you're a mage. This is a mage spell that is supposed to keep foes away from you. Martials don't get an on-demand unavoidable knockdown for a reason, so if you want to use this for close combat strategies, there's your issue.

anything86:
Some other fun things can be done with the spell as well, like increasing the AoE range or the number of tiles it flings you, both scaled to Arcane skill or intelligence. The range is potentially a detriment due to friendly fire but I'm happy to take suggestions.

anything86:
> does this also change the interaction of kicking knockdowns?

Correct, it would make them not disarm as is stated in the body of the PR. Right now there is no real defense against the kick knockdown (other than being able to parry or dodge the kick), and there is absolutely no defense against it if _someone else_ gets kicked into you. You can't parry that no matter who you are. 

You can be a 20 strength 20 constitution hulking barbarian God and you tumble onto your ass as your flimsy princess hands let go of your minotaur greataxe the **moment** a kobold kicks another kobold into you.

WhatTheFlip339:
Just realized this is also a nerf to lightning bolt and pain crit drops...

anything86:
> Just realized this is also a nerf to lightning bolt and pain crit drops...

Lightning bolt isn't a knockdown and neither is paincrit. Neither are affected by this. Did somebody hire you on Fiverr? I am losing my sanity here.

kiwedespars:
> i don't know if this is feasible by making repulse afflict the unstable effect cause iunno if being thrown into a wall is set to knock down unstable people and if it isn't maybe it should

making it give off-balance would do this yeah. it'd also be more balanced because you'd drop your item where you end up instead of 5 tiles away

code/modules/spells/spell_types/wizard/invoked_aoe/repulse.dm
```diff
@@ -1,7 +1,7 @@
 /obj/effect/proc_holder/spell/invoked/repulse
 	name = "Repulse"
 	desc = "Conjure forth a wave of energy, repelling anyone around you."
-	cost = 3
+	cost = 2
```
chaoko99:
```suggestion
	cost = 3
```

this change does not justify a cost reduction

anything86:
> This is also entirely wrong, classes such as barbarian and some absolvers use crit resistance or expert pugulist

I will add civilized barbarian as another trait check.

> There are also plenty of ranged experts with dodge expert that aren't melee experts as well.

Archers are still martial classes, though the number of classes that get dodge expert and do not get Expert weapon skill in at least one weapon type ought to be really small, no? I can't think of a single example outside of certain Adventurer subclasses, and Adventurer is just undertuned in general.

I'm not sure I get the reasoning for not running a check. Why is that better for the game? It's okay that someone not trained for combat loses their weapon upon hitting the floor, it's more soulful than everyone in the game having a vicegrip on their weapon. What is not okay, imo, is trained fighters losing their weapons for any knockdown, which is straight up a tgstation leftover that was never addressed and has been the status quo for years.

code/game/turfs/turf.dm
```diff
@@ -626,7 +626,15 @@
 		return
 	if(has_gravity(src))
 		playsound(src, "bodyfall", 100, TRUE)
-	faller.drop_all_held_items()
+		faller.drop_all_held_items()
+		return
```
Redrover1760: this code always early returns here due to the z level always having gravity

PR #1498, opened 2025-11-20
Ports burn damage fix

anything86:
## About The Pull Request
Ports https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/4175 

## Testing Evidence
I'm gonna be honest the main branch seems to have been shot at the time I'm opening this PR and I can't get it to compile but its unrelated to this PR.

## Why It's Good For The Game
Makes burn crit math a little less wonky, hopefully.

### DIFF
--- a/code/__DEFINES/combat.dm
+++ b/code/__DEFINES/combat.dm
@@ -49,7 +49,9 @@
 
 #define HEALTH_THRESHOLD_NEARDEATH -90 //Not used mechanically, but to determine if someone is so close to death they hear the other side
 
-#define DAMAGE_THRESHOLD_FIRE_CRIT 150
+// Actually a divisor. Where 1 / this * 100% value of burn damage on lethal zones (Chest & Head) causes you to enter hardcrit. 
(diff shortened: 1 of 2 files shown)


GrayRachnid:
Maybe I just don't understand what the original or is saying. What does this fix in effect rather than theory?

WeNeedMorePhoron:
> > Maybe I just don't understand what the original or is saying. What does this fix in effect rather than theory?
> 
> Burn damage hardcrits you as soon as it reaches 150 damage on your chest/head. It just capults you into hardcrit. This makes it so it respects Constitution and you instead have to deal 100% of the chest/head's max health in burn damage.

Just a technical correction. It hardcrit you when the average damage percentage on chest / head average out to 66% of one of the organ (Yeah, it is complicated). 

The PR changes it to 95% since 100% means you cannot actually aim only one organ. 

The divisor was 150 but it functionally translates to 66% health.

For example, 100 / 300 damage on chest is 33% 
And then 100 / 200 damage on head is 50% 

This adds up to 83%, this hardcrit you under the old threshold. But new threshold, you'd need to stack 15% more damage approximately on each part before you can hardcrit them.

PR #1499, opened 2025-11-20
Changes Featherstep enchantment

anything86:
## About The Pull Request
Makes the Featherstep enchantment give the Acrobatic virtue, rather than +1 Speed.

## Testing Evidence
Awawa.

## Why It's Good For The Game
People keep getting a billion Featherstep-enchanted items and stacking them. It's not an active enchantment, so it can stack at the moment. This makes it so there is no incentive to stack it by trading the speed bonus for a trait.

### DIFF
--- a/code/datums/magic_items/superior_items/superior.dm
+++ b/code/datums/magic_items/superior_items/superior.dm
@@ -37,7 +37,7 @@
 	else
 		active_item = TRUE
 		ADD_TRAIT(user, TRAIT_LIGHT_STEP, "[type]")
-		ADD_TRAIT(user, TRAIT_LEAPER, "[type]")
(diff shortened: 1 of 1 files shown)


GrayRachnid:
> > At least have it right before balance jakking it 😭
> 
> No offense but "it is a free +4 Speed, not a free +6 Speed" is not really a gotcha here.
> 
> > Is +4 a lot? Definitely. However I don't see this as an issue.
> 
> I do, I staff for this server and I've personally witnessed such horrors as Rituos wretch stacking four of these with frostveil darksteel plate and a lifesteal darksteel longsword. It didn't even require that much setup, this was one hour and thirty minutes in.
> 
> It either does not stack, or it goes from +1 Speed to Acrobatic virtue. That's it.

Then make it a max +1 from each item or limit it to just rings. Rings are still rare. +2 max.

PR #1500, opened 2025-11-20
gives the knight-captain too much Knightquil and takes him out back, distributes his drip as a unique Knight uniform

coffeecoffeecoffee:
## About The Pull Request
Small changes: The mounted and champion knights now select their primary and secondary weapons separately now instead of being stuck with sets. You may now take the lance and the longbow, or the rapier and the sling _or_ the crossbow, instead of being limited to premade loadouts. Also, I changed out the billhook option for the halberd, since it was by far the least-chosen knight option and I've never seen anybody use it.

Anyway, a lot of people on the discord seem to feel like the knight-captain takes away a large part of what being a knight _is,_ gets in the way of them doing what they do best — _serving the duke_, being members of court and actually protecting the Keep — and is overall an annoying and unneeded heirarchy. I've simply removed his slot and added a fourth knight slot in his stead.

However, instead of letting all his drip go to waste, I brought all the knight-captain's special armor in line with the integrity of the ordinary versions and made them the new Knight uniform instead! Finally, knights will actually be visually recognizable with a special armor piece and cape. This has the added benefit of cutting down on the excessive number of loadout menus that knights get bombarded with (as why would you ever pick anything but the special knight armor as your chestpiece? It's even got colorable layers!) The old KC's barbute remains only an option out of your selection of helmets. You can still pick whichever one you like. I've subtly reflavored it all to be specially-commissioned knight armor, as well.

## Testing Evidence
noiU1JPtV8

Spawned as knight and:
dreamseeker_bPRmN4Pr33

## Why It's Good For The Game
Check out the feedback thread "if you had to change the garrison's command structure" for a long discussion about this - but basically, knight-captain tends to turn knights into a goon+ role instead of the keep guard and courtier they're meant to be when they're at their best, and knights being another goon+ role is unenjoyable for _everybody._

### DIFF
--- a/code/modules/clothing/rogueclothes/armor.dm
+++ b/code/modules/clothing/rogueclothes/armor.dm
@@ -1617,7 +1617,7 @@
 	sellprice = 150
 	clothing_flags = CANT_SLEEP_IN
 	armor_class = ARMOR_CLASS_HEAVY
-	armor = ARMOR_CUIRASS
+	armor = ARMOR_PLATE
 	body_parts_covered = COVERAGE_ALL_BUT_LEGS
 	max_integrity = 350
 	prevent_crits = list(BCLASS_BLUNT, BCLASS_TWIST, BCLASS_PICK, BCLASS_SMASH, BCLASS_CUT, BCLASS_CHOP)


coffeecoffeecoffee:
> Hmm, okay that feels abit debatable, the normal brigandine just has the ARMOR_PLATE instead of the cuirass, making this essentially the worse option out of them all?, since fluted cuirass is medium and it has the same armor as this, and the brigandine is heavy but it has the ARMOR_PLATE Instead of the cuirass, but eh i cant find the actual stats on each of the thingy so i dont really know the difference, but it could be changed.

yeah i just want to make it identical to one or the other one of the roundstart options i don't care which
whichever people think is best/want
I could also just re-add selecting your own armor again, I just thought knights get SO FUCKING MANY roundstart selection menus.

anything86:
I like this conceptually and I've always been a proponent that the Knights are not goons and are supposed to be on very similar footing, societally, to the Knight Captain. But isn't this kind of getting rid of one of the Keep's strongest assets, combat-wise? There'll be no classes with Master weapon skill.

I think the idea is a bit underbaked right now and KC could maybe get reflavored rather than flat out removed. Maybe as a Royal Champion or somesuch, I dunno. Currently this also makes the unique glaive and sabre/buckler unobtainable unless I'm missing something.

Yellows91:
I wish to throw this here because removing KC feels like a VERY POOR move here.

Its about the only Keep role that is a Master swordsman / polearm fighter at the start and removing it is just asking for hostile forces to have an easier time to take down the City.

Brigands. - Hedge knight / Sawbones (2-1 roles respective) are all Master swordsmen
Foreigners / Mercenaries - They all have Master skills in their own respective gear. 
Inquisitor - Master Swordsman
Various antags - Master / Legendary swordsmen of their own..

Not every Veteran is here to minmax train the Garrison to Master swordsmen and knights alone are on par with Men at arms with their only main advantage being stats.

PR #1509, opened 2025-11-21
[TESTMERGE ONLY] status effects speed refactor

Ephemeralis:
## About The Pull Request

Pretty much just Deline's work from https://github.com/Scarlet-Reach/Scarlet-Reach/pull/1433/files with a few minor changes (and in a testmergeable format).

Changes status_effects to use an assoc list lookup instead of a return-early search operation, which should cut down the number of overall operations on a *huge* amount of calls by orders of magnitude. `has_status_effect` is called in an incredible number of places practically hundreds or thousands of times a second during high load. 

Code note: status_effects is retained and tracked as is for anything that requires it for maximum compatibility.

Players should notice no changes whatsoever.

## Testing Evidence

game open, status effects appear to function identically as they did before

## Why It's Good For The Game

speed run gooderer


DelineFortune:
Thanks 
I have no skills to maintain this code
Only to make some math logic

PR #1511, opened 2025-11-21
Fixes a weakref-related runtime in looping sounds

Ephemeralis:
## About The Pull Request

Noticed this when testing some spell-related optimizations. Atomised it and fixed it.


chaoko99:
this is why we don't crutch on weakrefs and just clear our refs properly 😤

PR #1512, opened 2025-11-21
[TESTMERGE ONLY] spell processing

Ephemeralis:
## About The Pull Request

Simple: when a spell is done recharging, it removes itself from the fast-processing queue instead of endlessly capping its charge_counter value and requesting a datum spell icon state refresh. FP sometimes hits 1700+ objects in it during highpop (most noticable during the deline churchloop era of acolytes having like 20+ spells each), and this should keep that wayyyy down. This wasn't actually producing a huge amount of load (the calls appear to be quite cheap) but you know, it's easy, so why not?

Less things in the FP queue means less things are fighting for subsystem budget during high tidi, meaning that spells actually come back off cooldown quicker and mages/acolytes want to kill themselves significantly less.

I wrote a whole-ass world.time related refactor for spell cooldowns before realizing this'd pretty much do the same thing. I still may do the world.time thing too.

NOTE: there is a possibility that some abilities may not refresh afterwards while this is merged. I tested this locally only briefly and did not encounter any of the obvious cases I could think of, but the possibility still remains. If this issue happens while this is TM'd, this is why.

## Why It's Good For The Game

game run gooderer

### DIFF
--- a/code/modules/spells/roguetown/acolyte/eora.dm
+++ b/code/modules/spells/roguetown/acolyte/eora.dm
@@ -376,6 +376,8 @@
 	else
 		recharge_time = base_recharge_time
 
+	START_PROCESSING(SSfastprocess, src)
+
 /obj/effect/proc_holder/spell/invoked/pomegranate
 	name = "Amaranth Sanctuary"
 	invocation = "Eora, provide sanctuary for your beauty!"


Ephemeralis:
I would rather flense myself than touch spell code for a moment longer, but I weep with envy for the prospect of being away from the abject monstrosity that is the current beast.

Ephemeralis:
so far, so good! only 1 report of spell weirdness and it was a visual issue post resurrection w/ actions not refreshing colors but still being castable

PR #1518, opened 2025-11-21
Adds the Philosopher Lord, a spellcasting Duke

JustSomeGalSS13:
## About The Pull Request

Adds the Philosopher Lord advclass for the Duke. They get T2 spells, 12 spellpoints (15 if old), and the Message spell for free. Skillswise they have Master reading (lol), Expert spellcasting skill, Journeyman alchemy, and only one weapon skill above Novice, which is Apprentice Knives. Other skills mostly taken from Merchant Lord. They also get Mage Armor and Intellectual but no reflexes or armor wearing.

## Testing Evidence

image

## Why It's Good For The Game

We have the fighter duke, a grown up Daring Prince, the mercantile duke, a grown up Sheltered Aristocrat, and the inbred duke, a grown up Inbred Wastrel. This is a grown up Introverted Bookworm. They get shit physical combat, even compared to most mages (who'll have something like polearms), but in exchange they get spells. Only T2 though to keep them being **too** good or stepping on the toes of other classes. Not sure about the tuning on the spell points, might need to go up or down.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/nobility/lord.dm
+++ b/code/modules/jobs/job_types/roguetown/nobility/lord.dm
@@ -229,7 +229,7 @@ GLOBAL_LIST_EMPTY(lord_titles)
 		/datum/skill/craft/alchemy = SKILL_LEVEL_JOURNEYMAN,
 		/datum/skill/combat/polearms = SKILL_LEVEL_NOVICE,
 		/datum/skill/combat/maces = SKILL_LEVEL_NOVICE,
-		/datum/skill/combat/wrestling = SKILL_LEVEL_APPRENTICE,
+		/datum/skill/combat/wrestling = SKILL_LEVEL_NOVICE,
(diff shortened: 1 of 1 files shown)


ophaq:
We should make the master rod equivalent of a dorpel or riddle one so they can channel quicker too.

DarknessWithin994:
> while this _is_ extremely cool and makes a lot of sense, I'd be worried that there _is_ a reason that almost all the magic-capable characters in the Keep are sequestered in something of a mini-faction among the mage tower's staff, and do worry about how absolutely sick this sounds in terms of dukes becoming combatants
> 
> It sure seems fucking awesome though

Off the top of my head, both Prince and Hand have spellcasting subclasses

DarknessWithin994:
They're also limited to T2, so they're not going full combat mage. It's intended to give them an interesting toolkit but not one that's overpowering. Repulse and Bolt of Lightning are neat tricks, but without STR or weapon skills to follow up on the downed foe they're not going to be getting kills.

PR #1519, opened 2025-11-21
[TESTMERGE ONLY] Known people refactor, roundstart optimizations & crash bugfixing

GrayRachnid:
# EVERYTHING BELOW MAY OR MAY NO LONGER EXIST.
# TOO MANY THINGS WERE CHANGED IN THE PROCESS OF OPTIMIZATION AND LIVE TESTING
# I DON'T THINK I CAN DOCUMENT THIS WELL ENOUGH TO KEEP IT SIMPLE. YOU WILL HAVE TO READ THE CODE AND UNDERSTAND IT
# I DO NOT RECOMMEND PORTING THIS PR TO ANY CODEBASE WITHOUT PICKING WHAT YOU WANT FROM IT.

### Player Facing Changes:
- Unknown should no longer show up as people's job, that bug is fixed
- Bandits/Wretches/Adventurers (and anyone else who usually doesn't) will no longer see the name/job of anyone who latejoins
- EVERYONE will know the people in Noble jobs (you know the nobles, after all)

### Take off the kiddie gloves coder man (Checked box means tested)
- [x] First, adds a new subsystem that is called after all players have been loaded. This is now where the knowledge exchange happens. 
- [x] Creates a new signal to track when a job is done going through its advjob setup (This was the biggest issue)
- [x] The latejoin check now ensures the one receiving information is meant to be part of the system (is the bank account var on their job true?)
- [x] Adds a global list and loop in the knowledge system to ensure every role knows about the nobles of the realm

## Testing Evidence
image

I tested what I can test with just 2 players, this will have to be tested live

## Why It's Good For The Game
it annoyed me

### DIFF
--- a/code/controllers/subsystem/ticker.dm
+++ b/code/controllers/subsystem/ticker.dm
@@ -518,15 +518,15 @@ SUBSYSTEM_DEF(ticker)
 		var/player_time = world.timeofday - player_start
 		
 		if(player_time > 10) // Log if any single player takes >1 second
-			log_game("EQUIP SLOW: Player [player_count]/[valid_characters.len] ([character.mind.assigned_role]) took [player_time/10]s")
+			log_game("EQUIP SLOW: Player [player_count]/[valid_characters.len] ([character.mind.assigned_role]) took [DisplayTimeText(player_time * 0.1)]")
 		
 		if(player_count % 10 == 0) // Progress update every 10 players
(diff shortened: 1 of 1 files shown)


GrayRachnid:
DNM until we test it for a week, then i'll clean up a lot of the code and comments. Will come back to this on 12/23 to clean up assuming all is well.

PR #1531, opened 2025-11-22
Removes duplicate equipment from Marshal, gives Sergeant a horn

JustSomeGalSS13:
## About The Pull Request

Fixes the issue with Marshal causing them to spawn with two horns and two daggers. Also gives the Sergeant a horn in their backpack.

## Testing Evidence

I forgot to grab screenshots, but I tested both and they spawn as expected now, with one signal horn (and only one dagger for Marshal)

## Why It's Good For The Game

Bug fix is self-explanatory. Sergeant getting a horn is for a few reasons. Firstly, it's what I usually see done with the Marshal's spare horn currently; intended or not the garrison has been making use of its two horns. Secondly, it's a natural thing for a combat leadership role to have, especially the kind that's supposed to be leading from the front. It lets them get reinforcements more reliably than trying to scream their location into their scomstone. I would give one to KC too if that role itself wasn't on the chopping block.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/sergeant.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/sergeant.dm
@@ -121,6 +121,7 @@
 		/obj/item/rope/chain = 1,
 		/obj/item/storage/keyring/guardsergeant = 1,
 		/obj/item/rogueweapon/scabbard/sheath = 1,
+		/obj/item/signal_horn = 1,
 	)
 	H.adjust_blindness(-3)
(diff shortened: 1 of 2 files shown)


DarknessWithin994:
> Why not two daggers? Alot of classes get 2 daggers as a secondary so why not the marshal too?

It was a bug, not a feature. I'm restoring them to their intended loadout. They've also got a big fuckoff mace which is the weapon they're intended to use, the classes that get two daggers generally don't have a different primary weapon. Marshal doesn't even have knife fighting skill.

PR #1534, opened 2025-11-22
Maps the eoran masks into the church, plus a few minor changes because I can't help myself whenever i open strongdmm

chaoko99:
## About The Pull Request
Replaces the eoramask helmets from my prior pr with masks, because i forgot.

Also adds bars to the stewards and merchant's windows, because everyone complains about that.

Also moves the gatehouse second floor 2turfs forward because the murderhole went over the gate. also gave them bricks instead of rocks, for reasons I cannot discern. Funny, I suppose.
image

seals a hole in the impenetrable wall in the church. Sorry fucko, you will be leaving your hold if you want to mine. 
image

Rearranges the stewardry so the candle on the wall doesn't spawn inside the wheelbarrow.
Additionally, I have added paper to the stewardry lobby, and moved a mailbox inside to optimize spying on your mail.
image 

And other tiny little changes that I can't recall, I'll adjust this once I check the map diffs. I forgor.

## Testing Evidence
mapjak

## Why It's Good For The Game
we can't trust players to be good so this is the next best thing.

### DIFF
--- a/_maps/map_files/dun_world/dun_world.dmm
+++ b/_maps/map_files/dun_world/dun_world.dmm
@@ -73779,6 +73779,7 @@
 /area/rogue/outdoors/bog)
 "tzF" = (
 /obj/structure/fermentation_keg/water,
+/obj/structure/roguemachine/atm,
 /turf/open/floor/rogue/hexstone,
 /area/rogue/indoors/town/dwarfin)
 "tzG" = (


GrayRachnid:
> Also adds bars to the stewards and merchant's windows, because everyone complains about that.

While the rest are fixes or QOL, I don't like a huge balance change like this one being slipped in...

PR #1402, opened 2025-11-10
Garrison "Adjustment" PR, or How A Skirmisher Fucked My Wife So I Shot His Dog

NotGoldenglow:
### Requested by a dear friend!

Okay, here we go.  I'd like to preface this by saying I am an avid Skirmisher player. I love taking two crossbows and being a menace. I speak here from experiece and my own point of view.

**THIS CAN ALL BE ADJUSTED AT THE REQUEST OF MAINTAINERS&CO, DO NOT TAKE ANY OF IT AS FINAL! PLEASE GIVE ME YOUR INPUT!!!**

First things first, let's break this down into three categories: **Nobles, Man-at-arms,** and **Garrison**.

### Nobles!
- Added a new trait and define called `TRAIT_GUARDSMAN_NOBLE` to be given to **Knights and the Captain.** 
This is used to give them a new, separate Keep area buff (relevant later!) as well as other unique gimmicks in the future.
... This also means they've lost the Town buff that all other guards get. The reasoning behind this is that their job is to protect the Keep, not the town. They're already a heavily-statted class (10/11! Points **_WEIGHED_**!) in many cases, and with this buff, they've upwards of 15 points spread around. That's a lot of points. They're very strong. They'll still be strong, but slightly less because...

noble

~~Knights now gain ³**+1 CON, +1 END, +1 SPE** while in the Keep's grounds.~~
~~This should make up for the lack of a Town buff, since their focus is inwards, to protect the Keep and not patrol the town. More CON will make them last longer (as this is their sworn duty to fight to the last), while a small END/SPE bonus should let them move around familiar grounds faster, and easier. The buff is marginally weaker than the Town's due to the fact that Knights are strong as-is, though it could be considered stronger due to the nature of CON.~~

- ³Knights no longer have one uniform buff shared between Town&Keep. Instead, it is split in two parts.
Combined, they'll add up to the **+1 CON, END, SPE, +2 PER** that MAA normally get. 
**THEY STACK!**
Individually, it work as follows:
**While in the Town but not the Keep:** +1 END, +1 SPE, +1 PER.
**While in the Keep:** +1 CON, +1 PER.
Hopefully this'll help them while in the town, while reinforcing that they belong in the Keep and should prioritize it.

- All **Knight** subclasses have lost their minds. Literally. From 4 and 3 INT, down to 2 all-around.
The feint meta and it's consequences have been a disaster to antag diamond formation forbidden circles and fair fights.
[...]

PR #1535, opened 2025-11-22
[TM Only] Gives the Hand a Hand in doling out punishments

GrayRachnid:
### The Hand and the Duke:
Can now Disgrace Knights, which gives them a poop colored examined text, a stress debuff on par with being exommunicated, and stat debuffs within the town and the keep that nullify their stat buffs. The reason they don't simply remove the buffs they get is because the knight buff is currently affected by a Perma-TM PR, and this is my workaround. Knights can regain their honor and have their stats restored by simply having the Hand/Duke use the same button on them again.

#1402 being merged/closed will mean I have to re-do the Knight's Disgrace. Do not merge until that PR is merged.

### The Hand, Duke, and Marshal
Can now fire guards. Fired guards lose their trait, which in turn makes them lose their stat buffs. Fired guards also lose their title (unlike Knights, who keep it). Fired guards cannot be rehired, once they are fired the trait is gone gone for the round. In exchange, fired guards do not have a permanent stress debuff. They're just, free now dude.

## Testing Evidence
image
image
image
image
image
image

## Why It's Good For The Game

After a very long discussion spanning over 2600 messages in a certain feedback thread on discord, it's been determined that the garrison and knights do not respect the social hierarchy. Now, my initial suggestion for disgracing Knights was much more heavy handed than this, but a fellow developer has informed me that the stat buffs are enough to make a difference. The ability has a cooldown so it cannot be used heavily against rioting Knights or Guards. Now, due to this not being tied to the Crown/Throat, I hope it would make the Duke and the Hand a little bit scarier when it comes to the Knights and Guards.

### Why the Hand?
The hand, in my opinion, suffers from a slight identity crisis. Not due to design fault, but just due to how the role ended up being down so far on our downstream. The role is based on the Hand role from LifeWeb, which was typically only unlocked if the Baron player put in the CKey of the person they want to play their hand. This meant that the Baron and the Hand had close OOC ties and reasons to trust eachother (barring shenanigans). No such system exists here, which makes the Hand as the Duke's most trusted person a bit of a weak link depending on the people playing either role. 

That is to say, I want the Hand to be the one responsible for all the punishments and dirty work of the Duchy. The Duke is far too busy listening to 40 whiners to deal with a shitty Knight or Guard.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/nobility/hand.dm
+++ b/code/modules/jobs/job_types/roguetown/nobility/hand.dm
@@ -58,6 +58,7 @@
 		if(istype(H))
 			H.verbs |= list(/mob/living/carbon/human/proc/disgrace_knight, /mob/living/carbon/human/proc/fire_guard)
 			H.hand_disgrace_cooldown = 0
+			H.hand_fire_guard_cooldown = 0
 	addtimer(CALLBACK(src, PROC_REF(know_agents), L), 5 SECONDS)
 
 /datum/job/roguetown/hand/proc/know_agents(var/mob/living/carbon/human/H)


GrayRachnid:
> hjello, this is peak if you have any proposed changes to #1402 that can help accomodate this/make it less aids to implement, i'd be happy to throw them over there!!

It's not hard to implement. It's the fact that it's a test merge. I can't remove the knights buffs unless I fork from your branch and I cba doing that. Instead I'll just apply a debuff that mirrors the 2 knights buffs, based on locations.

PR #1543, opened 2025-11-24
Fixes the bug where Goblins go blind 

twosecondstostop:
- Basically, the goblin eyes code check for "!isgoblinp" for if it should poison the owner or not.  This is from an unused race which is not used by our NPCs or Goblin siegers.
- This caused goblins to be poisoned and die over the course of a round. This was especially bad for besiegers
- This PR adds a check for the goblin mob we use, and if the owner is a goblin, it doesn't poison them. This fixes the issue.
- Screenshot attached. As you can see i am not getting the burning eyes poison messages
image_2025-11-23_223331883

### DIFF
--- a/code/modules/surgery/organs/eyes.dm
+++ b/code/modules/surgery/organs/eyes.dm
@@ -182,7 +182,7 @@
 
 /obj/item/organ/eyes/night_vision/wild_goblin/on_life()
 	. = ..()
-	if (!isgoblinp(owner))
+	if (!isgoblinp(owner) && !istype(owner, /mob/living/carbon/human/species/goblin))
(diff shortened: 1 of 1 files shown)


twosecondstostop:
> goblinp is player goblins, the toxicity was implemented in #613 and I added it specifically so player goblins but no other race could still implant the eyes afaik roundstart player goblins aren't actually the mob type goblin, I'd recommend checking how they work with this code

image
Player goblins still don't get poisoned. This is what I expected, since all this does is add an additional check for the Goblin mob, and if you are one it also doesn't poison you. The check for being a player goblin also still works.

PR #1551, opened 2025-11-24
Total Harpy Annihilation: Fixes fly up/fly down being too fast

Ephemeralis:
## About The Pull Request

do_afters work with deciseconds. This means that 1 second is the value 10, and 3 seconds are the value 30.

The fly up/down harpy counter was using the value **3** instead of the macro **3 SECONDS**. This means that harpies were able to shift up and down a z-level in **300ms** instead of **3 seconds.**

## Testing Evidence

image

## Why It's Good For The Game

This should make them significantly less annoying in just about every regard. There is no way a paltry 300ms delay was intentional. We can look at adjusting it down to 2 seconds or 1.5 seconds if this is too slow, but it's a good starting point.

### DIFF
--- a/code/datums/keybinding/mob.dm
+++ b/code/datums/keybinding/mob.dm
@@ -199,7 +199,7 @@
 		if(C.flying)
 			var/turf/open/transparent/openspace/turf_above = get_step_multiz(C, UP)
 			if(C.canZMove(UP, turf_above))
-				if(do_after(C, 3))
+				if(do_after(C, 3 SECONDS)) // excuse me this was fucking 3 (aka 300ms) before???
 					var/athletics_skill = max(C.get_skill_level(/datum/skill/misc/athletics), SKILL_LEVEL_NOVICE)
 					var/stamina_cost_final = round((10 - athletics_skill), 1)
 					C.stamina_add(stamina_cost_final)
@@ -246,7 +246,7 @@
 		if(C.flying)
 			var/turf/open/transparent/openspace/turf_below = get_step_multiz(C, DOWN)
 			if(C.canZMove(DOWN, turf_below))
-				if(do_after(C, 3))
+				if(do_after(C, 3 SECONDS))
 					var/athletics_skill = max(C.get_skill_level(/datum/skill/misc/athletics), SKILL_LEVEL_NOVICE)
 					var/stamina_cost_final = round((10 - athletics_skill), 1)
 					C.stamina_add(stamina_cost_final)


SadStarfish7:
It's peak. Normalize turning harpies into frybird.

RyszardParys:
Yeah i fw this

Elli-Skala:
If were ideas buying here what about tieing it to the same as climbing speed instead so harpies actually have a use for that skill

Cleric2145:
Harpies should NOT be able to fly more nimbly that bat form and gad form. They have thousands of times more mass than those forms (which also cannot as quickly land, shoot a bow, and take off.) That's like an eagle being more nimble than a hummingbird.

Harpies still have their uniqueness. They can still land and take off with all their armor protection and fire their bow and take off again. Beast form has a sluggish transformation cool down, no armor, and no weapons until they cool down. And they're slower in level flight. Y'all's hint this needed tuning was how there's a harpy bandit nearly every round. That's not because they have an OC who's a harpy bandit, it was because it was an extremely powerful build with a super high skill curve and specific equipment set required to defeat it. There's already not many ranged guards about to defeat them, they'll still slap. Just more responsibly.

NotGoldenglow:
> This is incorrect and not as simple as it sounds. Firstly, you can't wear any armor if you want to fly. At all, anything with an armor class.
> Secondly, you drop everything you are holding when you take off, this includes your weapon. This makes anyone with a decent climbing speed and wood walker better at thing you are describing than harpies.

You're acting as if they don't have sheathes or the ability to just... Put their weapons away? This is only a factor if they're already holding something they can't normally put away, which nobody does.

> Fall damage was nerfed pretty hard and the best thing about it is knockdown I don't actually think there's any difference between falling one and falling several z levels but anyone is free to correct me. Pro tip, if a harpy grabs you, you can just beat them up. Target their legs, they have no armor.

Fall damage is still a factor that destroys your armor massively, and hurts a LOT even after nerfs. Also, that pro tip is not the pro tip you think it is. A harpy can come down, dodge your attacks at 90% and initiate an instantly Z-level transition as soon as they grab you only to instantly drop you before you can even recover from the clickdelay, with a passive grab. Any other movement requires a reinforced/aggro grab, and massively slows down your movement. This is a massive issue and is being exploited by *several* people over the course of *several* rounds. It needs to change, period. There is no counterplay. 

I continue to hold my invite to show you exactly why this is problematic.

CaydenT:
Hey so, playing with this, it's terrible.

Harpies being able to "haha nope" and fly away instantly from any danger was awful. This DOES fix that..... The problem is, now it seems like their ability to actually fly once airborne is completely screwed. Like they are pathetic in the air, and they're trading all of their STR and CON to do it, making them just absolutely abysmal dogshit.

They NEEDED to have 3-4 seconds thrown on their initial windup to start flying, and then all other timers left alone. Navigating Z-levels is not a problem for combat encounters if you are unable to safely get in the air in the first place, a simple tweak to that would have fixed how annoying they are without completely gutting them.

If grabbing, specifically, was also an issue, that should be addressed directly instead of this workaround. They shouldn't even be able to grab people while in flight anyway, with talons occupying their hands. Could've easily removed grab intent from their talons.

Ephemeralis:
Somebody else can do those. I can hardly be bothered, also:

DarknessWithin994:
Self-merging a shit PR because you can't be bothered to fix the problems you created. Classic SS13 maintainer behavior.

Ephemeralis:
This is why I have the funny green button and you don't.

PR #1555, opened 2025-11-24
steel warhammer for footknight

afellguy:
## About The Pull Request

giving footknight's warhammer option the steel hammer instead of the iron one.

## Testing Evidence

Capture d&#39;écran 2025-11-24 184303

## Why It's Good For The Game

The fact that footknight gets the same primary as footsoldier is a bit strange, i thought. people I discussed it with thought it was also strange. So, footknight gets the steel version, while footman gets the iron version.


anything86:
Awesome balancejak, +1 this will save Scarlet Reach

PR #1557, opened 2025-11-24
Reflavors carpenter into Kunstmeister

annikaRU:
## About The Pull Request

Gives the town carpenter some guild architect skill adjacent skills including two notable skills that being master carpenter and master mason, allowing them to get multiple planks from one small log and multiple stone bricks from a single stone

A Kunstmeister is a germanic job that is roughly about what we would consider today an 'Engineer' they were a master in the arts of multiple professions and often worked clerical jobs in the medieval age, that being they were a master in a few skilled trades such as stonework, mining or carpentry and likely blacksmithing too, this is the most equivalent job to what our architect is, we already have a german name for a servant class, why not a towner class.

## Testing Evidence

Compiled on local host, spawned in checking my backpack and it contained no guild key, was given an iron pickaxe properly
image
image

## Why It's Good For The Game

This should allow for players to have skills similar to that of the guild architech while not being as good as the architect in certain aspects, notably with lesser quality equipment and _no guild key_

From experience trying to build anything while only getting a single plank from one small log is a complete pain in the ass and what would take a skilled player about an hour to make a simple house as an architect would take a once again skilled player the entire round to build something pretty small, this should help improve server creativity without having to take up a valuable guild slot!

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/carpenter.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/carpenter.dm
@@ -1,62 +1,65 @@
 /datum/advclass/carpenter
-	name = "Carpenter"
-	tutorial = "A skilled carpenter, able to manipulate wood to suit their needs \
-	building forts and stores, carpenting floors, putting up crosses. You can do it all with enough logs"
+	name = "Kunstmeister"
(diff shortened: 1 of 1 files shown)


Elli-Skala:
While Kunstmeister is a new take and interesting i am mentally lacking and will never remember how to spell this while in game. And im certian many others will also. Towner is likely many people first role and I think it'll be abit strange when they join to select between 
"Farmer
Miner 
Kunstmeister
Wood cutter"

Its abit out of place 

(If I didn't know of this pr id likely assume they're a foreigner by name)

PR #1569, opened 2025-11-26
[AP Port] Message QoL + Projections AP#4620

Cleric2145:
[Original PR here](https://github.com/Azure-Peak/Azure-Peak/pull/4620)
All credit to [hazelrat](https://github.com/hazelrat)

## Testing Evidence
Ported evidence-
### Projection
image

### Message
image

> ## About The Pull Request
> This expands the message spell in the following ways:
> 
> * You may now either issue a message or a projection to the recipient of the spell. Projections are brief visions, comprising visuals, smells, feelings, and anything else that isn't a spoken sentence.
> * More feedback is implemented for the sender - you get a message confirming the success of the spell in addition to receiving the same noise as the recipient.
> * If not anonymous, messages appear in the text colour of the sender.
> 
> ## Testing Evidence
> image image
> ## Why It's Good For The Game
> Projections open up a lot of roleplay potential to the spell that isn't otherwise possible - there's a lot you can do with direct mind-to-mind communications that don't have to just be bound up in the format of dialogue. You can write prose! Obscure visions to interpret! Convince someone they're seeing the future! Get artsy!
> 
> Projections aren't whispered like messages are. This does theoretically open up the bad faith interaction of speaking normally into a projection just to avoid the message being whispered, but this is all adminlogged and it's a pretty particular scenario where it could be abused. Beyond that, the rest of this is just QoL.

### DIFF
--- a/code/modules/spells/spell_types/wizard/utility/message.dm
+++ b/code/modules/spells/spell_types/wizard/utility/message.dm
@@ -87,7 +87,7 @@
 					to_chat(user, span_big("You whisper anonymously into [HL]'s mind: <font color=#[message_color]><i>\"[message]\"</i></font>"))
 				else
 					to_chat(HL, span_big("A brief vision suddenly flashes in my mind, originating from an unknown source: <font color=#[message_color]>\[<b>[message]</b>\]</font>"))
-					to_chat(user, span_big("You anonymously slip a brief vision into [HL]'s mind: <font color=#[message_color]>\[<b>[message]</b>\]</font>"))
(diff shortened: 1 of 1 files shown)


Cleric2145:
It's a lot smoother and more elegant than whispering "[Visions of flames]" awkwardly and being overheard by somebody. It opens the door for more surreptitious use of the spell. Honestly the same discussion could be had about having emotes vs just "Talking like this" *But emoting like this.* 

It's also worth pointing out that there's more to this PR than just that. Unidentified messages are all arcyne purple, while identified messages actually carry to voice color of their sender. The sender also gets feedback now regarding whether the message was received successfully or not, and whether their identity was exposed (by exceptionally high PER stats.)

PR #1571, opened 2025-11-26
Horn Color + Wings to Magic Mirror, Pretties Up Menu

ophaq:
## About The Pull Request
See title. Works fine, but I suggest to TM it to be on the safe side.

## Testing Evidence
Wingscolor
Horncolor
MirrorMenu

## Why It's Good For The Game
We had no horn color option in the menu. It bugged me I couldn't change them to do a gimmick, so I added them. Also, no wings? Well, now I can have nonfunctional wings for my aesthetic and RP too. Plus, you can change their color!

More options to customize with the mirror = good.

I took the liberty of alphabetically organizing the menu and making it look nicer with capitalization, too. It's probably not 100% alphabetically organized according to alphabetical rules but it looks nice regardless and I'm not scrutinizing it. I also renamed some bits for clarification, so dying hair and gradients is less confusing.

### DIFF
--- a/code/modules/spells/spell_types/wizard/utility/mirror_transform.dm
+++ b/code/modules/spells/spell_types/wizard/utility/mirror_transform.dm
@@ -45,7 +45,7 @@
 	if (!H)
 		return
 	var/should_update = FALSE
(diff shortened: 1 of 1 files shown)


ophaq:
> Unsure if its still a thing but I had issues before with sprite stacking when changing ears as a half kin

The way the code is, it's kinda wacky. This doesn't really touch that and I'm not sure how to fix that incident.

ophaq:
> Changing ears to anything then none removes he organ completely, without the ability to get human ears back(not 100% on if this was fixed or not)

That's funny. I guess don't do that. I don't know a fix for that but it would have to be another PR. Try putting it up as a post on the discord and maybe someone will fix it if you can confirm it's still around.

PR #1572, opened 2025-11-27
best effort attempt to make the clinic better and restore the execution platform.

chaoko99:
## About The Pull Request
I remapped the bottom floor to be a sort of rec room instead of the shitty alchemy basement with a genuinely impossible to use surgical room, did my absolute best to squeeze it upstairs. 

Moved the alchemy basement upstairs into an above ground level greenhouse with a cute skylight that lets them see who's stuck in the lobby (I also removed the public recovery beds because I hate you ❤️)  
I also mapped in some lockable doors so people don't fucking sack the place when you open the shutters.
I also gave another bottle of revival juice and some more body bags. 

I squished the top floor a little to make space for the aesthetic change but i can totally just undo that if it sucks.

Restored the execution platform because i notice lately that people just cut heads off in public, which gives a bad moodie, instead of doing it on the guillotine, which gives a good one. 

also moved a little bit of the hedge maze because i wanted to keep that little loitering nook that replaced the execution platform.

## Testing Evidence
mapjak

image
image
image
image

## Why It's Good For The Game

images
Makes the act of being a doctor perhaps suck a little less.

Perhaps.

### DIFF
--- a/_maps/map_files/dun_world/dun_world.dmm
+++ b/_maps/map_files/dun_world/dun_world.dmm
@@ -2267,6 +2267,13 @@
 /obj/structure/stairs,
 /turf/open/floor/rogue/dirt/road,
 /area/rogue/outdoors/rtfield)
+"aDl" = (
+/obj/structure/roguewindow,
+/obj/structure/bars/passage/shutter{
+	redstone_id = "surgery2"
+	},
+/turf/open/floor/rogue/concrete,
+/area/rogue/indoors/town/physician)
 "aDm" = (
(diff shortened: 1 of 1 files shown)


chaoko99:
it's going to sit for a while sadly I have an ear infection and am in a lot of pain

chaoko99:
> I like the apothecary changes, but dislike the fact that the execution platform (big, ugly) is replacing the cute little garden noticeboard plaza (cute and soulful).
> 
> Please keep it... It's peak.

I forgot to screenshot it but it actually still exists, I just moved it to the left. It's by the hedge a
Maze gate.

PR #1581, opened 2025-11-28
Aavnic Accent Update

Kregle:
## About The Pull Request
Updates the Aavnic accent to be slightly more comprehensible... and a little more diversified as was actually intended. The cyrillic unfortunately needed to go as it was causing issues with exclamations and the like.

## Testing Evidence
It compiles and the like 👍 

## Why It's Good For The Game
The Aavnic accent isn't getting chosen atm- I believe this is because it's simply a bit fucky atm. There were some issues that I didn't foresee with some replacements just being a pain in the cunt to manage. This should make things more sovfvl and less annoying.

### DIFF
--- a/strings/russian_replacement.json
+++ b/strings/russian_replacement.json
@@ -51,8 +51,6 @@
 		"water": "voda",
 		"wine": "vino",
 		"beer": "pivo",
-		"alchohol": "vodka",
-		"booze": "vodka",
 		"grandmother": "babushka",
 		"grandfather": "dedushka",
 		"grandma": "babushka",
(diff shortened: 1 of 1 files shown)


WhatTheFlip339:
image
unplayable

WhatTheFlip339:
personally i do not like the changes

Ephemeralis:
sure whatever we can try this since you've at least tried to make it somewhat more legible

Randumb-Phantomb:
> ## About The Pull Request
> Updates the Aavnic accent to be slightly more comprehensible... and a little more diversified as was actually intended. The cyrillic unfortunately needed to go as it was causing issues with exclamations and the like.
> 
> ## Testing Evidence
> It compiles and the like 👍
> 
> ## Why It's Good For The Game
> The Aavnic accent isn't getting chosen atm- I believe this is because it's simply a bit fucky atm. There were some issues that I didn't foresee with some replacements just being a pain in the cunt to manage. This should make things more sovfvl and less annoying.

As a user of the accent, this is terrible.

PR #1584, opened 2025-11-28
harpy flight re-rebalance

ShadowDeath6:
## About The Pull Request

increases flight spell cast time and gives it a telegraph message, decreases time to move z-levels
also increases time taken to move up/down as well as stamina cost if you're holding someone

## Testing Evidence

https://github.com/user-attachments/assets/919adda7-714b-4223-95f5-46575d76d85e

## Why It's Good For The Game

a more elegant (in my opinion) way to balance harpies for combat without making their main gimmick feel awful to use

### DIFF
--- a/code/datums/keybinding/mob.dm
+++ b/code/datums/keybinding/mob.dm
@@ -202,10 +202,10 @@
 				var/athletics_skill = max(C.get_skill_level(/datum/skill/misc/athletics), SKILL_LEVEL_NOVICE)
 				var/stamina_cost_final = round((10 - athletics_skill), 1)
 				var/mob/living/carbon/human/pulling = C.pulling
-				var/time_taken = 1 SECONDS
+				var/time_taken = 1.5 SECONDS
 				if(ismob(pulling))
 					stamina_cost_final *= 2 //double our stamina cost if we're pulling someone with us
-					time_taken *= 3
+					time_taken *= 2
 				if(do_after(C, time_taken))
 					C.forceMove(turf_above)
 					C.start_pulling(pulling, state = 1, supress_message = TRUE)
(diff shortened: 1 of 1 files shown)


code/modules/surgery/organs/feature_organs/wings.dm
```diff
@@ -117,7 +117,8 @@
 							to_chat(user, span_bloody("I am too energetic to control my flight!</br>AGHH!!"))
 							user.Knockdown(10)
 						else
-							if(do_after(user, 10))
+							user.visible_message(span_notice("[user] prepares to take flight."))
+							if(do_after(user, 3 SECONDS))
```
Ephemeralis:
```suggestion
							if(do_after(user, 1.5 SECONDS))
```

Ephemeralis (changes requested):
Needs to be 1.5s minimum. Why? Vibes.

code/datums/keybinding/mob.dm
```diff
@@ -246,25 +250,29 @@
...
-					var/athletics_skill = max(C.get_skill_level(/datum/skill/misc/athletics), SKILL_LEVEL_NOVICE)
-					var/stamina_cost_final = round((10 - athletics_skill), 1)
-					C.stamina_add(stamina_cost_final)
-					var/mob/living/carbon/human/pulling = C.pulling
-					if(ismob(pulling))
-						C.pulling.forceMove(turf_below)
+				var/athletics_skill = max(C.get_skill_level(/datum/skill/misc/athletics), SKILL_LEVEL_NOVICE)
+				var/stamina_cost_final = round((10 - athletics_skill), 1)
+				var/mob/living/carbon/human/pulling = C.pulling
+				var/time_taken = 1 SECONDS
```
Ephemeralis:
```suggestion
				var/time_taken = 1.5 SECONDS
```

ShadowDeath6:
> 1.5 seconds to shift a z-level is still insanely quick (probably too quick), fwiw.

the point isn't to make it completely unviable to move between z levels, it's to make it so harpies can't just bail out of combat instantly or suplex someone to death from a standing grab

Ephemeralis:
You're identifying a completely different issue, namely the fact that flying takes too much stamina over time. Z-level traversal HAS to be manageably slower on anything that isn't a direct round antagonist. 

I'd be personally amenable to making harpy flight less stamina draining in general to accommodate for the fact that you can't just instantly travel out of the danger zone of every single attack in the game in 3-6 seconds (assuming 2 z-level delta). I have a feeling that the insane drains were originally intended to buoy that sort of power, which they were completely wrong to do, because that power should've never been there in the first place.

zeroisthebiggay:
> Zad Form doesn't consume stamina either, and while I think some stamina consumption would be good, there's something to be said about making things consistent, maybe.

zad form is a power exclusive to just one person and doesn't even have functional hands it's not really fair to compare it to harpies picking people up to chain drop them

PR #1600, opened 2025-11-30
Alchemy Changes - Seperation of Mundane & Magic

Kroto-Ker:
## About The Pull Request

Adds four new medicinal chemicals that require medicine skill to craft. They heal better than strong red, but only one specific injury & they come with side effects.
**Calendugoric** - Rapidly heals brute, but induces nausea & toxin damage.
**Milk of Urtica** - Rapidly heals wounds, but causes blood loss.
**Saluvanum** - Rapidly heals burn, but induces toxin damage.
**Symphebrim** - Rapidly heals blood, but induces nausea & exhaustion. (Stamina damage.)

You create these via medicine crafting. First you must create a combination of calendula, salvia, and urtica at an alchemical table called 'herbes de medicia'. You use this in a cauldron with symphitum to create 'medicia precursor'. This chemical can then be used to create the medicines at an alchemy table, using an alchemy vial filled with precursor and the respective herb. Calendula, Urtica, Salvia, and Symphitium.

Some pre-existing potions have been given a new requirement, mana buffer.
Make this chemical by placing ground up arcanic melds and dried mana blooms in a cauldron. You then use it instead of water to make strong variants of potions. (Strong red, mana, etc.)

Health potions have also been very slightly weakened to encourage medicine use and to increase the divide between it and strong health.

Alchemist associate & Apothecary also now receive calendula, urtica, salvia, and symphitum herb seeds upon spawn.

Archivist now receives a summoner's pouch and rite chalk. They also now have journeyman medicine, allowing them to create the new medicines.

## Testing Evidence

image

image

## Why It's Good For The Game

Divides potion creation between mundane (apothecary) and magic (magos thrall, witch) roles.
Ideally, this should allow them both to have their own niche without them stepping on each other's toes.

### DIFF
--- a/code/modules/roguetown/roguecrafting/alchemy/alch_cauldron_recipes.dm
+++ b/code/modules/roguetown/roguecrafting/alchemy/alch_cauldron_recipes.dm
@@ -83,49 +83,42 @@
 /datum/alch_cauldron_recipe/str_potion
 	name = "Potion of Mountain Muscles"
 	smells_like = "petrichor"
-	fillreagent = /datum/reagent/manabuffer
 	skill_required = SKILL_LEVEL_EXPERT
 	output_reagents = list(/datum/reagent/buff/strength = 27)
 
 /datum/alch_cauldron_recipe/per_potion
 	name = "Potion of Keen Eye"
 	smells_like = "fire"
-	fillreagent = /datum/reagent/manabuffer
 	skill_required = SKILL_LEVEL_APPRENTICE
 	output_reagents = list(/datum/reagent/buff/perception = 27)
 
 /datum/alch_cauldron_recipe/end_potion
 	name = "Potion of Enduring Fortitude"
 	smells_like = "mountain air"
-	fillreagent = /datum/reagent/manabuffer
 	skill_required = SKILL_LEVEL_APPRENTICE
 	output_reagents = list(/datum/reagent/buff/endurance = 27)
 
 /datum/alch_cauldron_recipe/con_potion
 	name = "Potion of Stone Flesh"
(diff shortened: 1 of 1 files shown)


ChickyRooster:
Not a fan of the whole "mana buffer" requirement, this locks apothecarians or however you spell it out of making any kind of strong or attribute potions, the same goes for the archivist.

Kroto-Ker:
> Not a fan of the whole "mana buffer" requirement, this locks apothecarians or however you spell it out of making any kind of strong or attribute potions, the same goes for the archivist.

That was the intention! They get new medicines to make up for it.

Ephemeralis:
Okay, so, after mulling on this a bit, here's what I suggest you do:

- as per Chaoko's suggestion above, change the magic potion buffer reagent to aqua vitae. This gives brewers and soilsons (or anyone else who wants a slow craftery round) something valuable to make and trade.
- i'm not sure if we have reactions on this codebase, but I think it is important that we figure out some way to address people  mixing these new apoth-chems together. They should probably become a moderately poisonous slurry, or perhaps produce a new toxic metabolite if they're all in the patient at the same time. I personally enjoy chem mixes a lot but I don't think there's really design space for them in RT's medicine space - stuff like triple red/elite red (which someone needs to fix) is already pretty crazy given most of our chems are designed to be solely consumed.
- milk of urtica should disable water's blood loss restoration (or be disabled itself in the presence of water/strong red) otherwise holy smokes!!! MEGA wound healing.
- this will need mapping changes to ensure apothecaries start with at least a little bit of precursor and ideally, some parchment somewhere telling them what the heck the new stuff does. it's a big change!

PR #1601, opened 2025-11-30
[TESTMERGE ONLY] Re-Adds Crimson Curse

twosecondstostop:
- This re-adds the Crimson Curse virtue, updated to use the new Vampire antag.
- The Crimson Curse makes you a thin-blood, but you can still choose your clan and their unique subset of powers. 
- This PR also adds an extra check for the Crimson Curse so that they never become Clan Leaders, since those get extra traits like infinite stamina.
- Unlike old Crimson Curse, this one is disabled for Wretches, because trying to balance the new different clan powers with different wretch classes would be a fucking nightmare.

In my testing it works fine, the main concern is if letting people spawn as thinbloods would be balanced. Jury is out on that, but old Crimson Curse both had a less severe weakness to sunlight and an adminheal on it's regeneration ability, but also didn't have no_pain. I suspect things will be fine. 

This has the potential to be really buggy or badly balanced, and needs to be testmerged first

### DIFF
--- a/code/modules/jobs/job_types/roguetown/Inquisition/absolutionist.dm
+++ b/code/modules/jobs/job_types/roguetown/Inquisition/absolutionist.dm
@@ -25,6 +25,7 @@
 
 	virtue_restrictions = list(
 		/datum/virtue/combat/hollow_life,
+		/datum/virtue/combat/crimson_curse,
 	)
 
 	job_traits = list(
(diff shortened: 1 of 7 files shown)


Ephemeralis:
CC vampires also need to not be able to diablerie - either on themselves or as fodder for others.

Plaguewalker:
> Alright. I'm working on restricting CC from Garrison Roles, removing diablerie from them, and possibly bringing back sunspurned. I'm also changing it from using a var in the vampire.dm to track CC to using a Crimson Curse trait, which I should've done in the first place (the crimson curse define isn't even used for anything). Blame sleep deprivation.

It's less a combat role problem, and more so roles like bandit.

PR #1608, opened 2025-12-01
Matthimas Comes Early: Matthios grants Thieves' Cant and Churn Wealth buffs

Ephemeralis:
## About The Pull Request

- Matthios grants Thieves' Cant to all followers. Crazy that the literal Thief God didn't do this before. Collaborate with Xylixians and seedy underworld figures to grief the Keep for fun and profit.
- Churn Wealth will now actually straight up just (probably) kill targets with the maximum wealth value, as the Thief God intended. Previously, it did nothing except for a piddly explosion that didn't even set you on fire until recently. Like, if they had over 500 mammons, the spell was just worse than any of its other variants, and it's already not even particularly good!
   - It now does 150 burn damage (spread across all body parts randomly), 50 oxy damage (which will darken your screen slightly) applies 10/20 fire stacks, and stuns you for 6 seconds, AND it keeps the piddly explosion that after the fire_act changes, will ensure you hit 15 firestacks before you're able to stand up again (unless someone drags you out).
   
I have seen **so** many tickets about people wondering why the Churn Wealth spell didn't wreck the guy who is Incredibly Wealthy. Now it does!

I also rewrote the whole block for Churn Wealth's calculations because it was killing me. They should've called that shit Churn Coder instead.

## Testing Evidence

it compiles and that's all you get

## Why It's Good For The Game

we'll see

### DIFF
--- a/modular_azurepeak/code/modules/spells/pantheon/inhumen/matthios.dm
+++ b/modular_azurepeak/code/modules/spells/pantheon/inhumen/matthios.dm
@@ -274,9 +274,9 @@
 			if (501 to INFINITY)
 				thing_to_say = "YOUR FINAL TRANSACTION! THE FREE-GOD REBUKES!!"
 				stun_to_apply = 60
-				fire_stacks = 12
+				fire_stacks = 10
 				fire_dmg = 150
-				oxy_dmg = 100
+				oxy_dmg = 50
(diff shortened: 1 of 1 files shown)


Brabodie:
You are a gem and don't let anyone say otherwise

PR #1616, opened 2025-12-02
vampire disciplines touchups

kiwedespars:
## About The Pull Request

rebalances a few vampire things.

PRESENCE

```
presence is guaranteed (but not as harmful if you int-gap the vampire)

if the victim has a 3 or more int advantage than the vampire, they will be immobilized for a second
if the victim has 2 or 1 int advantage, they'll become immobilized for a shorter or longer duration.
if they're equal, nothing changes, you will walk at them/do as they say.

if the victim is in combat mode, awe's duration is halved.
kneel immobilize 3 seconds -> 4 seconds
cooldowns for awe and dread gaze reduced from 60 -> 30
cost on majesty increased from 35 -> 125
```

POTENCE

```
potence now grants +1 strength per level while active
potence now grants strength unbound starting at level 4
level 3 potence gives you fall-resist + z-level jumping
( might move it to 4 idk it just sounded cool and fits potence )

potence extra unarmed damage scaling removed
```

FAE TRICKERY

```
- fae trickery t1 now selects between 1 and 6 options of bad effect. reason is that an on-command disarm is really, really fucking strong.

- 1 is make them meow. this is harmless. the low roll.
- 2 is dizziness, confusion, and jitteriness. this'll mess someone up just a little bit.
- 3 is zizus. someone said it'd be funny
- 4 is making them close their eyes. might be really funny.
- 5 is tripping/knockdown.
- 6 is disarm.

- fae trickery traps are harder to see now but can be broken and destroyed
- fae trickery brutal trap does more damage + offbalances people before they're thrown (because otherwise it did literally nothing)
- also adds a new sound for them to differentiate it from the xylix laugh
```

DEMONIC (fix)

`no longer nuclearizes you to burn damage after you use the first ability`

CELERITY

```
early celerity costs only 30 blood to have active.
each subsequent level is only +10 to blood cost. celerity level 5 cost is untouched.
level 4 celerity grants leaper trait
level 5 celerity grants dodge expert
```

QUIETUS

```
t1: lasts 20 seconds now instead of 2 seconds, 20 second cd
new t2: you touch someone to give them bloodrot. bloodrot is 50% more bloodloss and 50% more pain for 20 seconds.
new t3: dagon's call. you snap your fingers, curse the last person who attacked you with poison. they'll puke blood probably and take stamdamage. it gives them 3 units of bloodacid poison basically
new t4: baal's caress. you stab yourself with your sword to give it +10 force and +40 ap for the next 4 hits, or for the next 30 seconds.
new t5: it's the shitty ass acidball it's unchanged i'm sorry
```

MISC

```
sun now ignites vampires properly with sunderfire
siren t5 and quietus t5 have costs up
siren t5 cooldown doubled
siren t5 now confuses and has a sound
diablerization now grants you their spent RP rather than unspent RP
thinbloods no longer get disciplines if they aren't sired by someone (no cc disciplines)
transfix will always succeed against someone in compliance mode?
```

## Testing Evidence

i tested it on local because it has to run. mostly just number changes w/ slight little additions here and there, shouldn't be explosive.

## Why It's Good For The Game

rebalances vampires because they're pretty broken right now, sorry if it sounds rambly.

```
PRESENCE

Presence is getting touched because it's just a "be a 17 int guy or die" tree.
What it /was/ expecting of you if it was working properly was that you beat out the vampires int by **5** in order to not be forced to walk at them for 9 seconds, or run away for 9 seconds. 
This fixes presence and makes it less overbearing, might look a bit weak even with the cooldown changes, so I'm happy to tune things around. 
Majesty also gives you basically melee immunity for the cost of (350 vitae per 50 seconds), so now that's just been tuned up to 1250 vitae for 50 seconds. Still probably broken.
```

```
POTENCE

Potence is in a weird spot where it's technically strong in that it gives you a +40 to unarmed, right?
But the +5 to weapon damage is a hit or miss depending on your weapon, and it actually doesn't provide any strength otherwise.
This should make it feel way better to invest in and also give the weaker vampires some form of strength to match what it is.

edit: above is incorrect. you get a 2.5 increase to strength, which is 25%, but it still respects the softcap. this sucks for something you need to invest in.

```
```
FAE TRICKERY

Having a 1 minute CD disarm that /all/ your thralls have access to is pretty broken, so it's just no longer a guarantee while still having a few effects to pick from.
Might be stronger in groups since they can hit you with a bunch of different small tricks, but not losing your sword constantly is a good step up 
Also makes their traps breakable but they might still be really hard to see.
also brutal was just useless considering we don't actually have knockdowns on hitting a wall or anything for that matter.
```

```
CELERITY

someone said the early celerity was shit. 
i kinda agree but it justifies it by level 4-5 celerity being god's gift.
late is just untouched while early is a bit less taxing.
```

```
QUIETUS

old t3 poison-stocking was kinda lame in that you could just have a bunch of daggers to instakill people so i replaced it with something more active.
it's mostly just personal touches i don't really have a good explanation for them other than that quietus felt kinda underwhelming or too strong with no inbetween (dagon's call literally stamcritting people for fighting you)
```

if you have any thoughts or concerns/ideas ping me on the discord or just put it down here. especially looking for opinions on the disciplines that i haven't touched on (i have no idea how people feel about them)


RandomGooberCoder:
Actually awesome, +1 ERP point.

Though, presence kinda raises questions -
```if the victim has a 3 or more int advantage than the vampire, they will not be affected```
If I dump INT and get converted by a VL who has presence in their covens - welp, I'm fucked!
I'd suggest instead giving presence the absolute minimal duration, say even if intdiff is >= 5, the effect will still be applied but for 1 SECOND or something

PR #1623, opened 2025-12-03
Resizes sickles, enchantment scrolls and love potions

zVineatic:
## About The Pull Request

Resizes a few items and adds new storage options.

## Testing Evidence

image
image

## Why It's Good For The Game

The love potion itself is not actually a potion in the code and just an item. But it feels like we are carrying a cauldron in our bag since it was 3x3 and couldn't be attached to our hips. Made it act like an actual bottle even if its not one.
Enchantment scrolls are also included, takes a lot of work to do and its shape is 3x3, a pain to carry. Now it is 1x1.
Finally the sickle, its size is now 2x1 instead of 3x3. Glory to the soilsons.

### DIFF
--- a/code/datums/magic_items/mages_mechanics.dm/enchanting_scroll.dm
+++ b/code/datums/magic_items/mages_mechanics.dm/enchanting_scroll.dm
@@ -8,6 +8,10 @@ T1 Enchantments below here*/
 	desc = "A scroll imbued with an arcane enchantment. Can be used on certain items to imbue them."
 	icon = 'icons/roguetown/items/misc.dmi'
 	icon_state = "enchantment"
+	grid_height = 32
+	grid_width = 32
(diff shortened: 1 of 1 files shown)


Brabodie:
Good fucking shit, If you want to knock off 2 birds with 1 stone though, I'd like to let you know that spell scrolls and drained spell scrolls are 2x1 or similar, could use the touching up to 1x1 too

zVineatic:
> Good fucking shit, If you want to knock off 2 birds with 1 stone though, I'd like to let you know that spell scrolls and drained spell scrolls are 2x1 or similar, could use the touching up to 1x1 too

i may leave those as they are. I edited the previous stuff because their size or weight was not specified and that was causing them to be so big by default. And I think those scrolls you talk about are dungeon loots as well so they should stay 2x1

PR #1630, opened 2025-12-04
[TM] Re-Adds Crimson Curse V2, Deluxe No Diablerie edition

twosecondstostop:
- This re-adds the Crimson Curse virtue, updated to use the new Vampire antag.
- The Crimson Curse makes you a thin-blood, but you can still choose your clan and their unique subset of powers.
- Crimson Curse vampires can not use Diablerie or be Diablerized, and can never be Clan Leaders.
- New CC is stronger or weaker then old Crimson Curse in some instances. It's been role-restricted from Bandit & Wretch due to being impossible to balance with the combination of strong classes + varying clan disciplines. I also restricted it from most Garrison Roles but this might not be necessary (Exceptions are Squire and Dungeoneer, which should be harmless(?))
- Unlike the previous edition, this tracks Crimson Curse using a trait which I should've done in the first place, and the check for Clan Leader works. 

Did not have any bugs in my testing, Vampires can still Diablerize other vampires fine but can't Diablerize or be Diablerized by Crimson Curse vampires. Crimson Curse Vampires also could not become Clan Leader, but regular vampires joining a clan still became Clan Leader.

Definitely needs to be test merged first for bugs + balancing issues

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/dungeoneer.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/dungeoneer.dm
@@ -29,6 +29,10 @@
 	round_contrib_points = 2
 	social_rank = SOCIAL_RANK_PEASANT
 
+	
+	virtue_restrictions = list(
+		/datum/virtue/combat/crimson_curse, //no gamer MAA vampire abuse, sorry
+	)
 	cmode_music = 'sound/music/combat_dungeoneer.ogg'
 
 	job_subclasses = list(
--- a/code/modules/jobs/job_types/roguetown/garrison/squire.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/squire.dm
@@ -26,6 +26,10 @@
 	round_contrib_points = 2
 	social_rank = SOCIAL_RANK_PEASANT
 
+	virtue_restrictions = list(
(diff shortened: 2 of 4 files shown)


chemistrymain3:
> Let em be guardsmen and Knights, if bad faith actors take it and ruin it. It should be because of the person themselves risking a job ban. Not the entire virtue with people actually adding substance to the round who take it. If not then let em be wardens, it's a role that's sequestered enough to their own devices.

No.

kiwedespars:
image

> Exceptions are Squire

remove it from squire. still keepsided dodge expert.

giving it infinite blue + nopain and regeneration on top of all that is mildly coal

twosecondstostop:
> This cannot be available to garrison/keep combat roles in any capacity while it has pain & bloodloss immunity present. Had to un TM it.

I updated it to be fully role restricted from all garrison roles, period, no ifs or buts. This includes Squire, Dungeoneer, and Veteran/Town Guard since I realized I forgot those last time.

ConstantineII:
A more up to date version made by the original creator of the virtue is up right now, closing this. 

https://github.com/Scarlet-Reach/Scarlet-Reach/pull/1734

PR #1640, opened 2025-12-04
Longjump Race Rebalancing

Kroto-Ker:
## About The Pull Request

People have informed me that -1 CON for an extra jump tile isn't a worthwhile trade to most people, and may be putting them off from playing fluvian. 
This removes the CON malus, but adds a new stress text to fluvians. It tells the examiner that fluvian wings are valuable and clicking on it gives a short guide on how to cut them off. 

I believe this is more of a reasonable counterbalance to the extra jump tile.

image

image

## Testing Evidence

image

## Why It's Good For The Game

Should make the race's benefits more equally weigh with its negatives.

### DIFF
--- a/code/modules/mob/living/carbon/human/examine.dm
+++ b/code/modules/mob/living/carbon/human/examine.dm
@@ -786,6 +786,8 @@
 			if(H.dna.species.name != dna.species.name)
 				if(dna.species.stress_examine)//some species don't have a stress desc
 					. += dna.species.stress_desc
+				if(istype(src.dna.species, /datum/species/moth))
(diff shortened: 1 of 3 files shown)


chaoko99:
BABY CAN'T HANDLE -1 CON
SO WE GIVE MECHANICAL INCENTIVES FOR HATE CRIMES
LMFAO

chaoko99:
I would suggest increasing the worth of fluvian wings substantially to incentivize the kind of griefing you are about to incur.

Ephemeralis:
Gently, I don't think this is the way to do it. -1 CON is negligible. I think you should just ignore the idiots whining at you over this.

PR #1642, opened 2025-12-04
Veteran Tweaks

SadStarfish7:
## About The Pull Request

I've noticed we have a lot of veterans picking "Middle-aged" instead of "Old", which means that they're not able to actually TEACH keep roles much at all. This PR aims to fix that.

Specifically, it gives a good deal of veteran roles GOODTRAINER iff. they're middle-aged. This is to incentivize veterans actually training people. Too many of them are picking middle-aged to be anime protagonists with no responsibilities (why train under a veteran when there's four knights who can train you to expert just as fast?
## Testing Evidence

image

## Why It's Good For The Game

Veterans need to do things! Incentivizing them to not do the thing vet is made to do is bad and contributes to so many veterans wasting the slot. EIther we make them old only or let them be middle-aged.

Note: Some of the veteran subclasses are hot garbage and could use some love. I'll likely take a look at those in another PR.


Elli-Skala:
> Ah yes let us restrict a very delicate role that doesn't have that many dedicated players AND is very important in combat situations with the constant infinite wars and force them to all to rewrite their characters to be old AND suffer mega stat penalties FOR being old... Yes this will surely be good for this role and will totally not cause this job to end up being unplayed every round.

Thats not what this pr does.

GrayRachnid:
My way to balance this is to instead give all veteran subclasses old-age skills but also apply old-age stats to them.

PR #1644, opened 2025-12-05
Re-Adds Holy Steel Smithing and changes 'swift end's with 'swift jour…

zVineatic:
## About The Pull Request

Holy steel weapons can be forged again. Also changes the titles of swift end with swift journey across the code, given that the weapon itself is called 'swift journey'

## Testing Evidence

image
image
image

## Why It's Good For The Game

Holy steel forging was disabled when the ritualist virtue was still a thing and smiths were spamming holy weapons with it. No reason for them to stay disabled when the virtue itself is gone.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/church/templar.dm
+++ b/code/modules/jobs/job_types/roguetown/church/templar.dm
@@ -315,7 +315,7 @@
 		if(/datum/patron/divine/noc)
 			weapons += "Moonlight Khopesh"
 		if(/datum/patron/divine/necra)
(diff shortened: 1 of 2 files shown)


RandomGooberCoder:
I believe holy anvil recipes were disabled for a reason, as it allows *everyone* to craft them. Which is not the goal, at least I would prefer to keep them unique.

RandomGooberCoder (approved):
Eh, fuck it. 
Since malumite steel can only be done via malumite rite or smelting the deep duke's stuff - IG it is fine.

PR #1651, opened 2025-12-06
unarmed nerfs 2: the unswiftening

kiwedespars:
## About The Pull Request

swiftstam modifier (FOR UNARMED) up to 5 from 1.5
grab stamcost from 10 to 2

## Testing Evidence

numbertweaks.

## Why It's Good For The Game

swifted unarmed punches cost 2 stamina. i apologize for all prior misinformation on the matter i looked at the wrong bit of code when i thought they costed 5 to begin with.

this makes them 5 stamcost on swift so they don't just throw a barrage of punches and never stam out

everything else in the game costs 10 stam to swift.

this means you could throw 5 punches for the equivalent of throwing 1 swifted dagger attack.

### DIFF
--- a/code/_onclick/other_mobs.dm
+++ b/code/_onclick/other_mobs.dm
@@ -30,7 +30,7 @@
 	SEND_SIGNAL(src, COMSIG_HUMAN_MELEE_UNARMED_ATTACK, A, proximity)
 	var/rmb_stam_penalty = 1
 	if(istype(rmb_intent, /datum/rmb_intent/strong) || istype(rmb_intent, /datum/rmb_intent/swift))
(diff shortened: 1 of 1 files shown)


kiwedespars:
> random balancejak
> 
> ah yes giving a 7.5x, up from 1.5x stam modifier to swift/strong punches is logical, you think.
> 
> No, no its absolutely not.

yes. because the swiftstam modifier is 10.
keep in mind that every intent in the game is 1 stam on miss, 1 stam on hit.

keep in mind, that prior to azure buffing it the same way we did, it costed 5 on hit and less on miss.

It's also not random because everyone's complaining that unarmed is oppressive, and as someone who plays it, I agree. I was under the assumption swifted unarmed costed 5 per swing, when in reality it costs (2), while also having insanely high potential damage.

Redrover1760:
> yes. because the swiftstam modifier is 10.

ok that makes more sense

PR #1661, opened 2025-12-06
Mage nerfs 1: The unkeep slottinging

WhatTheFlip339:
## About The Pull Request

Now that the keep mages get free message spell, they essentially free up 2 spell points for anything else. I am reigning it in a little, so they have a -1 in spell points. Also, a known build is bow wizards which is surprisingly strong. Guidance, arguably one of the best buffs in the spell list is also only 2 spell points? Of course I am taking it. Here now that it is at 3 spell points it is in a better state as well as it being canceled on movement now. 

I removed some slots from keep mago's, it has so much impact and with the duke now having the ability to run a mage build.... There is so much power in the keep, lowering it to 3 slots is better overall. 

Never see people run counterspell, I think it can increase skill combat especially wizard versus wizard combat. Here is giving it a little more room to shine in those really special situations potentially making or breaking a fight.
## Testing Evidence

Basic changes and numbertweaks.

## Why It's Good For The Game

See above

### DIFF
--- a/code/modules/jobs/job_types/roguetown/youngfolk/mage_apprentice.dm
+++ b/code/modules/jobs/job_types/roguetown/youngfolk/mage_apprentice.dm
@@ -3,8 +3,8 @@
 	flag = MAGEAPPRENTICE
 	department_flag = YOUNGFOLK
 	faction = "Station"
-	total_positions = 4
-	spawn_positions = 4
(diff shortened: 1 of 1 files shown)


Ephemeralis:
You've lost balancejak privileges, possibly permanently.

PR #1681, opened 2025-12-09
Makes the Avatar of Graggar ogre more obviously antagonistic

Ephemeralis:
## About The Pull Request

Ensures the Avatar of Graggar player knows they're an antagonist and also adds an examine text to them that paints them as an allcaps MARAUDING OGRE so people do the same thing they do when they're confronted with The Text: frag.

## Why It's Good For The Game

If the heir hires the figurative avatar of the organ-eating, conquest-driven death god one more time j'am going to kill monself.


John-Dang:
As the guy who came up with the idea and had it commissioned, yeah this is very much in the vision of it. Can you add images?

KiteAE:
Please fix the conflicts on this.

PR #1687, opened 2025-12-09
Kick targeting & Strangle stamdrain

EvenInDeathIStillServe:
## About The Pull Request

kicks will autotarget whatever mob is on the tile if you kick at the floor
strangling will drain the target's stamina, so they will stam out before actually passing out
can strangle through armour now, with a malus of 66% damage reduction from the neck armor's slash protection

## Testing Evidence

image

## Why It's Good For The Game

image


Quink-Fix:
how is bro gonna strangle thru a gorget..?

PR #1694, opened 2025-12-09
Who Even Asked For This: Bloodloss healing adjustments (mostly nerfs)

Ephemeralis:
## About The Pull Request

Put simply, red pots heal far, far too much bloodloss. With the increasing proliferation of elite red mixtures (lesser red + red + strong red), people were healing over 40% of their total blood amount with three sips from a bottle in about 10 seconds. This practically necessitates the need for sweaty medicine predosing.

So, this PR does as follows:
- Fixes an apparently botched merge that was causing water to heal almost as much bloodloss as strong red (???), reducing +15 to +5.
   - Yep, this is an actual fix, not a poorly-veiled nerf. It was never meant to be +15. I am 100% certain on this.
- All forms of red potion (lesser, normal and strong) have had their bloodloss healing reduced by 70%.
   - However, their effective blood loss healing is now doubled when you are severely low on blood, meaning that they pan out to be only 45% worse than they were before on critically-injured patients.
   - What this means is that they'll still stop you from dying, but if you combat-chug them, you're going to be spending a LOT longer in bleeding tiers 2 and 3 and eating the stat hits that accompany them

Miracle healing is presently unaffected. This does sort of establish its niche at being particularly excellent at blood recovery compared to alchemical means, which is the complete opposite of how it was before this PR.

Values subject to change.

## Testing Evidence

image

## Why It's Good For The Game

We cannot do anything interesting mechanically with healing while potions are as strong as they are. How the fuck do you squeeze **anything** into the medical gameplay design space when people can heal 40% of their blood in 3 sips?

This will probably be a very unpopular change but it is **sorely** needed.

### DIFF
--- a/code/modules/roguetown/roguecrafting/alchemy/reagents.dm
+++ b/code/modules/roguetown/roguecrafting/alchemy/reagents.dm
@@ -17,7 +17,7 @@
 
 /datum/reagent/medicine/minorhealthpot/on_mob_life(mob/living/carbon/M) // Heals half as much as health potion, but not wounds.
 	var/list/wCount = M.get_wounds()
-	var/blood_restore = M.blood_volume <= BLOOD_VOLUME_BAD ? 2 : 1
+	var/blood_restore = M.blood_volume <= BLOOD_VOLUME_BAD ? 7 : 6
 	if(M.blood_volume < BLOOD_VOLUME_NORMAL)
 		M.blood_volume = min(M.blood_volume + blood_restore, BLOOD_VOLUME_POTION_MAX)
 	if(wCount.len > 0)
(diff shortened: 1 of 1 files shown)


Redrover1760:
I think everyone is tired of potion meta

Plum-but-gayer:
I was wondering why water was so good at healing blood! I would prevent people from bleeding out by like, giving them 3 sips of water while they were a bloody mess on the floor.

Ephemeralis:
I'll adjust these upwards so that every red pot is at LEAST very slightly better than water (1/3 -> 6/8, etc).

Ephemeralis:
Okay, they now restore about 1/3 less blood than previously (basically flat increased by +5 across the board), since they inherit the +5 restore that water has by default.

You can stack both for quick blood healing. Still might be too much. We'll see.

PR #1697, opened 2025-12-10
Changes MINSTR on all wieldable weapons to account for wielding halving the minstr requirement

GrayRachnid:
## About The Pull Request
Wielding a weapon halves the MINSTR requirement, meaning that a lot of wieldable weapons with minstr were completely moot. This changes that.

### AXES 
Oath axe: 12 → 24
Iron greataxe: 11 → 22
Steel greataxe: 11 → 22
Double-headed steel greataxe: 12 → 24
Minotaur greataxe: 15 → 30
Troll axe: 13 → 26
Dreamscape greataxe: 13 → 26
### BLUNT WEAPONS
Goedendag(grand mace): 10 → 20
Deep duke's staff (grand mace): 11 → 22
Tetsubo: 15 → 30
### POLEARMS 
Spear: 8 → 10 (adjusted, not doubled)
Halberd: 9 → 18
Silver halberd (Psydonian): 11 → 22
Eaglebeak: 11 → 22
Greatsword: 9 → 18
Silver greatsword (Psydonian): 11 → 22
Estoc: 8 → 16
Partizan: 10 → 20
Naginata: 7 → 14
Ogre sword: 15 → 30
Ogre mace: 15 → 30
Ogre Graggar mace: 15 → 30
### SPECIAL WEAPONS 
Militia greataxe: 10 → 20
Militia spear: 8 → 10 (similar 2 spear)
### SWORDS (bug fix only)
Psydonian shortsword: Removed duplicate minstr line

### The effect:
If you do not have the minstr to use a weapon, you will instead deal 30% of its max damage. (As in, your damage is reduced on it by 70%). 
This, combined with #1694 will mean that a lot of combatants will lose the ability to do effective damage the more bled out they are when using two handed wielded weapons. Swords and other smaller weapons do not use minstr to an effective enough degree to warrant the same changes, but this can be balance jakked as seen.
I am considering this a fix of a design flaw rather than a bugfix or a balance change, but the balance implications are still there.

## Testing Evidence
image

## Why It's Good For The Game
I am confident that high minstr was not designed to be bypassed on wieldable weapons by just.. Wielding them.
I believe this will reign in the fact that some of the strongest weapons that were previously thought to be gatekept behind minstr are actually gatekept by minstr.


GrayRachnid:
> I dunno when I see "It requires 20 strength to wield" on examine I would be very very confused so I don't like that. It'd be better just to remove the wielding modifier entirely.

As in, wielding should not halve the required str?

> Most two handed weapons do shit damage without wielding anyways.

I'll adjust that.

NotGoldenglow:
This seems like a bad idea. Holding a weapon you do not meet the minstr for in the first place with just one hand drops it to the floor. You are essentially bandaid fixing one thing while making it impossible to JUST hold a weapon without wielding it.

As it stands, they're ineffective onehanded. That is the point! This fixes nothing except take more steps to accomplish something already being handled.

anything86:
> This does not happen. The only thing using a weapon beyond your strength does is reduce its damage by 70%

Unless what you mean is that your PR removes that effect, then yes, it does happen. I recall playing Royal Champion knight and having to quickly 2-hand my double-headed greataxe whenever I pulled it out, or else my teeny dainty hands would drop it clattering to the floor. You're gonna have to go find where that happens, remove it, and ideally add some feedback when hitting something with a weapon you lack the strength to use (so that people aren't using weapons they can't hold and dealing 30% damage without noticing).

I'll also say that I feel this is only relevant as a balance consideration for a couple of weapons (minotaur greataxe, ogre weapons, tetsubo) and for everything else it feels unnecessary.

GrayRachnid:
> > This does not happen. The only thing using a weapon beyond your strength does is reduce its damage by 70%
> 
> Unless what you mean is that your PR removes that effect, then yes, it does happen. I recall playing Royal Champion knight and having to quickly 2-hand my double-headed greataxe whenever I pulled it out, or else my teeny dainty hands would drop it clattering to the floor. You're gonna have to go find where that happens, remove it, and ideally add some feedback when hitting something with a weapon you lack the strength to use (so that people aren't using weapons they can't hold and dealing 30% damage without noticing).
> 
> I'll also say that I feel this is only relevant as a balance consideration for a couple of weapons (minotaur greataxe, ogre weapons, tetsubo) and for everything else it feels unnecessary.

I'll double check if it does or doesn't happen but it didn't happen to me at all during testing. It also does give feedback.

For now, can you provide me a list of wield able weapons where you'd consider this an issue that should be resolved?

PR #1698, opened 2025-12-10
weakens skeletons slightly as asked

chemistrymain3:
## About The Pull Request

im kinda inclined to agree 

they now get slightyl worse armor in the case of the spammable soldier guy (chainmail, coif, half-plate instead of fullplate)
their stats are slightly worse
easydismember is back
i gave sapper some qol
they specialize in the weapon they pick (accept no substitutes)
## Testing Evidence

works

### DIFF
--- a/code/modules/jobs/job_types/roguetown/other/lich_skeleton.dm
+++ b/code/modules/jobs/job_types/roguetown/other/lich_skeleton.dm
@@ -69,40 +69,40 @@ LICH SKELETONS
 		if("Gladius")
 			beltr = /obj/item/rogueweapon/sword/iron/short/gladius/pagladius
 			backr = /obj/item/rogueweapon/shield/wood
-			H.adjust_skillrank(/datum/skill/combat/swords, 4, TRUE)
+			H.adjust_skillrank_up_to(/datum/skill/combat/swords, 4, TRUE)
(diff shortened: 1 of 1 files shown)


Lioneliuss:
what you mightve just did is made legendary skill skeletons (if this gets merged) if the order of things is correct.. they could get legendary by picking a virtue that gives them 2 skill in a weapon and then specializing in it adds 4 more. You should use H.adjust_skillrank_up_to instead of just H.adjust_skillrank just in case that is true

chemistrymain3:
> Cosigned, skeletons getting pretty much full steel kit is pretty strong.
> 
> I will say that the sandals + easy dismember addition means that it's Free Real Estate for any guard with a bardiche/cutting polearm/cutting weapon with long trait to just aim foot and remove skeleton legs.

this requires guards to think so they'll be fine

benj8560:
> Cosigned, skeletons getting pretty much full steel kit is pretty strong.
> 
> I will say that the sandals + easy dismember addition means that it's Free Real Estate for any guard with a bardiche/cutting polearm/cutting weapon with long trait to just aim foot and remove skeleton legs.

I'm roughly 70% sure that you can not dismember via attacking feet/hands so should be double fine.

PR #1708, opened 2025-12-11
Ties wretch stat bonus to outlaw status

UntoldTactics:
## About The Pull Request
High effort balancejak PR. (Totally)
Original PR changed.
Wretches will only get their omnistat +1 bonus if they have chosen to have their face known (they're tagged as an outlaw)

## Testing Evidence
image

## Why It's Good For The Game
The game is currently in a state where wretches frequently gang up together into powerful groups of combatants.
If they're gonna do this and get their stat bonus, it should be made more clear that they are valid from the start, I suppose.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/wretch.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/wretch.dm
@@ -84,6 +84,13 @@
 		if (!my_crime)
 			my_crime = "crimes against the Crown"
 		add_bounty(H.real_name, bounty_total, FALSE, my_crime, bounty_poster)
+		H.change_stat("strength", 1)
+		H.change_stat("perception", 1)
+		H.change_stat("intelligence", 1)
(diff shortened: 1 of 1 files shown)


pyrzal:
Yeah do it

saltprjak:
long time coming, could use their descriptors being added to their bounty log alongside their name too - as it stands you're much likelier to get ctrl+f'd with descriptors rather than be asked to stop and take your mask off

Fatherless1Xx:
It is absolutely wild that people want to shape balance around groups that metacomm instead of simply punishing metacomming. Anyone who plays a wretch without their ooc buddies is going to get absolutely wrecked by this.

Plum-but-gayer:
As a dedicated member of the wretch killblob I think this is fair. Hopefully cuts down on wretches who just kind of kick it in town, also. Which is fine, but like... kill somebody, why don't you? Live a little!

PR #1709, opened 2025-12-11
dolls/golems unable to be stigmata

Ephemeralis:
## About The Pull Request

The races that do not bleed shouldn't be doing the horrendous lux-sharing blood-twisting thing.

## Testing Evidence

one-liner

## Why It's Good For The Game

I'm getting so tired of chasing this sort of thing around.


ShadowDeath6:
at this point there should probably just be a “user needs to have blood to cast this” variable for spells

Ephemeralis:
implying that we actually use gating variables on half of our spells anywhere (range is set but not checked for like half the game's spells)

PR #1711, opened 2025-12-11
Fixes A Huge Naledian Oversight

ophaq:
## About The Pull Request
**- TM first, subject to change on feedback!**

Currently, all Naledian mercenary subclasses got the Hierophant's +15 spell points and T3 magic. HOLY SHIT! **This fixes that.**
- Pontifex & Vizier now get 6 spell points each.
- Pontifex gets T1 spells, for utility. If I recall right, they always had T1 for utility stuff... maybe I'm wrong? They had apprentice arcane regardless. They might have gotten 15 spell points in the past with T1. Since it is dropped to 6, maybe they could get T2 too but I want to know what people think on that.
- Vizier gets T2 spells, so they are not completely useless in combat or can grab a utility spell like recall. Currently they suck other than being pretty decent in self-buffing and using a polearm in combat. This is probably a bad idea, though... I want to know people's thoughts. It's a niche class that rarely got hired because they weren't combat viable (but were godly healers and a team supporter).
- Vizier also gets apprentice arcane and gets the empath + nostink traits because it made sense to me. They are constantly healing, and most players who play vizier will use their surgery kit and operate on dead bodies.
- EDIT: Vizier has 8 points and journeyman polearms, up from 6 and apprentice.
## Testing Evidence
EDIT: Just letting you know I put T2 for the Vizier, it might go back to T1, depending on feedback. Otherwise the rest is mostly accurate.
Naledifix

## Why It's Good For The Game
I don't think rendering one of the best mage classes rivaling the court magos useless should be a thing. Pontifex can currently be a kung fu fireball slinger... and vizier is well, healer fireball slinger. I took the liberty of adding some traits that make sense to the vizier while I was at it.

**You balance guys and naledian mains need to let me know if giving them each 6 spell points with Vizier having T2 should be OK, otherwise it goes back to T1.**

### DIFF
--- a/code/modules/jobs/job_types/roguetown/mercenaries/classes/warscholar.dm
+++ b/code/modules/jobs/job_types/roguetown/mercenaries/classes/warscholar.dm
@@ -205,7 +205,7 @@
 		STATKEY_END = 2,
 	)
 
-	subclass_spellpoints = 6
+	subclass_spellpoints = 8
 
 	subclass_skills = list(
 		/datum/skill/combat/wrestling = SKILL_LEVEL_APPRENTICE,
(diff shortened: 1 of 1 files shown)


CaydenT:
Yeah, agree that Pontifex is very strong. The funny thing with them is with their current number of points they can just buy _every_ tier 1 spell and flood their command bar with stuff they probably won't even use.

Vizier with jman polearm would be a fair trade for some of their points, for sure!

CaydenT:
Oh, Novice arcane skill if we're open to it, to let them get the most out of weapon enchant spell or make their own spellbook. I think almost every class in the game gets some amount of arcane skill if they are able to cast whatsoever, it may be an oversight that they have trait but no skill in arcane

CaydenT:
Ah! My mistake, I misread that. Excellent work! A nerf but also a good jump in QoL for them.

Elli-Skala:
AP finding the issue gave them a preset list of spells for both

PR #1734, opened 2025-12-14
Crimson Curse Partie Deux

EvenInDeathIStillServe:
## About The Pull Request

readds crimson curse properly
they do not get access to Any of the clan stuff. no menu no nothing. all they get is minor bloodheal and disguising.
they get a limited nopain that gulps up their vitae in exchange for letting them grit through pain. 250 vitae for every save. once they're below 500 vitae their no pain goes away
they do not get set on fire in the sun, instead getting the sunspurned debuff

disabled for church and inquisition roles. garrison and nobles can have it since it's very much not an ultimate gamer virtue after being neutered so thoroughly.

## Testing Evidence
image
image

## Why It's Good For The Game

vampyyy

### DIFF
--- a/code/modules/jobs/job_types/roguetown/church/druid.dm
+++ b/code/modules/jobs/job_types/roguetown/church/druid.dm
@@ -28,6 +28,7 @@
 	virtue_restrictions = list(
 		/datum/virtue/utility/noble,
 		/datum/virtue/utility/blueblooded,
+		/datum/virtue/combat/crimson_curse,
 	)
 
 	job_traits = list(TRAIT_SEEDKNOW, TRAIT_OUTDOORSMAN, TRAIT_RITUALIST, TRAIT_CLERGY)
--- a/code/modules/jobs/job_types/roguetown/church/monk.dm
+++ b/code/modules/jobs/job_types/roguetown/church/monk.dm
@@ -23,6 +23,7 @@
 	virtue_restrictions = list(
 		/datum/virtue/utility/noble,
 		/datum/virtue/utility/blueblooded,
+		/datum/virtue/combat/crimson_curse,
 	)
 
 	job_traits = list(TRAIT_RITUALIST, TRAIT_GRAVEROBBER, TRAIT_CLERGY)
--- a/code/modules/jobs/job_types/roguetown/church/templar.dm
+++ b/code/modules/jobs/job_types/roguetown/church/templar.dm
@@ -25,13 +25,13 @@
 	job_subclasses = list(
 		/datum/advclass/templar/monk,
 		/datum/advclass/templar/crusader,
-		/datum/virtue/combat/crimson_curse,
 	)
 
 	//No nobility for you, being a member of the clergy means you gave UP your nobility. It says this in many of the church tutorial texts.
 	virtue_restrictions = list(
 		/datum/virtue/utility/noble,
 		/datum/virtue/utility/blueblooded,
+		/datum/virtue/combat/crimson_curse,
 	)
 
 /datum/job/roguetown/templar/after_spawn(mob/living/L, mob/M, latejoin = TRUE)


EvenInDeathIStillServe:
> will they keep transfix at least?

yeah they still got it

Ephemeralis:
even conditional nopain is still like, way more powerful than a virtue aught to be. I think you should probably just drop the nopain stuff entirely. They're already getting strongbite, notox, etc as well, which is a *lot*.

EvenInDeathIStillServe:
> even conditional nopain is still like, way more powerful than a virtue aught to be. I think you should probably just drop the nopain stuff entirely. They're already getting strongbite, notox, etc as well, which is a _lot_.

The intent is you get fragged if found out. That's the balancing.. unless that's not happening?

kiwedespars:
do they still get infinite blue and immunity to bloodloss, if so the combat classes probably shouldn't have access to it

bloodheal itself is really strong compared to an aheal because it's just constant combat sustain

and also people usually don't get fragged if they get found out. garrison players literally find out and then proceed to not say anything because the vampire is on their side.

envoyin:
> even conditional nopain is still like, way more powerful than a virtue aught to be. I think you should probably just drop the nopain stuff entirely. They're already getting strongbite, notox, etc as well, which is a _lot_.

Is notox actually that good? The nopain goes away real quick depending on the fight you're having, too.

Plum-but-gayer:
we should frag harder 👍 create your inquisition validhunter today

NuIIpunkt:
> I think this is seriously broken in more than one way.
> 
> Current vampire mechanics are just way too powerful for this to be a virtue option. It clearly stomps on every other choice. In exchange for silver weakness and a pretty easy game of hide and seek, you get:
> 
>     * immunity to pain
> 
>     * immunity to bleeding out
> 
>     * immunity to choking or drowning
> 
>     * infinite energy, allowing constant activity
> 
>     * fast-healing powers
> 
>     * transfix which can knock people unconscious
> 
>     * bites turned to pierce damage (they have perfect accuracy and never miss, making it incredibly easy to get people's arteries)
> 
>     * the ability to replace your own limbs without surgery
> 
>     * inherently become allied with 2 different antagonists (wretch vampire, vampire lord) unless you behave aggressively or betray them
> 
>     * can't have your heart eaten by your opponent in the culling if you picked graggar, because you turn to ash when you die, which cucks them out of an earned victory (I did this to someone by accident yesterday)
> 
> 
> Furthermore, having so many people running around as vampires is not great for roleplay. We already have issues with too many sparkledog protagonist OCs and I think this makes that problem drastically worse.
> 
> As much as I get wanting to make it easier for you to roll it so you can do vampire RP, I don't think this is good for the game and it would really be better just to get rid of it.
> 
> If we want to have more vampire slots than the 2 wretch ones, we can: increase the number of vampire wretches possible OR add a thinblood coven people can opt into that can't convert people, with a max of a reasonable amount of players (maybe 6-8) OR add a vampire migrant wave etc. I think those would all be better options than having it be selectable by anyone.

The issue with having "too many vampires" fixxes itself after a week or two, look at how the virtue was when first introduced or ANY OTHER RACE (dullahans were a big outcry now).

People will play it alot in the first few days but then the wonder will die out...when have you last seen a dullahan? (the anti decap race)

EvenInDeathIStillServe:
> It very much feels like I am losing blue as a CC.

ok so basically i am being troled 💔

envoyin:
> > It very much feels like I am losing blue as a CC.
> 
> ok so basically i am being troled 💔

As far as I can tell, CC just has... more blue than normal people? I don't think it's by an insane amount, but it just seems like they have more blue and regain it somewhat faster... which is odd? I was still losing a lot of blue in fights.

PR #1737, opened 2025-12-14
Hearing Range Changes

Kroto-Ker:
## About The Pull Request
You can now **hear** people talk regularly from up to **9** tiles away, so the entire screen + 2 tiles extra. 
**With regular hearing** you only understand what people are saying within **7** tiles (anyone on screen.) 
**Exclaiming (!)** has a range of **13** tiles. At regular hearing you can understand what they’re saying from **10** tiles away.
**Yelling (!!)** has a range of **16** tiles. At regular hearing you can understand what they’re saying from **13** tiles away.

Screenshot 2025-12-09 120458

Elves, tabaxi, wild-kin, goblins, tabaxi, vulpkians, & lupians all have +1 hearing range. 

The Keen Ears trait also gives +1 hearing range and now shows you exactly how far away someone off-screen speaking is.

## Testing Evidence

image

## Why It's Good For The Game

Allows for traits that adjust hearing range. It also makes exclaiming / yelling range happen in a more understandable way. Previously, yelling with only one exclamation point (!) sent your message 17 tiles away, and yelling with two (!!) did nothing. 
Also makes for some extra spy gameplay with eavesdropping and hearing range.

### DIFF
--- a/code/game/say.dm
+++ b/code/game/say.dm
@@ -113,6 +113,7 @@ GLOBAL_LIST_INIT(freqtospan, list(
 			var/dist = get_dist(speakturf, sourceturf)
 			var/mob/living/M = src
 			var/hear_limit = 7 + M.extra_hearing_range
+			var/scramble = 60
 			var/yelling = say_test(raw_message)
 			if(yelling == "2")
 				hear_limit += 3
@@ -123,6 +124,7 @@ GLOBAL_LIST_INIT(freqtospan, list(
 			if(HAS_TRAIT(src, TRAIT_KEENEARS))
 				if(ishuman(speaker) && ishuman(src))
 					hear_limit += 1
+					scramble -= 30
 					numpart = "[dist]"
 					var/mob/living/carbon/human/HS = speaker
 					var/mob/living/carbon/human/HL = src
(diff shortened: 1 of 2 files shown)


GrayRachnid:
How does this affect multiz hearing?

If not at all, how hard would it be to separate multiz hearing from normal hearing?

GrayRachnid:
Instead of "yells something unintelligible" can you scramble words/letters instead?

Kroto-Ker:
> How does this affect multiz hearing?
> 
> If not at all, how hard would it be to separate multiz hearing from normal hearing?

The hearing radius for multi-z might be slightly reduced, but otherwise it should be about the same.

> Instead of "yells something unintelligible" can you scramble words/letters instead?

If this PR gets merged I wouldn't be opposed to you adding that feature, but I think 'something unintelligible' works well enough for what I want.

chaoko99:
https://i.imgur.com/puLeXC8.png the dir numbers are showing up in the chat

Kroto-Ker:
> https://i.imgur.com/puLeXC8.png the dir numbers are showing up in the chat



> also these markers show the voices and the whole point of keen ears is to be able to recognize voices at a distance.

These are both intentional behavior.

Kroto-Ker:
> The text size on distant stuff is nightmarishly small if you have to play on larger font sizes for whatever reason.

Increased size of text in latest commit to be the regular smallyell. Should be more readable.

Elli-Skala:
Walking through town its really iffy seeing half my chat being "x says something unintelligible"

What about giving it the ss13 tcoms black out effect?  Where the further and less intelligible someone in the more of their text replaced with random symbols

Githubaccount222:
Keen ears seems to be working much better now!

PR #1743, opened 2025-12-15
Firecrit now kills you

petwir1:
## About The Pull Request
Adds OxyLoss back to being in firecrit. You will no longer have to throat chop or bleed people to death to kill them as a wizard.

I put it under breathing code because it automatically filters out those who do not breathe. They will not be affected by this.

I intend for this to be some kind of band aid for fire damage until someone reworks it. If there's things that should be changed let me know, I'm extremely new to BYOND coding so I'd appreciate any constructive criticism.

## Testing Evidence
Here's it on a breathing person:

https://github.com/user-attachments/assets/54930353-6c68-456b-9358-df233f3d31ca

Here's it on a lich:

https://github.com/user-attachments/assets/0416d207-5cbc-40a8-a03d-c97aa75e1e9b

## Why It's Good For The Game
Fire should be deadly, this makes playing a class that benefits from burn damage not need to veer out of their playstyle to kill someone. Being a wizard that has to stab someone to kill them after they've been fireballed is stupid.

image

### DIFF
--- a/code/modules/mob/living/carbon/life.dm
+++ b/code/modules/mob/living/carbon/life.dm
@@ -114,7 +114,7 @@
 		adjustOxyLoss(10)
 	var/list/lethal_zones = list(BODY_ZONE_HEAD, BODY_ZONE_CHEST);
 	if(health <= HEALTH_THRESHOLD_FULLCRIT && getFireLoss(lethal_zones) >= (100 / FIRE_HARDCRIT_DIVISOR) * getMaxHealth(lethal_zones))
-		adjustOxyLoss(10)
+		adjustOxyLoss(4)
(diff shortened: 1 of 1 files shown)


DiscordWizard:
I thought enough fire damage, like twice your base health, already killed you outright as it does in base ss13. Does being in fire crit not keep dealing burn damage?

Wesreidau:
A little walking around got me still thinking.

How about if they are burning in firecrit, oxyloss. If they aren't burning in firecrit, no oxyloss. We presume someone burning in firecrit cannot breath due to the flames, but someone not burning with severe burn damage still can.

You stated this was a temporary fix, and there's nothing more permanent than a temporary fix.

PR #1748, opened 2025-12-15
Baothan "Content": Succ Those Stats

EvenInDeathIStillServe:
## About The Pull Request

adds the Baotha's Banquet buff, and Vitality Drained debuff

baothans can statdrain ppl if they cum into or get cummed in while they have combat mode on.
the buff and debuff both stack 3 times, give and take 1 of each stat per tier save for fortune. they both last for 15 minutes
the buff only stacks once for every victim. its duration can be refreshed using any partner.
the debuff stacks every time the victim is statdrained, and refreshes likewise.

also gave baothans some sex pain resistance. they won't get damaged as easily. you can be a lot rougher with them.

## Testing Evidence

image

image

## Why It's Good For The Game

leveldrain funni hehe


Seashine:
I wonder if this stacks with WW "mechanics". 0/1 int possible on the victim sounds hilarious. Baothan WW meta build...

Elli-Skala:
While funny, I cant imagine many things good things will come from  stats like this. I think it could make a interesting mirical though

CormanQuewye:
> As funny as this is. This is will probably never be used... I think it should probably last longer like a day. (30 minutes)
> 
> Practically speaking, usually by the time you get to actually use the stats... It may already be over?

30 minutes, refreshing on each new partner, that'd let bandits just pick up a prisoner, get buffed, come back and grab another prisoner, repeat. I don't think it would be amazingly good or anything even at 30 minutes but I can see niche situations where you could just have a band of buffed up gods that would be at +2 stats for ~1 hour. The 15 minute timer is a little short, but I think it's one of those cases where playing it safe is for the best.

It's a cool fluff ability and doesn't really need to be more than that I don't think it should be good enough that we start theory crafting on how to hyper-optimise our sexcon gameplay to reach max combat effectiveness.

zeroisthebiggay:
> It's a cool fluff ability and doesn't really need to be more than that I don't think it should be good enough that we start theory crafting on how to hyper-optimise our sexcon gameplay to reach max combat effectiveness.

on the contrary this would be really funny and i like the idea of baothans having more incentive to just be demons of pleasure and engaging in plapbeast behavior to minmax stats

Aracino:
This is awesome but needs to not work on vampires.

kittysmooch:
this is rly funny but the idea that bandits and knights are now mechanically inclined to noncon each other is uhhhhhhhh it's uhhhhhhhhhhhhhhhhhm

chaoko99:
Doesn't seem to work with riding anal either. I don't seem to get what's up here.

chaoko99:
yeah i literally just sucked someones' cock on combat mode and didn't get the buff when they came in my mouth, despite already having a level of it from someone else, which I had acquired through sucking their cock from the SAME tile. I'm not sure if there's some correlation there.

edit: nope, rode him and it didn't work either.

chaoko99:
they had a knotted penis if that's a factor. The first did not.

PR #1752, opened 2025-12-15
Makes Wall SCOMs destructible

GrayRachnid:
## About The Pull Request
They cost 1 iron ingot and a rat to make.
Same integrity as an anvil (which also costs 1 iron ingot)

## Testing Evidence
image

## Why It's Good For The Game
There is no reason for a machine that can be created to not be destructible.


DeteledSawI:
so if you destroy Scom #72, will it still be listed in the scom list or is the number PERMENANTLY DESTROYED when the building is? like would the engineer be able to set it back to #72 so that people could call the castle, or would that scom be on a different number for the rest of the game?

PR #1758, opened 2025-12-16
Make Whole: Repairing rework (it's way less awful for everybody now)

Ephemeralis:
## About The Pull Request

Pretty big set of changes here, all centered around making repairs immensely less CBT for 90% of the server:

- Repairing BROKEN objects via needle/hammer below Journeyman in the relevant skill now counts as a "field repair".
- Field repairs only restore an item to 60% of its max durability, and field-repaired items cannot be repaired past 60% by unskilled (read: below Journeyman) users.
   - **PLEASE READ THE FOLLOWING**: field repair state is only flagged on an item if you repair it from it being BROKEN. If you are diligent with your repairs and don't let your items get broken, they can be repaired to 100% by anyone, not just skilled crafters.
- Skilled users (read: at or above Journeyman) will fix field-repaired objects back to full working order naturally when repairing, allowing them to be repaired to 100% max durability as normal.
- Weaponsmithing/armorsmithing repairs now function at the old squire level minimum for EVERYBODY. This means you can no longer fail strikes - you'll at least get a little bit of repairing for every hit you do.
- Needle/sewing repairing is now significantly faster at a base level, taking only 4 seconds instead of 6 at zero skill.
- Needle/sewing repairing cannot cause damage on failure anymore, instead reducing the amount of repairing you perform by 50% for that action.
- Successfully fixing an item from broken with a needle (either as a field repair or a full repair) consumes 1 use of thread from the needle.
- The mending spell fixes field repairs as a skilled craftsman would if the mage is at or above Journeyman arcane magic skill.
   - This is intended to give the garrison the means to seek out mages from the tower to help repair their gear in the absence of smiths or other emergency scenarios.
   - Toilers and liches also can provide valuable frontline emergency mending services to the Skelelon Hordes this way, too.
- Failed Squire is removed as a virtue. Squire repair is now exclusive to squires and squire-themed subclasses.
- Squire repair makes all repairs consider your character as if they're journeyman or higher in all respective repair skills for the purposes of performing repairs - this means they can restore field repairs to full integrity, and so on.

## Testing Evidence

dreamseeker_b8V5rg1ej5

dreamseeker_bZdT9pqTRv

image

## Why It's Good For The Game

Given that integrity serves as more or less a second health bar, repairing your gear is essential for any combat role to remain able to participate in the game. 

The current implementation of repairing essentially forces people to either stare at a table for 10 minutes hearing CLANG CLANG CLANG or wet squishing noises as they repair their gear, or join the abject hell that is the Smith Line. Nobody enjoys doing this. Adventurers get especially fucked over by this if they're out of town for any serious length of time, doubly so if they're a parry or shield build.

A 40% max durability reduction on field-repaired items and quicker field repairs introduces a real and interesting decision to make in longer fights - do you repair your fucked-up gear and chance it at nearly half strength, or do you retreat back to safety to recover? Craftsmen and mages retain valuable roles as being the defacto people to visit if you want your sweet-ass gear repaired to full nick - which you will, because 40% reduced durability is nothing to sneeze at.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/gatemaster.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/gatemaster.dm
@@ -22,9 +22,6 @@
 	round_contrib_points = 3
 	social_rank = SOCIAL_RANK_YEOMAN
 
-	virtue_restrictions = list(
-		/datum/virtue/utility/failed_squire,
-	)
 	cmode_music = 'sound/music/combat_guard.ogg'
 
 	job_traits = list(TRAIT_GUARDSMAN, TRAIT_STEELHEARTED, TRAIT_MEDIUMARMOR)
--- a/code/modules/jobs/job_types/roguetown/garrison/manorguard.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/manorguard.dm
@@ -28,7 +28,6 @@
 	social_rank = SOCIAL_RANK_YEOMAN
 
 	virtue_restrictions = list(
-		/datum/virtue/utility/failed_squire,
 		/datum/virtue/utility/blacksmith, // we don't want you repairing your stuff in combat, sorry...
 	)
 	job_traits = list(TRAIT_GUARDSMAN, TRAIT_STEELHEARTED)
--- a/code/modules/jobs/job_types/roguetown/garrison/sergeant.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/sergeant.dm
@@ -21,7 +21,6 @@
 	social_rank = SOCIAL_RANK_YEOMAN
 
 	virtue_restrictions = list(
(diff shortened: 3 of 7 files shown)


brewonloop:
While good for townies this does screw over antags quite a bit, bandits and wretches rarely ever have anyone with any smithing skills and tend to break their armor alot from fights they get into. So this will often just permanently lock them to 60% durability since they can't ask for help from town.

Otherwise pretty good, though there really does need to be a way for bandits or wretches to get full repairs done.

Ephemeralis:
I could probably just make repairing check to see if you're in the centcomm z-level (which is where wretch church, VL manor, bandit camp) is, and if so, you repair there as if you're a skilled craftsman.

annikaRU:
I personally think this is an amazing way to approach this problem, props for that

PR #1763, opened 2025-12-17
Noc blindness removal

Ephemeralis:
## About The Pull Request

Bad spell. Has to go. Can't be balanced - no degree of chaincasting soft CC this oppressive can be reasonably balanced in a way that makes it worth having and not also worth spamming.

## Testing Evidence

one-liner

## Why It's Good For The Game

Watched two fights today where someone spent pretty much a full minute or more just chain-blinded by a random passing-by acolyte with no prelude or counterplay available to them. We can do better than this.


chaoko99:
this was recently made to scale the length with your miracle skill IIRC... Which might be related.

CaydenT:
Nobody thought to add a buff to the target that makes them immune from chain castings for a while????

Ephemeralis:
> Nobody thought to add a buff to the target that makes them immune from chain castings for a while????

"no degree of chaincasting soft CC this oppressive can be reasonably balanced in a way that makes it worth having and not also worth spamming."

PR #1771, opened 2025-12-17
Changes the craftable Qaba to be generic and armorless, adds to loadout

pyrzal:
## About The Pull Request

Powergamers who play roles that have nothing to do with Naledi are using the Naledi Pontifex Qaba to armorstack because it's an armored cloak-slot item. This PR makes them uncraftable so that only the Naledi Pontifex gets them, as it should be.

EDIT: Adds a craftable unarmored version without the golden Naledi psycrosses instead, and makes it available in the loadout. 
Also removes the Heirophant robe crafting recipe for the same reason. I'm too lazy to make an unarmored variant. Sorry.

## Testing Evidence

image

## Why It's Good For The Game

Dilution of regional fluff and flavor for the sake of frag-hungry gamers is always bad.

### DIFF
--- a/modular_azurepeak/code/datums/loadout.dm
+++ b/modular_azurepeak/code/datums/loadout.dm
@@ -35,10 +35,6 @@ GLOBAL_LIST_EMPTY(loadout_items)
 	name = "Hand Mirror"
 	path = /obj/item/handmirror
 
-/datum/loadout_item/short_satchel
-	name = "Short Satchel"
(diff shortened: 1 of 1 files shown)


ophaq:
Noooo not my qaba! You have a fair point but muh drip! Oh well.

pyrzal:
Not interested in any of the feedback so far other than that there should be a cosmetic armorless variant, which I'm going to add as a generic non-Naledi Qaba.

PR #1773, opened 2025-12-17
Makes the steel chain skirt medium armor, plate tassets heavy armor

pyrzal:
## About The Pull Request

Changes steel chain skirts to medium armor and grants them groin coverage, making them in-line with the steel chausses since they have the same armor values. Does the same with steel plate tassets.

## Testing Evidence

image

## Why It's Good For The Game

Light-tier armor should not be getting medium-tier protection. Medium-tier armor should not be getting heavy-tier protection. This makes them cosmetic options, and opens up the option of a future PR that gives knights pants or skirts depending on clothing preference.

### DIFF
--- a/code/modules/clothing/rogueclothes/pants.dm
+++ b/code/modules/clothing/rogueclothes/pants.dm
@@ -465,7 +465,6 @@
 	desc = "A knee-length mail skirt, warding cuts against the thighs without slowing the feet."
 	icon_state = "chain_skirt"
 	item_state = "chain_skirt"
-	armor_class = ARMOR_CLASS_MEDIUM
 	grid_height = 32
 	grid_width = 64
 
@@ -475,7 +474,6 @@
 	gender = PLURAL
 	icon_state = "plate_skirt"
 	item_state = "plate_skirt"
-	armor_class = ARMOR_CLASS_HEAVY
 	grid_height = 64
 	grid_width = 64
 


RasbeRissa:
They don't have the same values. Chausses covers Groin and Legs. Skirts and Tassets only covers Legs. Same thing with the corselet and bikini armors; they drop a weight class but don't cover the stomach.

pyrzal:
> If you're going to make it identical to the chain chausses, it should be made fully identical. Have it cover the groin, and it can be a cosmetic side option to the chain chausses like the Kilts are.

Sure. It will change nothing because nobody aims groin, so I'm fine with that.

pyrzal:
> I don't quite agree, the reason they were made in Azure in the first place was for there to be a slash-covering alternative to light armor, who did not (and will not if this is merged) have **any** way to acquire at leash slash-resistant leg armor.
> 
> Drop the integrity on it if you must, but leave the armor class. It's only best in slot because everyone defaults to swords and cutting weapons for some reason. Pick up a mace, try it some day.

This was not an Azure change. The change from medium to light armor chainskirts was originally made on Blackmoor with Neo_Aurelian's approval: https://github.com/BlackmoorHold/Blackmoor-hold/pull/107
It had the following justification:
> Following a concept of 'reduced coverage for higher protection', shifted skirts to be chainmail leg armor with the groin as a weak spot. Allows light and medium armor users to 'step up' their leg armor without losing dodge or requiring heavy armor training, but simultaneously creates a new risk of groin strikes. This in turn shifts attacks to the chest, creating a shift in the combat meta away from limb strikes.

I don't find that logic sufficient to justify it. I don't see why dodge builds should be able to 'step up' their armor when **the entire balancing point of dodge is that you sacrifice your armor potential in exchange for extremely high dodge chances with no cooldown.** "Reduced coverage for higher protection" falls apart when you're barely getting hit in that zone to begin with, and I take issue with dropping the integrity of a steel armor to fit a niche use-case like this. **I've been observing players who use this armor item and I've yet to see anyone aim groin against them.** In practice, it's just a pure buff, and I think the resistance to this PR is coming from people who don't want to lose that buff.

I don't think there's a way to balance the integrity that would make this armor piece sensible and fair as light armor. There's no reason steel gear should be flimsy for the sake of dodge builds — it's already unbelievable enough that you can effectively dodge around in a chain skirt made of heavy steel, which really shouldn't be the case. Light armored classes should be using light armor, and steel is anything but. Besides, when you're consistently hitting 90% dodge chance, you don't need a unique slash-resistant armor for your leg slot. Dodge builds are competitive enough, especially with the lack of wallsplats. They really don't need the help.

I play dodge builds too, but I don't want to encourage niche and unintuitive strategies that unfairly benefit players who dive through the code to find obscure power-boosts like this. **The fact that dodge-build characters are actually going out of their way to grind out smithing skill specifically to make this armor _and only this armor_ tells me that something is wrong.**

**TLDR:** **It's steel armor. Steel is medium.** Let it be what it was meant to be. This PR won't affect 99% of the people who use dodge builds — only the ones who powergame this particular item due to their obscure knowledge of the codebase. Let it happen. Your build will be fine.

RasbeRissa:
Chiming in. I have one light armor character that dodges, and I simply can not be bothered with waiting in the smith line or grinding armor smithing to make a skirt. I also don't like the clink noises it makes on my light armor character, so I don't use it.

I'm not opposed to this PR because I "don't want to lose the buff." I'm opposed to the PR because it sets a precedent that nerfing things because people don't bother to use the tools in their kit (a skill issue, as I said before) is okay. It's not like the counterplay is esoteric "You need to play a high perception mage with Testicular Torsion memorized" or anything either. It's "Hey I should use my Assess to see their layers. Oh they have no coverage on their groin. I should hit the 2 key on my numpad to exploit that."

And don't tell me Assess isn't viable. Almost every gamer is running Intellectual because of the +1 Int bonus. Even if Assess wasn't in the game you can just take the skirt and shift click it to see it doesn't cover groin.

PR #1775, opened 2025-12-17
It's All Ogre: Ogre slot reduction & PQ requirement increase

Ephemeralis:
## About The Pull Request

Ogres are now capped to 2 slots and require 35 PQ minimum to play.

Stop-gap change while we decide on whether we still want them around at all.

## Testing Evidence

two-liner var adjustment

## Why It's Good For The Game

Done and discussed to death in Discord over the past few days. The tl;dr is that Ogres are simply too powerful mechanically (people have discovered that a character with 11 morbillion strength and huge force weapons can crush armor and delimb people off a single bait, unsurprisingly) and too flat in available character expression to be a net positive for the server.

Reminder: just because you paid for a code bounty does not mean we're obligated to keep the result.


ConstantineII:
Good PR

PR #1781, opened 2025-12-18
Sun Elves 

pyrzal:
## About The Pull Request

Adds the Sun Elf subrace, a species of elf hailing from Raneshen. They get the Silver Blessed and Deadite Immunity trait, making it impossible for them to be converted into deadites, werewolves, or vampires.

**NOTE FOR MERGING:** This PR requires `ROUNDSTART_RACES elfc` to be added to game_options.txt in the server config in order to be enabled. Otherwise it will do absolutely nothing.

## Testing Evidence

image
image

image

## Why It's Good For The Game

It's my fetish.

### DIFF
--- a/modular_azurepeak/virtues/origin.dm
+++ b/modular_azurepeak/virtues/origin.dm
@@ -24,6 +24,7 @@
 	desc = "I originate from the dark boreal woods of Grenzelhoft, the birthplace of humanity. The Empire has a long history of conflict with Otava. Long live emperor Kovel II!<br>"
 	restricted = TRUE
 	races = list(/datum/species/elf/dark,
+				/datum/species/elf/sun,
 				/datum/species/halforc,
 				/datum/species/tabaxi,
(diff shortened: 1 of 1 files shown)


pyrzal:
Added immunity to deadite, werewolf, and vampire conversion.

pyrzal:
> Not that anyone asked for my opinion - but it does seem weird to add a race of people that is in essence a demi-god lore wise.

They're not demigods. They just think they are. Their empire collapsed, they're obviously not demigods.

Also, we already have Aasimar, which are ACTUALLY literally demigods.

pyrzal:
> Does Deadite immunity not mean if they die somewhere out of the way, they are just not getting revived? I imagine it prevents them from rotting. Not a problem in my view, I just think it's funny anybody would see any of those immunities as a perk.

It prevents them from raising as zombies. It doesn't directly interact with revives, but you can argue that it makes it harder for their bodies to be found, yes.

PR #1782, opened 2025-12-18
The Cloak Stays On During Hunt Night: Miscellaneous druid-related fixes/QoL improvements

Ephemeralis:
## About The Pull Request

List of minor changes:
- cloaks and a character's underwear remain stowed away on their saved human mob when wildshaping, meaning that they'll come out of wildshape afterwards wearing said cloak and underwear (+ socks)
- druid wildshapes no longer spray gibs everywhere upon shifting into them
- socks/stockings are now wclass_tiny items like underwear and thus can be stored
- wicker cloak now has cloak storage
- druids will shift with their neck and ring slot items, negating the need to pick them up and re-equip them after the shift
- added a preference toggle w/ OOC verb to choose whether your wildshape is "shape (real name)" or just "shape". So with it off, "volf (real name)" becomes "volf" instead.
- adds meaty squishy morphing sfx to transforming in and out of wildshape, both for sovl and also to tell people that something is actually going on. sfx from pixabay
- adds half-elves to gronn origin (half-elves should arguably be wherever humans are)

## Testing Evidence

relatively simple changes

## Why It's Good For The Game

Wildshape sort of forces druid into the 'I must be naked all the time' thing. This now lets them decide just how naked that naked really is, and gives them a whopping whole 4 extra slots of guaranteed storage to carry their valuables in outside of a neck pouch. If you still want a fully naked druid as before, just don't put underwear on your char sheet, and don't wear a cloak. Easy.

A druid shifting out of bear form into a person wearing *just* a cloak is also pretty kino, too.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/church/druid.dm
+++ b/code/modules/jobs/job_types/roguetown/church/druid.dm
@@ -30,7 +30,7 @@
 		/datum/virtue/utility/blueblooded,
 	)
 
-	job_traits = list(TRAIT_SEEDKNOW, TRAIT_OUTDOORSMAN, TRAIT_RITUALIST, TRAIT_CLERGY, TRAIT_WOODWALKER, TRAIT_DODGEEXPERT)
+	job_traits = list(TRAIT_SEEDKNOW, TRAIT_OUTDOORSMAN, TRAIT_RITUALIST, TRAIT_CLERGY, TRAIT_WOODWALKER)
 
 	advclass_cat_rolls = list(CTAG_DRUID = 2)
 	job_subclasses = list(


ophaq:
Oh nice, I wanted to fix this. This applies to both underwear and stockings, right? Currently if you wildshift you drop both of those and you transform back with them on I think. Also, stockings take up almost the entire satchel funnily enough.

Ephemeralis:
It only applies to underwear but I can probably do the same thing for stockings easy enough.

ophaq:
> It only applies to underwear but I can probably do the same thing for stockings easy enough.

Thanks! If it is not too much trouble can you reduce the inventory size of stockings as well to 1x1 like underwear is?

Ephemeralis:
- adds meaty squishy morphing sfx to transforming in and out of wildshape, both for sovl and also to tell people that something is actually going on. sfx from pixabay

Githubaccount222:
Can the druid robes have storage added, too?

PR #1783, opened 2025-12-18
Minotaur axe change

ConstantineII:
## About The Pull Request
Changes the drop for the minotaur, they get nothing in return now.

## Testing Evidence
I fought a minotaur and it didn't drop an axe.
Screenshot 2025-12-18 022940

## Why It's Good For The Game
This just causes balance issues and we've had many issues with it for awhile now.

### DIFF
--- a/modular_azurepeak/code/modules/mob/living/simple_animal/rogue/creacher/minotaur.dm
+++ b/modular_azurepeak/code/modules/mob/living/simple_animal/rogue/creacher/minotaur.dm
@@ -76,7 +76,7 @@
 	icon_dead = "MinotaurFem_dead"
 
 /mob/living/simple_animal/hostile/retaliate/rogue/minotaur/axe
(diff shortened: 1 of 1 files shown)


chaoko99:
this weapon was actually designed with the intent of it being strong and easily broken, but with the removal of shafts, it lost its super weak shaft.

pyrzal:
> I think the outright removal of the axe is another issue of treating symptons and not the main issue. The issue with it came in only after shafts were removed like how chaoko mentioned. I think the decrease of drop chance alone would suffice and doubt removing an item entirely over punishing powergaming individually won't help as much.

If it's good enough to use, then it's good enough for people to metarush it (bad player behavior we don't want). If it's not good enough to use compared to the effort that it takes to get it, nobody will really bother with it. And if it's not worth using, then it's not worth having minotaurs drop them.

PR #1795, opened 2025-12-19
Rosa Crown + Fancy Coat to Loadout

ophaq:
## About The Pull Request
Adds the following to loadout selectables:
- Rosa Crown
- Salvia Crown
- Fancy Coat

## Testing Evidence
Rosacrownloadout

## Why It's Good For The Game
I am pretty sure someone mentioned the rosa crown not being a loadout item, despite its popularity. The fancy coat is nice too. I also added the salvia crown on request. These are in the RW codebase's loadout list so I don't see why not here. Both are pretty easy to make regardless, and more drip is good.

### DIFF
--- a/modular_azurepeak/code/datums/loadout.dm
+++ b/modular_azurepeak/code/datums/loadout.dm
@@ -136,6 +136,10 @@ GLOBAL_LIST_EMPTY(loadout_items)
 	name = "Rosa Crown"
 	path = /obj/item/flowercrown/rosa
 
+/datum/loadout_item/salviacrown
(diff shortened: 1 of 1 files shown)


JonathanTheII:
Could the salvia crown be added too? It'd make sense, with the rosa one.

NuIIpunkt:
These are popular options and yes I love having them myself aswell but I have to disagree with adding more fancy clothes like the cloak to the loadout menu. I am still against most of the fancy dresses aswell as you should be forced to find a tailor imo. :heart:

PR #1806, opened 2025-12-20
Councillor Adjustments

JCarve:
## About The Pull Request

Brings Councillor in line with their Adventurer counterpart - Aristocrat.

## Testing Evidence

Just stats and skills, should work no problem.

## Why It's Good For The Game

While Councillors are a less important role they still have potential to influence a round through RP. 
I found it weird that an adventurer subclass gets better stat and skill spread than them when one is slot limited and the other is a free-form unlimited slot role.
Councillors will still get easily fragged by antagonists and pretty much every other combat role so this is more of a QoL.


RandomGooberCoder (approved):
Seems fine as it really brings them to ADV level. W/e, I'd say.

In the future - I'd suggest choosing better names for your PRs, aka not a vague adjustment but something like "Brings councillor stats and skill on part with aristocrat adv"

PR #1807, opened 2025-12-20
Salt PR 1: Pickpocketing Nerf

GrayRachnid:
### Pickpocketing Difficulty System Redone with Salt:
- Roll: 1d12 + (skill × 2) + (SPD / 6) vs Target Perception + Slot Difficulty
- Slot Difficulties: Belt +4, Back +6, Necklace +14, Ring +10
- SPD provides minor bonus - skill is the primary factor
- Novice thieves can attempt belts with ~17% chance at average stats (PER 10, no combat mode)
- Journeyman thieves have ~50% on belts, can start attempting back slots
- Expert thieves have reliable success on belts/backs, can attempt rings
- Master thieves can attempt neck, and have high success on all other slots
- Combat mode adds +6 to effective perception, making all attempts much harder
- Matthios worshippers roll with advantage, taking the higher of two rolls
- Slot difficulty modifiers (added to effective perception)

## Testing Evidence
I tested it but forgot to take screenshots. It compiles. I apologize.

## Why It's Good For The Game
Right now you can stack speed, have high athletics, some endurance, and no pickpocketing to guarantee a high enough chance to steal almost anything from anyone and then get away before they can notice you.
This changes it so that you need skill first, with speed as a secondary stat. Speed now factors in the speed of your do_after, rather than the default being 0.5 seconds. It scales from 5(10 spd) to 0.5 seconds(at 19 spd).
Zones you target are now harder to steal based on your skill level. The hips are the easiest, the back slot is next, rings are harder and the necklace is hardest (fuck gorget thieves midcombat).

### DIFF
--- a/code/game/objects/items/rogueweapons/intents/mmb/steal.dm
+++ b/code/game/objects/items/rogueweapons/intents/mmb/steal.dm
@@ -65,8 +65,8 @@
 
 		to_chat(user, span_notice("I try to steal from [target_human]..."))
 
-		// Pickpocketing delay scales with SPD: 10s at SPD 10, 1s at SPD 20 (deciseconds)
-		var/delay = clamp(round(100 - 10 * (user.STASPD - 10)), 10, 100) // 10 seconds at SPD 10, 9 seconds at SPD 11, ..., 1 second at SPD 20
+		// Pickpocketing delay scales with SPD: 5s at SPD 10, 0.5s at SPD 19+ (deciseconds)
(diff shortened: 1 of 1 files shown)


GrayRachnid:
> Have you considered making it require a brief do_after to pickpocket? (One that's only visible to the user)

Yes. The ability to display a do_after only to the user and no one else is not something that the codebase inherently supports as far as I could tell.

PR #1808, opened 2025-12-20
Salt PR 2: Removing the Swimming Skill

GrayRachnid:
## About The Pull Request
This removes the swimming skill entirely from the game.
Base swimming speed is now the same as master, applied to everyone.
Stamina drain is now based on athletics skill, with the same parameters that were had with swimming.
Skeletons are still slower in the water, but now they can't become faster by leveling up swimming. They are just slower.
The Abyssor swim trait functions the same.

## Testing Evidence
it compiles
image

## Why It's Good For The Game
All combat roles being forced to grind this skill or be locked out of a battlefield is bad game design.
The skill itself shouldn't exist in the way it does, because the only people who benefit from its existence are those who level it up to gain an upper hand on the other side.
It's an always damned if you don't scenario and results in the most boring gameplay I've ever seen in my life of everyone jumping into the water and using the fact that they spent the first 30 minutes of the game swimming against a current to their advantage.

Read my comments in the thread below for most of my thoughts that are better articulated.

This makes it so your roundstart athletics and whatever it is after is what influences your ability to last long in the water, while speed and endurance assist for speed and even more endurance.

### DIFF
--- a/code/game/turfs/open/water.dm
+++ b/code/game/turfs/open/water.dm
@@ -275,7 +275,7 @@
 
 /turf/open/water/get_slowdown(mob/user)
 	var/returned = slowdown
-	returned = max(returned - SKILL_LEVEL_JOURNEYMAN, 0) // Everyone has journeyman swimming now
+	returned = max(returned - SKILL_LEVEL_MASTER, 0) // Everyone has journeyman swimming now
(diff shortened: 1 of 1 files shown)


GrayRachnid:
Further evidence of this awfulness: 
image

GrayRachnid:
I still have a lot of problems with water, but i'm pulling myself out of balancing things that personally affected me.

PR #1819, opened 2025-12-22
Revival Debuff is Stronger, Lasts Longer, Duke is DNR (with a few exceptions)

anything86:
## About The Pull Request
Makes the Devitalized debuff from being revived last one hour, up from fifteen minutes. Penalizes constitution by a further -1.

Makes it so Dukes cannot be revived via lux or Anastasis. Cue lore about how Astrata forbids it to keep things stable and ensure mortal rulers do not rule for abnormally long. Some rarer revival sources will still work.

## Testing Evidence
~~One line change.~~
It compiles.

## Why It's Good For The Game
Makes revival more punishing and death less inconsequential, without actually depriving people of their ability to continue playing (and thus making them just Alt F4).

As it stands, all acolytes can revive, and all acolytes get wound heal, so bringing someone back to life is piss easy. There needs to be guard rails on it, or else it encourages very bad behavior on all sides of the playing field.

Duke assassinations tend to make for good plot and it is always sad when it gets screwed over by revivals.

### DIFF
--- a/code/datums/status_effects/rogue/debuff.dm
+++ b/code/datums/status_effects/rogue/debuff.dm
@@ -390,7 +390,7 @@
 	id = "revived"
 	alert_type = /atom/movable/screen/alert/status_effect/debuff/revived
 	effectedstats = list("strength" = -1, "perception" = -1, "intelligence" = -1, "endurance" = -1, "constitution" = -2, "speed" = -1, "fortune" = -1)
-	duration = 1.5 HOURS		//Punishing, meant to last for the rest of the round if it happened about halfway through.
+	duration = 1 HOURS		//Punishing, meant to last for the rest of the round if it happened about halfway through.
 
 /atom/movable/screen/alert/status_effect/debuff/revived
 	name = "Revival Sickness"


modular_azurepeak/code/modules/reagents/reagent_containers/rotcure.dm
```diff
@@ -7,7 +7,7 @@
 	description = "A putrid substance that appears to be in perpetual motion. It smells and looks of living-sludge."
 	color = "#034212"
 	taste_description = "living sludge"
-	overdose_threshold = 10	//Stops people from downing too much of it.
+	overdose_threshold = 21	//Stops people from downing too much of it.
```
zeroisthebiggay: what is this arbitrary number
a phial is `volume = 27` so if you drink a full phial you'll overdose but you need twenty to cure it
cure threshold should be `25` and overdose should be `30`

zeroisthebiggay: also undocumented change lmao

brewonloop:
Nobility as a whole should honestly be DNR, you should not be rushing people with valuable hostages.

PR #1824, opened 2025-12-22
Horny ogling

RandomGooberCoder:
## About The Pull Request

Now if your character's arousal is higher than AROUSAL_HARD_ON_THRESHOLD (20) * 2 (40 in total, I do not think we need a separate define here ngl) - they will hungrily ogle their ~~target~~ victim.

## Testing Evidence

image

## Why It's Good For The Game

Sovl, methinks? Also fun. Two in one package.


RandomGooberCoder:
I can't be assed to deal with endless complaints, w/e
The impact of this PR is too small to make me even bother with it

PR #1828, opened 2025-12-22
Corpsejak - makes necromancers able to summon carbon skeletons. Need to fix command undead. 

chaoko99:
## About The Pull Request
Title, but not presently finished.
Also gives necromancers ray of sickness and reworks ray of sickness to use spewium instead of toxin damage

- [x] Finish command undead carbonization

- [x]  Add raise undead for crit targets as fresh deadites

- [x] Make zizoids able to become deadites, and raise slower. Presently they are bite AND rot immune, which is odd.

## Testing Evidence
Not there yet.

## Why It's Good For The Game
I mean I like it. Carbon skeletons are like, silly bad.


Aracino:
Carbon skeletons would be a marked improvement over the simple skeletons. At least carbons can keep away a dumb adventurer from your evil skeleton lair. Won't do much good against a person who knows how to fight, obviously.

chaoko99:
also i added npc zombification and it was actually surprisingly easy.

WhatTheFlip339:
Raise dead with crit targets, do you mean purely player? Or could I raise an NPC bandit, or lets say... the mount decap wicker bandit? I also had a thought it would be cool to raise dead on a slain volf, but that is probably out of scope

WhatTheFlip339:
If you cant find a solution to the crit deadite conversion problem. Could just animate dead only work after they die, raising them immediately after death.

chaoko99:
I found a few solutions.

Ephemeralis:
Summons ideally need to not be dropping any gear at all, and disappearing quickly once killed.

chaoko99:
Is this impression made in context with my recent change?

PR #1833, opened 2025-12-23
Dark wood wall crafting, Iron Doors and more floors.

annikaRU:
## About The Pull Request

Adds crafting recipes for several floor types, mostly the church floors you can find and the floors inside the bathhouse

Adds a recipe for the reinforced doors that you can find at the keep and several houses

and most importantly adds a crafting recipe for dark wood walls and dark wood murderholes, no longer will you have to suffer with a disgusting wood wall inbetween two beautiful dark wood walls

## Testing Evidence

Compiled upon local host

Shop bars now place over tables as expected but has the downside of being able to be placed in walls or over objects.

image

## Why It's Good For The Game

Allows people to now build dark wood walls, they aren't stronger than ironically, regular wooden walls so they use the same materials but not much different.

More floors is always nice to be able to craft, you can make now iron doors which is something people wanted.

### DIFF
--- a/code/game/objects/structures/crates_lockers/roguetown.dm
+++ b/code/game/objects/structures/crates_lockers/roguetown.dm
@@ -18,7 +18,6 @@
 /obj/structure/closet/crate/chest/gold
 	icon_state = "chest3"
 	base_icon_state = "chest3"
-	sellprice = 70
 
 /obj/structure/closet/crate/chest/inqreliquary
 	name = "otavan reliquary"


annikaRU:
image

Snuck in glass floors being a high level masonry thing, if someone manages to do this...honestly good on them but I am not good enough to make it so you can see the literal voids of space if done above the lowest z-level

More than happy to not include it if maintainers deem it so

Edit: Decided to not include after discovering once you build these you can't actually place floors over them, someone could theoretically get a shitload of glass then build a lot of these to grief, best left to mappers I don't trust our playerbase to use these as intended

annikaRU:
New edit:

Adds in several furniture types for people to craft and a new door

The most notable ones being

Gold Chest, it is now able to be sold and crafted
Bookcases
Telescopes and Floor Clocks
all types of drawers
and two new table types

code/modules/roguetown/roguecrafting/structure.dm
```diff
@@ -129,6 +129,145 @@
...
+	reqs = list(/obj/item/natural/wood/plank = 2)
+	verbage_simple = "construst"
+	verbage = "constructs"
+	skillcraft = /datum/skill/craft/carpentry
+	craftdiff = 2
+
+/datum/crafting_recipe/roguetown/structure/goldchest
+	name = "Gold Chest"
+	result = /obj/structure/closet/crate/chest/gold
+	reqs = list(/obj/item/natural/wood/plank = 2, /obj/item/alch/golddust = 1)
```
RandomGooberCoder: Ok so let me get this straight.
Gold ore's sellprice is 10. Planks are basically free.
Meanwhile your edit is...

```/obj/structure/closet/crate/chest/gold
	icon_state = "chest3"
	base_icon_state = "chest3"
	sellprice = 70```
	
	No. Just outright no.

RandomGooberCoder (changes requested):

code/modules/roguetown/roguecrafting/structure.dm
annikaRU: Resolved, set to zero

RandomGooberCoder (approved):
prob ok

PR #1839, opened 2025-12-23
punchdagger balancejak

kiwedespars:
## About The Pull Request

punchdagger gets a faster version of pick intent. came to me in the shower
clickcd 14 -> 10
swingdelay 12 -> 6

## Testing Evidence

numberjak

## Why It's Good For The Game

punchdaggers are just shittier daggers that scale with unarmed, i don't really know what their purpose is i just get sad whenever i see someone using them since they're -5 force daggers

gives them their own identity kinda (good/bad thing? depends on if it becomes meta)

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/special.dm
+++ b/code/game/objects/items/rogueweapons/melee/special.dm
@@ -304,6 +304,10 @@
 	force = 25	
 	max_integrity = 120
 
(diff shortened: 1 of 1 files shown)


RandomGooberCoder (approved):
code ok, balance seems to be a-okay, let's give it a loook

PR #1840, opened 2025-12-23
Fixes vampire vitae being low not debuffing them, and makes it more severe

envoyin:
## About The Pull Request

Adds a debuff called "Bloodstarved", using already made (changed) values for when they occur. These are gained and lost depending on your VITAE. 

<1000 = -1 STR, -1 SPD

<750 = -2 STR, -2 SPD, -1 END, -1 CON

<300 = -3 STR, -3 SPD, -2 END, -2 CON

## Testing Evidence

image_2025-12-23_154347306

image_2025-12-23_154354744

image

## Why It's Good For The Game

There are two downsides to running out of VITAE. 

1) You cannot disguise. This does not matter if it's night time, which is most of the time.

2) You cannot use your funny vampire powers.

These don't actually do much to a vampire, and there was code already for making vampires getting thirstier when they get low on VITAE, but being thirsty didn't feel like much of a downside for spamming vampire powers.


envoyin:
> can these have examine texts or no

Probably? I know other debuffs can, so I'd have to see. I worry it would be gamered, but I think it'd be better to have an examine like "They look delirious." or "They look starved."

envoyin:
> I meant like examining the debuff itself, it shows the stat maluses but I figured it could have some flavor saying that you feel weak from lack of vitae or whatever

Oh, yeah. I forgot to add that in.

envoyin:
hwoops forgot to push that for a bit

envoyin:
okay after some thought and realization, most people won't hit 500 vitae in a fight even while spamming their abilities super hard unless they're losing like crazy

sooo... number experimentation, maybe, I guess

i will be performing the greatest balancing measure of our time, which is gauging it on personal experience and my schizophrenia, someone call me a retard if the numbers are bad

PR #1841, opened 2025-12-23
towerjak - kills south tower

chaoko99:
## About The Pull Request
don't wanna deal with the slog? build a better base, guys.
this was discussed at length in the dev chat and i figured it would be an interesting experiment.

build a wall guys

## Testing Evidence
nah

## Why It's Good For The Game
it may not be? Let's see.

### DIFF
--- a/_maps/map_files/dun_world/dun_world.dmm
+++ b/_maps/map_files/dun_world/dun_world.dmm
@@ -138,12 +138,6 @@
 "abQ" = (
 /turf/open/floor/rogue/hexstone,
 /area/rogue/indoors/inq/office)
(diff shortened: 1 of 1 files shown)


Plum-but-gayer:
does this replace the tower with a wall or just get rid of it?

pyrzal:
If the goal was to end the south tower palisade fort, a better approach would have been to move the Warden tower deeper into the south forest, then add more ways to travel from the forest to the town through the mountain range. That way you get rid of the cave choke. Could put the Warden tower near the broken bridge to guard it, that way it's just an optional pseudo chokepoint.

PR #1845, opened 2025-12-24
Antlerhoods on mask slots

NuIIpunkt:
## About The Pull Request

Adds ITEM_SLOT_MASK to the antlerhood

## Testing Evidence

one liner

## Why It's Good For The Game

I like using it, its a cosmetic item that costs 1 hide and 2 bones. F and D+ protection with 80 durability. 
COSMETICS COSMETICS COSMETICS.

This makes it viable for wearing without entirely giving up armor on the head
its a hood...all hoods fit on the mask slot


NuIIpunkt:
thiisss kind of isnt needed

PR #1802, opened 2025-12-20
Completely removes the Knight Captain from the Game. Adds a Knight Slot and a new unique Knight Subclass.

GrayRachnid:
## About The Pull Request
Completely removes the Knight Captain from the game. All mentions of him are gone.

Renames Royal Champion subclass to Irregular Knight
Adds a new Subclass: **Champion of The Duchy** to be renamed to Royal Champion down the line when people get used to it.

Champion of the Duchy gets a merged skillset of both KC Subclasses, in exchange for lack of orders and a crownstone:
image

**Middleaged Champions get a +1 SPD** so don't cry too much about the -1, that's just for Adult Champions.
They also get a saiga buck, like the mounted knight and the mounted KC before.
They also get the unique KC platelegs (alongside every other unique KC item) on roundstart, unlike other knights who only get chain legs.

### This is a unique slot!!
First come, first serve.
PQ required is 30 as per Maintainer discussion.

### Map Changes
- Marshal gets a whole new fucking floor. All for themselves.
- Knights get a fourth room(with mossy walls, it's the shit room) and a very, un-noticeably, smaller armory.
- Also adds unique keys to the Knight's doors, similar to the Bathhouse doors so Knights can now actually claim their fucking rooms.
- Also adds a 4th landmark spawn for knights in the throne room

### Champion
image
image
image
image
image

### Map Changes
image
image
image
image
image

image
image

## Why It's Good For The Game
There is a 3k+ message feedback thread on discord that had a majority of people agreeing that KC is a pointless middle-management role. Knights are self-sufficient. The previous PR wanted to just remove the role completely and give the drip to all the knights, but the lack of a Master-level keep-sided NOT VETERAN member would be felt and so they continue to exist but without any leadership aspects.

PR #1846, opened 2025-12-24
 Completely removes the Knight Captain from the Game. Adds a Knight Slot and a new unique Knight Subclass. 

GrayRachnid:
#1802 

Mapdiff bot broke


Dustssires:
If you want to know more about why this was done there is a thread in the discord in the feedback channel that talks about changing command structure of the garrison.

saltprjak:
> You're right. Adult champion is noob bait because middle aged champion gets their +1 spd back (-1 speed from base, +1 from middle aged, -1 again from middle aged).

I don't think it's a bad idea to let it *not* be noob bait. Royal Champion sorta pushes them away from an explicitly experienced leadership role and into the niche of the Duke's pet, or some sort of fighter prodigy. Only a minority will ever check Github to spot the age nerf, and whoever wants to play an adult royal champion will play an adult royal champion regardless. If anything, if this one singular point makes a difference, it'll be everyone else that suffers from the loss. The adult champion will roll on next round regardless.

EvenInDeathIStillServe:
need you to fix the icon conflict pookie

PR #1850, opened 2025-12-24
Fixes fixedeye getting stuck sometimes

FrownTown:
## About The Pull Request

Resolves an old bug that caused the fixedeye state to become stuck if you used the hotkey while holding down RMB.

## Testing Evidence

Tested it in game - held down RMB and hit the hotkey, let go of RMB and stayed fixed in the last direction I looked at. Hit the hotkey again and it turned off. Works in both scenarios (turning it on or off while holding down a mouse button).

## Why It's Good For The Game

Janky input handling makes me salt.

### DIFF
--- a/code/_onclick/drag_drop.dm
+++ b/code/_onclick/drag_drop.dm
@@ -201,12 +201,15 @@
 	if(mob.curplaying)
 		mob.curplaying.on_mouse_up()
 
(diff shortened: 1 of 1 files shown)


Plum-but-gayer:
YES thaaaank yoooou i hated that bug

PR #1852, opened 2025-12-25
makes heavy gronn mercenary Ever So Slightly less Ass

zeroisthebiggay:
## About The Pull Request

replaces heavy gronn mercenary's shitty ass poop mace with an actual grand mace
yes it's not much of an improvement and i guess it goes against the idea that everything gronn uses should be made of iron but there's no iron grand mace and i can't be assed to go like
port the maul that they're supposed to have so consider this a temporary solution until someone goes and does that

also removes comments referencing the maul and comments insistent upon the class's """"""""""weakspots"""""""""" as if you can't replace them very easily

## Testing Evidence

webedit if this breaks the video game i'll genuinely post feet pics to after round discussion

## Why It's Good For The Game

makes the giant armor heavy knight guy have an appropriately not dinky weapon


zeroisthebiggay:
> Give them the goedendang

what's that

GrayRachnid:
https://github.com/Scarlet-Reach/Scarlet-Reach/blob/f88085c61e9fe8a7570ce78af285b8be42717bd1/code/game/objects/items/rogueweapons/melee/blunt.dm#L381

PR #1857, opened 2025-12-25
Golems die at 800 brute damage taken

GrayRachnid:
## About The Pull Request
Yooris asked me to do this a week ago

## Testing Evidence

https://github.com/user-attachments/assets/ec6a6052-8b91-4f3e-8bd3-966c3062f237

## Why It's Good For The Game
Race with high con and no bloodloss is kind of hard to kill when they can only be decapped for it.
This fixes(?) that.
Number is up in the air, balancejakk it all you want. I don't personally have a stake in this.


Redrover1760:
It might be a good idea to make the number variable with CON, so a CON 20 golem and CON 1 golem die at different damage numbers

Like 800 + (CON - 10) * 20 or something

Ephemeralis:
We're just gonna do the 800 health thing for now. CON being a dumpstat for constructs is fine and satisfies my primal Pathfinder 1e urges.

CandleJaxx:
admittedly id love higher con = higher damage

PR #1858, opened 2025-12-25
rpjak: Rumours & Gossip

JustMeTheIInd:
## About The Pull Request
This PR adds a new way of (hopefully) improving RP potential; Rumours & Gossip. Both of these can be set by the player from the Character creation menu, and serve as a quick and easy option (that is also, hopefully bit more accessible "in the moment" when compared to opening, and reading an entire flavor text, some of which can get quite long.) But still one that requires being able to see that person's face, and in the case of Gossip, also being of Noble status.

Rumours:
Rumours are the more general, all-accessible option of the two as mentioned earlier. Visible to all who can see your face it's meant to serve as both general rp-hook and a way to hopefully flesh out your character and their interactions a bit more. Rumours don't have to be precise, nor true, in fact, maybe, the rumour is false, and if you allude to it during a conversation with said person, they will get angry at you for believing some stupid lie some idiot made about them?

Gossip: 
Gossip is almost the same as the above, with the sole exception that only Nobles can see it. This is so you can provide special infomration to other Nobles, without the common peasantry learning what they (logically) would have no way of knowing. You can, also in fact, be the target of Noble Gossip while also not being a proper Noble yourself, I decided to keep it. (You don't have to be a Noble yourself, to leave some impression  (be it positive, or negative) on them

I invite you to share your opinion on this PR and suggest additions/removals, but I warn you I am NOT an experienced coder, so some of the requested features might be just straight-up.... not really viable for me to add without outside help. Thanksjak. 

## Testing Evidence

It compilejaks 

Zrzut ekranu 2025-12-25 160108
Zrzut ekranu 2025-12-25 160053
Zrzut ekranu 2025-12-25 153334
Zrzut ekranu 2025-12-25 171453

I would like to thank the person that wrote the original flavor text code, as well as the "CTRL", "C" and "V" keys on my keyboard, for making this PR possible. You guys are the best.

## Why It's Good For The Game

The point of rumours is to provide an easy, fast and accessible way to learn more RP-oriented information about a character **without** having to open a whole ass new window and possibly having to spend a few moments reading through flavor text. (which, in some cases can get rather expansive.) An example, if you care: Janette Thief gets stopped for a Search by John Guard due to being a generally shady individual, that was spotted hanging around several places that suffered a break-in earlier, now, thankfully Janette Thief recalls that John Guard is known around the Town as a heavy drunkard, willing to do just about anything for ale- or for money for ale. So, maybe she strikes up a conversation, and offers John a bottle of strong wine she "somehow" came into the possession of. Maybe John Guard really is a drunk, maybe he takes the bottle and just waves his hand at Janette and tells her to ferk off before he starts thinking clearly. Or maybe, John Guard ain't a drunk, and indeed, he is **_fucking pissed_** that Janette would consider him some easily bribable drunkard, so maybe Janette gets one in the teeth, and is dragged off into the Dungeon both for disrespecting her betters, and for theft.

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -210,9 +210,6 @@ GLOBAL_LIST_EMPTY(chosen_names)
 	var/gossip
 	var/gossip_display
 
-	/// backward-compatible alias used by some examine code
-	var/rumour_noble
-	var/rumour_noble_display
 
 	var/tail_type = /obj/item/bodypart/lamian_tail/lamian_tail
 	var/tail_color = "ffffff"
@@ -1953,8 +1950,6 @@ Slots: [job.spawn_positions] [job.round_contrib_points ? "RCP: +[job.round_contr
 					g = html_encode(g)
 					g = replacetext(parsemarkdown_basic(g), "\n", "<BR>")
 					gossip_display = g
-					rumour_noble = gossip
-					rumour_noble_display = gossip_display
 					is_legacy = FALSE
 					to_chat(user, "<span class='notice'>Successfully updated Noble Gossip</span>")
 					log_game("[user] has set their noble gossip'.")
@@ -2916,11 +2911,11 @@ Slots: [job.spawn_positions] [job.round_contrib_points ? "RCP: +[job.round_contr
 
 	character.ooc_notes_display = ooc_notes_display
 
-	// Rumours / Noble gossip
+	
 	character.rumour = rumour
 	character.rumour_display = rumour_display
-	character.rumour_noble = gossip
-	character.rumour_noble_display = gossip_display
+	character.gossip = gossip
+	character.gossip_display = gossip_display
 
 	character.is_legacy = is_legacy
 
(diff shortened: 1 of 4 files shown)


GrayRachnid:
Fucker beat me to commenting

pyrzal:
Nice. Just work on the examine space it takes up. Otherwise, great feature.

GrayRachnid:
> Soulful. Make them into tooltips instead of buttons and we're golden.

wouldn't that make them 2 big as tooltips? hard to read. Unless we limit how much we can write there.

Ephemeralis:
You should be limiting how much is written in there anyway. People putting 600+ character gossip notes defeats the entire fucking point of gossip in the first place.

Short, sweet and punchy. That's how hook-bait is supposed to be.

JustMeTheIInd:
Added a 400 character limit for now, I suppose it can be TMed and then just adjusted according to community needs.

code/modules/client/preferences.dm
```diff
@@ -204,6 +204,16 @@ GLOBAL_LIST_EMPTY(chosen_names)
 	var/ooc_notes
 	var/ooc_notes_display
 
+	var/rumour
+	var/rumour_display
+
+	var/gossip
+	var/gossip_display
+
+	// backward-compatible alias used by some examine code
```
RandomGooberCoder: Triple /// allows coders to see the comment by hovering cursor over it in vscode.
God bless DMdoc. If you are going to comment vars - please, do it with a triple slash.

```suggestion
	/// backward-compatible alias used by some examine code
```

PR #1860, opened 2025-12-25
Wretched toiler Athletics  

benj8560:
## About The Pull Request

Gives toiler expert athletics. Onpar with most other wretch classes. The brotherhood ''spells'' are not ''magic'' thus they use athletics skill when calculating stamina draw. This means using any of these powers the toilers would instantly be stam crit (effectively like trying to cast a fireball with 5 int its not fun)

I put it on the base class rather then the mastermind subclass since well...running kinda important when you have 0 offense as the servant  does. 

## Testing Evidence
Compiles. Just numbers can't fuck up.

## Why It's Good For The Game
QOL. Just makes sense. This was most likely a oversight in the original PR and not intended.


Plum-but-gayer:
Smiles. This was definitely an oversight, the mastermind wasn't supposed to stamout after giving a single order. Even though that's an extremely funny image to think about.

PR #1862, opened 2025-12-26
graggar touchups

kiwedespars:
## About The Pull Request

```
heretics in general now spawn with max devotion and abilities.

graggar stuff should all be working now.
graggar cloak has storage now
graggarites with rituals can now sacrifice a noble/priest/templar to make portals of goblins upon the land
graggar armor now grants MOTIVE. +1 str +1 end. attached to the chestplate.
graggar net is a longer cd, hurts you to use.

all heretics get variants of hauberk/bracers/neck protection to go along with their sets of armor

nets are less cbt and fall off after 10 seconds if they bug out.

ravox now heals based on nearby enemy worshippers (non-tennites, this includes you, psydonites)

zizo got +20 to their boneshard because that shit ass.
```
## Testing Evidence

image

## Why It's Good For The Game

graggar stuff. touchups. let the violent khorne-lites feel like violent khorne-boys.

being more serious though. graggar is pretty underwhelming because their "I AM VALID" armor is actually the weakest in comparison to zizo/matthios. this should make them feel more like berserkers/melee dominating fighters/etc.

the miracles they have are also broken and this should fix them and also remove the funny bug of endless net debuff (we can re-enable those now or something). it's also more flavorful that they thrive in bloodied environments and that ravoxians love validhunting

also cool little goblin portal event.

also because zizo shard is some dogshit but idk i don't use it maybe it's great

### DIFF
--- a/modular_azurepeak/code/game/objects/items/ritualcircles.dm
+++ b/modular_azurepeak/code/game/objects/items/ritualcircles.dm
@@ -832,12 +832,14 @@ var/forgerites = list("Ritual of Blessed Reforgance")
 		H.dropItemToGround(I, TRUE)
 	H.drop_all_held_items()
 	armor = /obj/item/clothing/suit/roguetown/armor/plate/full/zizo
+	shirt = /obj/item/clothing/suit/roguetown/armor/chainmail/hauberk/zizo
 	pants = /obj/item/clothing/under/roguetown/platelegs/zizo
 	shoes = /obj/item/clothing/shoes/roguetown/boots/armor/zizo
+	wrists = /obj/item/clothing/wrists/roguetown/bracers/zizo
 	gloves = /obj/item/clothing/gloves/roguetown/plate/zizo
 	head = /obj/item/clothing/head/roguetown/helmet/heavy/zizo
 	backr = /obj/item/rogueweapon/sword/long/zizo
-	neck = /obj/item/clothing/neck/roguetown/bevor
+	neck = /obj/item/clothing/neck/roguetown/bevor/zizo
 
 
 /obj/structure/ritualcircle/zizo/proc/rituosbone(mob/living/carbon/human/target)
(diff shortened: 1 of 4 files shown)


saltprjak:
> yes and the idea of swapping out your armor for something better on coverage is silly. with the other ascendant sets, you have 750 + 600 armor on legs. graggar set has only 400 leg armor.

It's medium armor, which you can give out to pretty much every non-dodge combat role, and it can be taken off - which, again, means you can always take off your supervillain suit and go back to playing adventurer/towner and do your groceries like everyone else (I play wretch whose gimmick is built around other wretches, even lunacy embracers stand in The Line). Or just repair that armor, something that can't be done without spells or rituals for the other two.
Their axe is still by far the most busted part of their kit that the sword and the flail from others don't nearly come close to, and the extra STR point on top synergizes all too well with their uncapped strength. And now with nets fixed, getting stuck in a 1v1 with a Graggarite means you'll be unlikely to even escape if you wanted to.

You could make a point that a Zizo heretic can be on par with a Graggarite through spells - but Graggarites just don't sacrifice anything for their powers. No getting armor glued on, no losing stats or getting a permanent >KILL THIS GUY< tag from skeletonization. They're the ultimate wretch antag of convenience, and, even without the buffs, IMO the strongest in the roster thanks to their deathstick and the stat synergies they get.

tl;dr fixes and qol are great, i don't think they need buffs because they're strong as is, and unlike others have ZERO long-term consequences to fraggery so long as they keep their frag-sona identity secure when they inevitably need to restock on str pots at the merchantry

modular_azurepeak/code/game/objects/items/ritualcircles.dm
```diff
@@ -1024,30 +1022,43 @@ var/forgerites = list("Ritual of Blessed Reforgance")
...
+			if(!do_after(user, 5 SECONDS))
+				return
+			user.say("LET PORTALS OPEN, LET 'HIS' CHILDREN COME!!")
+			if(!do_after(user, 5 SECONDS))
+				return
+			icon_state = "graggar_active"
+			if(perform_warritual())
+				user.apply_status_effect(/datum/status_effect/debuff/ritesexpended_high)
+			else
+				to_chat(user, span_smallred("The ritual fails. A noble, member of the inquisition or a tennite churchling body must be in the center of the circle!"))
```
zeroisthebiggay:
```suggestion
				to_chat(user, span_smallred("The ritual fails. A body of noble blood - from the inquisition - or from the church must be on the circle!"))
```
there is no center it's just a circle and also better clarification on who we want

PR #1865, opened 2025-12-26
Ports Lets slurbows be onehanded + let dualwielder affect slurbows

kiwedespars:
## About The Pull Request

ports https://github.com/Azure-Peak/Azure-Peak/pull/4978

## Testing Evidence

image

## Why It's Good For The Game

everyone believed they could fire slurbows one-handed and now reality matches that expectation


JonathanTheII:
Is this the case for the old slurbows that monster hunters get also? The one handed firing would go nice with their silver sword.

kiwedespars:
ye

PR #1866, opened 2025-12-26
destroys bait on omni intent

kiwedespars:
## About The Pull Request

destroys bait on omni intent

## Testing Evidence

i just removed it if it looks fucked up idk what to say

## Why It's Good For The Game

baiting has risk involved. if you fuck up your gambit, you put it on cooldown for 30 seconds. 

omni intent specifically checks for if they're aiming the same limb and if it is, then it'll do it, otherwise they won't. 

there's no risk or gamble involved it's just cheatcodes

### DIFF
--- a/code/game/objects/items/rogueweapons/rmb_intents.dm
+++ b/code/game/objects/items/rogueweapons/rmb_intents.dm
@@ -339,15 +339,6 @@
 		if (ishuman(target))
 			HT = target
 
-		// RMB on mob (priority 0): check to see if a bait has any chance to succeed (match targeting zones between us and the target), if so, attempt it (ONLY check for matching zones, nothing else).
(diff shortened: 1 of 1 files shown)


Ephemeralis:
The ontologically correct way to do this is to just put a 6 second last-time check into the intent itself before it attempts a bait check. As in: it will attempt a bait only once every 6 seconds no matter how hard you right-click spam it.

zeroisthebiggay:
> baiting has risk involved. if you fuck up your gambit, you put it on cooldown for 30 seconds.

fucking up your bait also gives the person you were trying to bait their resistance to being put off balance from being baited twice
using omni intent straight up bypasses this feature and lets them fumble baiting you with impunity

PR #1871, opened 2025-12-27
Eoran Sallet

NuIIpunkt:
## About The Pull Request

I found the sprites for an eoran sallet wich seemed to be a reskin of the regular one. Since it wasn't a donator item I made it craftable

## Why It's Good For The Game

Gives purpose to an item that already existed


NuIIpunkt:
> Do you think it could be added as an option for templars on spawn, like the astrata/necra options? The regular Eora helmet looks... odd.

Likely but it would be a downgrade from what they get. This PR doesnt aim to giving it to templars or paladins. Sorry!

PR #1873, opened 2025-12-27
Ports Helmet Aesthetic Storage

kiwedespars:
## About The Pull Request

ports https://github.com/Azure-Peak/Azure-Peak/pull/4374

shift rmb helmets to open up a mini inventory on them.

## Testing Evidence

image 
image

## Why It's Good For The Game

drip.

### DIFF
--- a/code/datums/components/storage/storage.dm
+++ b/code/datums/components/storage/storage.dm
@@ -52,6 +52,7 @@
 	var/atom/movable/screen/close/closer						//close button object
 
 	var/allow_big_nesting = FALSE					//allow storage objects of the same or greater size.
+	var/allow_nesting = FALSE						// fits in storage items of the same same or smaller size
 
 	var/attack_hand_interact = TRUE					//interact on attack hand.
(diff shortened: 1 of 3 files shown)


TheMoralOne:
Screenshot 2025-12-28 023226
Screenshot 2025-12-28 023301
Screenshot 2025-12-28 023310

I was playing around with this and found a few things.
1. Items (at least, the scarf and capes that I used, unsure if it is the same with others. I would assume it to be.) within hats/helmets will not respect whether the item itself is set to display above or below hair, and if they have no toggleable display option will always display above. I tried taking the item in and out after changing its display setting, but no dice.
2. I noticed that capes can also be inserted into hats, and will lose their side and back sprites, and for the front it appears as though it shows the back sprite, as though it was being worn backwards.

Just some things I found, over all I do really love this though!

JonathanTheII:
> Stuff like hats, rosas and zigs start to float next to you when stored that way, as if they were held in hand or something. 

After further testing I think this was just a glitch with giant. Now it seems to work normally.

PR #1874, opened 2025-12-28
flowstate - bait changes

kiwedespars:
## About The Pull Request

```
How Baiting Used To Work!

Match your zone to your opponent's, exactly.
If you do, you place them on 5 seconds of severe clickdelay malus, being unable to swing back or do anything.
Give them 10 seconds of exposed, allowing you to beat on them freely in this time.
Give them a stack of bait, resulting in the next bait to immobilize and put them off-balance.
Sap their stamina somewhat.
```

```
How Bait Works With This PR!

Match your zone to your opponent's. Right hand counts as right-arm, anywhere on head counts as head, so on and so forth.
If you do, place them on 4 seconds of severe clickdelay malus, being unable to swing back or do anything.
Give them 4 seconds of exposed, allowing you to beat on them freely in this time.
Give them a stack of bait, resulting in the next bait to immobilize and put them off-balance.
Sap their stamina somewhat.

Grant yourself twenty seconds of Flow
Flow returns a third of your stamina to you and essentially gives you Guidance (+20% parry/dodge defense, +20% parry/dodge bypass)
Flow grants you +1 SPD, PER, and INT.
While you have Flow, you're immune to all incoming baits.
```

```
OTHER CHANGES

Bait immunity now has an icon. It's blue feint.
You have to be within someone's view to bait them. No more back-baiting.
You can always feint a guard to cancel it, rather than it being chance-based.
Having a bait failed on you will make you immune to further baits for 15-20 seconds.
Failing a feint via making a fool out of yourself will slow you down for two seconds.

```

## Testing Evidence

compiles and everything works. (except for the vision check for bait. npcs are hard to test with)

## Why It's Good For The Game

precise inputs means when someone smacks you on the head and you go "heh they're on the head" they were actually aiming for your left occipital lobe you should've understood that stupid.

bait itself is far too punishing to the point where most people will hang off until they've placed your bait on cooldown. ten seconds of beating on someone is easily 7-8 hits with a weapon, not even counting swift, and most armor layers will break to this or be /greatly/ damaged.

baiting in third party is incredibly frustrating because it exacerbates the issue further. instead of 10 seconds of being beat on by one guy, it's now two, three, or four. everyone has the potential to throw a bait out and turn you into mincemeat because you cannot do anything for five seconds to begin with. 

generally the idea is to shift "fuck this guy in particular HARD" to "i'll get a few hits in and then have a buff for a bit as a prize for that read"


kiwedespars:
i wanna do some more changes don't merge this yet.

saltprjak:
Could make failing a bait more punishing? Feels like there's way too much reward for too little risk, at least compared to poorly calling a riposte.

anything86:
Being 100% honest, I think this just shifts the meta towards chest aiming. You are now actively encouraged to start a fight aiming chest, because it is braindead easy to land a bait otherwise and the first person to land a bait can spend the next 30 seconds not worrying about the mindgames (because they are now immune to baits for 30 seconds).

Conversely, because chest aiming is now so encouraged (as the only way to be immune from baits which were made super easy to pull off), this **primarily** hurts anybody aiming for locations that are typically not covered by chest, **because their damage will be getting split up.** 

``Match your zone to your opponent's. Right hand counts as right-arm, anywhere on head counts as head, so on and so forth.``
It shafts Head, Hand, Feet, and face aiming, while keeping arms undesirable to aim for and keeping legs mostly unaffected (because unlike Head, **you can still chip at an opponent's leg armor**, because leg armor integrity is shared between both legs - letting you avoid baits while chipping leg armor).

In essence, I'm still very much opposed to making baiting easier and I am not super thrilled about it. I recall you briefly saying "Just swap to chest after every hit on the head!", which is rich considering you quite literally would not need to do that while aiming for the legs, as you can just swap to the other leg. Your solution to a pretty glaring issue is "flick two extra hotkeys every 0.8 seconds".

EDIT: Somebody else already suggested you this, so I will repeat their thoughts here: Split Head baiting into two separate zones. I don't care which, choose something that makes sense and do it. Having Eight (8) (VIII) locations share the same zone for baiting is retarded.

anything86:
> there is no room for error on this, you know /this/ is what it is. i swung at your right leg.

You can add this extra information on the message it sends when they hit you, "X tries to swing at Y's A but hits Y's B instead".
I am not sure off the top of my head how close the accuracy check is to the message being sent (and such if this information would need to be carried through multiple procs) but its worth a look and much more definitive a solution than this.

PR #1883, opened 2025-12-29
nervemaster visuals revamp + Economy changes

chaoko99:
## About The Pull Request
revampst he steward console a little bit, deleted compact mode for default behavior, visualized PPU for import and export, and also killed passive regen entirely. I have also widened the roundstart stewardry range for funnies.

## Testing Evidence
image if it's green, you're ready to export.

## Why It's Good For The Game
playing steward sucks.

### DIFF
--- a/code/controllers/subsystem/rogue/treasury.dm
+++ b/code/controllers/subsystem/rogue/treasury.dm
@@ -1,4 +1,4 @@
-#define RURAL_TAX 15 // Free money. A small safety pool for lowpop mostly
+#define RURAL_TAX 50 // Free money. A small safety pool for lowpop mostly
 #define TREASURY_TICK_AMOUNT 6 MINUTES
 #define EXPORT_ANNOUNCE_THRESHOLD 100
 #define FOREIGNER_TAX_MULTIPLIER 1.5 //Amount that the tax rate is multiplied by for foreigners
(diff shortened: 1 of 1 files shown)


IronDragoon:
Yeah just to say that during low pop yesterday we had several rounds with NO food items in the stockpile at all. We really need *some* regen, it was pretty crap

chaoko99:
You don't need skills to grow your own food but that's beside the point

SylvinSeik:
This makes playing cooking roles (We have so many ingredients that never get farmed and servant/cook farming is already rather terrible), tailor (fat/tallow), and similar roles a total pain in the ass to play.  

I really don't think we should be making playing noncombats more annoying when possible, given the state and reputation of the server.

EvenInDeathIStillServe:
>playing steward sucks
>makes every other role dependant on the stockpile suck instead

it's spectacular

chaoko99:
I'm kind of blazed right now, if I have the energy to do so I'll try to add seed vendor. 

For the other roles, what is missing?

annikaRU:
Adding a seed vendor doesn't fix the main problem that no one plays farmer, making it more inconvenient to get food related items isn't going to make people play soilson more

saying "growing crops is easy" is a "making your own electricity isn't hard why don't you make your own" level of response

Id argue that resources like iron, wood, and rocks aren't needed to have passive regeneration, because they arent difficult to get, you could say the same for like blackberries but to make a simple cake you now need to grow wheat, the berries and get eggs plus sugarcane that you need at minimum novice to not spoil the produce when harvesting and you need to watch it, feed the soil and water it versus just chopping a tree to get a small log

chaoko99:
I actually canot be fucked to go back and manually revert the change for the stockpile regen so uh
yeah you just get nothing until i care

chaoko99:
WAIT A SECOND I JUST COMMENTED 3 LINES

chaoko99:
this can be merged given the literal only complaints have been whining about it being too hawd for bawbie townies to hit go outside and pick berries

chaoko99:
economy changes reverted
I do not want to continue working on this specifically.

PR #1885, opened 2025-12-29
Psydon Lives! Aka makes roundstart-selected Psydon rule forever and close bandit slots

RandomGooberCoder:
## About The Pull Request

Adds a few things to storytellers.
- on_vote_chosen()
Called when a storyteller is selected via roundstart vote. For custom roundstart behavior.
- var/rules_forever
Does what it means, yes. If set to TRUE - storyteller should not be replaced in influence calculations
- Sets **roundstart** Psydon to rule_forever = TRUE.
- Makes **roundstart** Psydon close bandit slots
- *Should* make Psydon disable raids and character injections completely.

## Testing Evidence

image

## Why It's Good For The Game

A guaranteed way to get a chill round without bandits, liches and shit.


CaydenT:
I think turning bandits off entirely might be heavy-handed. Lowering their slots might be better, so that they only have the numbers to do.... Banditry, instead of being a whole-ass army. 

Turning off all other RNG spawned antags is awesome though, and much needed. 

Its likely out of scope for this PR but a system to prevent the same storyteller from being voted in each round may be ideal, too, to prevent Psydon from being *every* round

benj8560:
what tends to happen is all the antags that would likely ready for bandit end up reading for a WAY more problematic antag and..such antags bandits are likely to fight (apocalypse is not very matthosian). Psydon just ends up being guaranteed big antag without fail so far. These rounds have been some of the highest chaos i've ever seen.

Plaguewalker:
I'm just a simple man, but if we get a psydon round it should disable possible round-start antags (Mostly the big ones like Werewolves, VL's, Lich). Along with halving the slots for wretch, bandits and lunatics this way everyone still wins.

PR #1891, opened 2025-12-30
Church bell spam reduction

Flamm10:
## About The Pull Request

Title. Yes, it's a salt PR.
Big bell's cooldown 3 seconds  -> 30 seconds
Small bell cooldown 3 seconds -> 5 seconds (collateral damage)

## Testing Evidence
*groanirl
1. Compiles
image

2. Spawned as priest, ran to the tower, rang both bells. Works.

## Why It's Good For The Game

1. This. No, that particular round was not the first with such abuse, but something something camel's back.
image

2. Personally i feel this is an oversight for two reasons: 
First, when it was added (eh, whatever, churchlings will just beat the abuser up)
Second, when the range got inreased to global (NOW we are talking)

The big bell is used usually for 2 reasons. 
A) CHURCH UNDER ATTACK SEND DUDES THE YELL IS ON COOLDOWN
B) Ingame wedding.

P.S: If you have a better solution, let me know anywhere and i will take this down.
Or speedmerge this if you are tired of the spam too :^)

### DIFF
--- a/code/game/objects/items/rogueitems/bells.dm
+++ b/code/game/objects/items/rogueitems/bells.dm
@@ -11,7 +11,7 @@
 	damtype = BRUTE
 	force = 5
 	hitsound = 'sound/items/bsmith1.ogg'
-	var/cooldown = 30 SECONDS
+	var/cooldown = 5 SECONDS
 	var/ringing = FALSE
 	resistance_flags = FIRE_PROOF
 	grid_width = 32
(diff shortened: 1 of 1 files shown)


Githubaccount222:
just normalize killing characters who do this??

Flamm10:
> just normalize killing characters who do this??

In that specific round:
1) Admins were busy/afk/erping, i'd imagine at least one ahelp about this was sent.
2) Church was indeed under attack, so nobody could afford to go and beat the one ringing.
3) I shouldn't be running **across the fucking map** to beat some sense into someone.

30s isn't even a lot, deal with it.

PR #1900, opened 2025-12-31
Etruscan Bascinet now as durable as all other visored helmets

saltprjak:
## About The Pull Request

etruscan bascinet durability moved to 325, in line with knight helmet/armet/hounskull and other same ingot-price and class distribution equivalents

## Testing Evidence

image

## Why It's Good For The Game

It's CUTE! No reason it shouldn't be viable.


DerChud:
AVANTI ETRUSCA

NotGoldenglow:
> image if you like changing helmet durabilities feel free to standardize all of these lol.

Don't wide-sweep standardize them all. They have key differences.

Notably, the 450 Froggenmund covers your neck BUT limits your view to a 90° frontal cone, as opposed to the standard 180°.
The visored sallet does not protect your mouth.
A fair share of those helmets either do not have a visor, or do not protect your ears, or both. Thus should be less durable than helmets that do.
A fair share of the 'weaker' helmets are also iron, and thus, should be less durable.

If anyone standardizes these, they should take all hese factors into account.

PR #1903, opened 2026-01-01
rewrites throwing and jumping, slight jumping "buff", widens glide size clamp.

chaoko99:
## About The Pull Request
Throwing rewrite, can now move diagonally instead of weird stair stepping.  The throwing coverage is kinda.. square, right now. But I don't care.

Jumping uses a callback instead of a sleep now, so should feel a little more fluid.
halved the immobilization when long jumping, because it kinda feels like dogshit.

## Testing Evidence
https://streamable.com/jtu112

## Why It's Good For The Game
i honestly don't recall what possessed me to do this in the first place. But it's done now.


RandomGooberCoder (approved):
Uh boy surely we are not shooting ourselves in the knee some time later

PR #1912, opened 2026-01-01
Giving Roles Tempo

kiwedespars:
## About The Pull Request
```
- ADDS TEMPO TO -

hedgeknight
iconoclast
berserker
deserter (both variants)
frei
gronn norsii
heretic
disgraced custodian
onimusha
warlord
avatar of graggar
knight captain
the martyr
the inquisitor (plate armor variant)

and

ANYONE OLD
```

additionally, reduces the armor integ damage reduction from 20%/40%/60% to 20%/30%/40%

old people in general can now lock in when being attacked by multiple people. it was mostly just because it's funny

## Testing Evidence

yes it works. it just gives traits.

## Why It's Good For The Game

tempo tempo tempo tempo.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/wretch/onimusha.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/wretch/onimusha.dm
@@ -67,7 +67,7 @@
 	cmode_music = 'sound/music/combat_swordhunter.ogg'
 	origin_override_type = /datum/virtue/origin/kazengun
 	subclass_languages = list(/datum/language/kazengunese)
-	traits_applied = list(TRAIT_DODGEEXPERT, TRAIT_TEMPO)
+	traits_applied = list(TRAIT_DODGEEXPERT)
 	//Bad stats for a wretch, instead they get some really good traits and skills.
 	subclass_stats = list(
 		STATKEY_SPD = 2,
--- a/code/modules/jobs/job_types/roguetown/nobility/knight.dm
+++ b/code/modules/jobs/job_types/roguetown/nobility/knight.dm
@@ -102,7 +102,7 @@
 		/datum/virtue/combat/crimson_curse
 	)
 
-	traits_applied = list(TRAIT_HEAVYARMOR)
+	traits_applied = list(TRAIT_HEAVYARMOR, TRAIT_TEMPO)
 	subclass_stats = list(
 		STATKEY_STR = 2,
 		STATKEY_CON = 2,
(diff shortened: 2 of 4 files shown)


benj8560:
> inquisition gets fucked by wretches because they don't play smart. they get into bad fights, their classes are incredibly strong with good statlines, not to mention people already run from them because the absolver carrying the musicbox makes them a crazy ass threat.

more like entire bandit gang /5 wretches appear on inquisitions doorstep while their RPing and mob them to death. Is more what I'm thinking of...not sure that's a ''bad play'' on their end. (maybe they get tempo on their home turf??)

Rest is fair I guess was just a suggestion. I'm like 80% sure darksteel tempo will still delete the garrison but! TM and find out I guess. Maybe silverweakness finally being on rituos will do the trick to.

benj8560:
> 3 confessors, an inquisitor, and an absolver makes it into a 4v5. these are not bad odds at all

1 healer, 4 fighters. Most likely one of them is out handling something so lets call it 3 fighters vs 5 people 2 of them most likely in gigapowerful ritual armor/weapons. 3 of those people are most likely healers so attritions not a big deal. absolutely anytime this happens it's gonna be loud. SO its gonna be ritual armor so its gonna be a pitched battle. Statspreads don't reflect having the best armor and weapons in the game. btw I have seen just TWO people in ritual armor pull this off so.

I can't count how many times I've seen the inquisition all slain in the manor. Only time it does not happen is if someone has a scomstone to call a army

wretches/bandits almost always have full slots. Same can't be said for inquisition (since their a punching bag half the time and getting mobbed like this is really depressing) I'd also like to say! I don't even play inquisition anymore mostly for this reason. I play a lot of wretch and I am in fact. Not scared of these guys at all.

> what do you think tempo does

if you mean more what its doing to allow 2 players to kill the entire servers fighting classes? Ritous's stamina effect plus nearly free parrying ontop of no parry cooldown. ONTOP of mending spell means they can actively fight indefinitely with very little risk of over extending and getting killed. The build can already do this kind of thing without tempo. Tempo just makes it extremely forgiving.

The only way to ''counter'' tempo is to fight a 1v1 or 2v1 fight. How often do you see a duel with the one in ritual armor lose? Can't say I have ever personally seen it. Not even once.

ChickyRooster:
I am not sure how balanced it would be, but the druid class has their massive mole form they can only get via doing a ritual, that one would make sense to have tempo but I haven't played around enough with tempo to know how good (or bad) it is.

FreeStylaLT:
don't forget you can tweak what it actually does at any point

Sgt-Solid01:
Oh yeah i've been seeing that too but that's just the current influx with randos wanting to try the new weapons, im pretty sure the hype will die down soon

But you're the coder, you decide, im just a player, my opinion comes from experience playing said role

PR #1913, opened 2026-01-02
Nitetouched Virtue (Nite-Kin)

Kroto-Ker:
## About The Pull Request

Adds a new virtue exclusive to Wild-Kin that allows them to access a unique subrace named 'Nite-Kin'. 
This changes their body type, ears, and eye appearance. 
Their origin is switched to the Underdark and they gain a spell called 'Nitelink'.
Nitelink works similarly to mindlink, except that it only works on those with the Underdark origin. It also is cast directly on people within your vision radius, instead of using a 'known people' list.

image

image

image

## Testing Evidence

No runtimes caused by this PR. The division by 0 is due to a different merged PR.
image

## Why It's Good For The Game

An implementation of a subrace virtue that gives people more tools to create interesting characters, and helps expand some of the setting related to the Underdark.


ChickyRooster:
Shadekin in scarlet reach will be real in 2015

Kroto-Ker:
Closed until better sprites can be obtained.

PR #1921, opened 2026-01-02
The Old Ways II: More witch flavor and miscellaneous improvements (WIP)

Ephemeralis:
## About The Pull Request

Warning: I'm going to be locking down Witch slots as part of this PR chain at some point in the future. There's not going to be more than 4 of them in a round. This is already probably too many.

Anyway, stupid lore/soul PR purely because I feel like it:

- Confessors/puritans can no longer force witches to confess their patron.
   - This is because witches mechanically do not really worship their patrons, or are at least, not supposed to. You can of course, choose to ignore this at your discretion. The game will still portray your worship as non-standard, because you're a witch.
   - The "Old Ways" are essentially lost magickal practices that allow witches to draw upon pantheonic power through a variety of esoteric and broadly unknown methods.

## Testing Evidence

fuck you it compiles

## Why It's Good For The Game

More flavor gooderer.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/Inquisition/puritan.dm
+++ b/code/modules/jobs/job_types/roguetown/Inquisition/puritan.dm
@@ -291,6 +291,12 @@
 		)
 		src.visible_message(span_warning("[src] shoves the silver psycross in [H]'s face!"))
 		say(pick(faith_lines), spans = list("torture"))
+		if (HAS_TRAIT(H, TRAIT_WITCH))
+			to_chat(H, span_notice("The weepers don't understand. They don't know what you are."))
+			to_chat(src, span_notice("[H] does not respond - your compulsion is fruitless. How can this be?"))
+			if (prob(30))
+				to_chat(src, span_warning("The hairs upon the back of your neck rise..."))
+			return
 		H.emote("agony", forced = TRUE)
 
 		if(!(do_mob(src, H, 10 SECONDS)))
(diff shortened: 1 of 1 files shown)


RandomGooberCoder:
>Warning: I'm going to be locking down Witch slots as part of this PR chain at some point in the future. There's not going to be more than 4 of them in a round. This is already probably too many.

You sure about this? I believe that if a role deserves to be slot-limited, it should not be a part of adv/towner job, being its own job instead.

CaydenT:
I agree that they probably need to be limited, however try not to be heavy-handed. Witch covens tend to be three or so, in most media. 18 in one round is crazy though. The middle ground of about 6 should be fine. I'm curious as to whether the round full of them was immediately after their new features? Because of course people are going to play them when they have shiny new toys.

Kroto-Ker:
> I agree that they probably need to be limited, however try not to be heavy-handed. Witch covens tend to be three or so, in most media. 18 in one round is crazy though. The middle ground of about 6 should be fine. I'm curious as to whether the round full of them was immediately after their new features? Because of course people are going to play them when they have shiny new toys.

I mean stuff like towner blacksmith is slot-limited to 2, thug is 4 iirc. I don't see why witches can't get a similar treatment.

ophaq:
Not a fan of the witch slot limit, if anything just make it so thug and blacksmiths can be unlimited too. Witch is probably the most fun class to play for towner roles, which is why people play them. Forcing them to choose other jobs is just going to make them not play the round at all by leaving the round after spawn manually which SUCKS if the role you want is filled, and end up switching characters. This just drives people away from the game in the long run.

Sure, you can go sorcerer. And then waste a virtue slot to be like a towner? That's lame. Plus you don't get the cool witch transform spell, and you can't be godsblood/etc.

PR #1928, opened 2026-01-03
The Old Ways III: Spell tier rebalances for Noc & Zizo, profane spell qol improvements

Ephemeralis:
## About The Pull Request

Mostly intended to make the godsblood witch worth picking, this PR touches a few things:

- Profane (Zizo T1) has been significantly improved and should now actually be worth using.
   - The spell has a 4 second cooldown (down from 15s) and closely matches Spitfire for overall ease of use and user feel.
   - If you're wearing a zcross around your neck, every Profane cast is considered 'empowered' and shoots the 55 damage lance, with extras explained below.
   - Devotion cost reduced to 5 from 15.
   - New bespoke casting sfx and impact sfx (from pixabay), and the whisper invocation is removed (so as not to spam your chatlog).
   - Simple animals now receive +10 bleeding on hit when other carbon mobs would've normally received an embedded splinter.
   - Embedded splinters cause 5 bloodloss per tick. Four embedded splinters is equivalent to your standard arterial wound. Note: they have a chance to fall out on each movement unlike other embeddables.
   - Profane now deals 150% damage to NPCs by default, with the empowered bone lance (from using bones or wearing a zcross) dealing 200% instead, similar to arcyne bolt and other mage PvE spells.
      - Empowered profane does 110 brute damage to npcs with a 30% chance of +10 bleeding, making it one of the strongest pve  spells in the game.
   - Unempowered casts now do PIERCE damage instead of STAB damage.
   - Empowered casts now do CUT damage instead of STAB, making it theoretically possible to dismember with empowered casts on unarmored targets.
- Rituos (the miracle) has been moved from T3 to T2.
- Zizo's woundheal has been moved from T4 to T3.
- The Noc spell bundle has been moved from T3 to T2.
- Noc's woundheal has been moved from T4 to T3.
- Witch transforms have been reduced in cooldown from 1 minute to 15 seconds, but every point of damage a witch's human form has adds that many deciseconds onto the recharge for the next shifting attempt. 200 damage = 20 seconds of extra cooldown time.
- Witch shapeshifts can be immediately attempt to be shifted OUT of instead of needing to wait the cooldown to try and hop out of them.
- Damaged witch forms now take significantly longer to shift out of, up to 6 seconds at 10% health remaining (used to be 3 flat).
- Witch shapeshifts now play shapeshifting SFX added in the druid updates.

## Testing Evidence

https://github.com/user-attachments/assets/e6543a57-588f-4471-964d-66f2cc04ba5e

## Why It's Good For The Game

Profane hypersucked before, and I'm sorry, because it was my original design in the first place. AP miracle balancing is a very different beast to what is expected from SR and modern magery, so profane has always kind of been left behind in the past year as an actual viable option. With these changes, it becomes a mainstay bone-hurling monster, and if you're willing to out yourself by wearing a zcross, you don't have to lug a stack of bones around with you for it to become especially terrifying.

As for the spell tier changes, the only class in the game that gets T3 cleric stuff innately is Lunacy Embracer, which people typically roll Dendor for and nothing else. Making spell bundles/rituos accessible to T2 mostly affects the witch, but also makes other less played options like cantor have something new to fuck around with as well.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/witch.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/witch.dm
@@ -142,6 +142,7 @@
 
 /obj/effect/proc_holder/spell/targeted/shapeshift/witch/Shapeshift(mob/living/caster)
 	// Do-after before transforming
+	playsound(caster.loc, 'sound/body/shapeshift-start.ogg', 100, FALSE, 3)
 	if(!do_after(caster, 3 SECONDS, target = caster))
 		to_chat(caster, span_warning("Transformation interrupted!"))
 		revert_cast(caster)  // Refund the cooldown
@@ -167,6 +168,7 @@
 	var/total_damage = shape.getBruteLoss() + shape.getOxyLoss() + shape.getFireLoss() + shape.getToxLoss()
 	var/shift_time = 3 SECONDS + (total_damage / 10)
 	// Add do-after for witches when reverting
+	playsound(shape.loc, 'sound/body/shapeshift-end.ogg', 100, FALSE, 3)
 	shape.visible_message(span_warning("[shape] begins to shift back!"), span_notice("I begin to transform..."))
 	if(!do_after(shape, shift_time, target = shape))
 		to_chat(shape, span_warning("Transformation revert interrupted!"))


WhatTheFlip339:
conflicts with my PR, wound heal is T4 and I intended it to be strong only for a specific class. Wound heal T4 was also to stop it from being round start unlockable, lowering it to t3 is power creep and will make everyone else go "why is my wound heal t4? theirs is t3". IK lunacy embracer is the only one that gets T3, it would be too strong on it. idrc about rituos

PR #1935, opened 2026-01-03
Adds Palisade around the Southern Warden Tower

JustMeTheIInd:
## About The Pull Request

This PR adds palisade around the southern Warden tower.

## Testing Evidence

Zrzut ekranu 2026-01-03 143736
Zrzut ekranu 2026-01-03 141521

## Why It's Good For The Game

QOL.

Building the palisade is not fun, not engaging, not interesting nor does it present any actual challenge. It is simply a timesink. The entire """gameplay""" loop amounts to: click tree, click log, click small log with open hand, walk over, click "craft", click "wooden palisade" And the palisade itself isn't some great and miracolous boon to the Wardens, it's easy and fast to break, and even if it wasn't there's still like- ~5 different way to get past the palisade with the risk of any Warden spotting you being minimal. This PR simply wishes to remove the tedium from the game. Base building will still be possible, and helpful since just about anything is better than the palisade-- but the Wardens won't have to spend ~20 minutes **each** round clicking mindlessly.

### DIFF
--- a/_maps/map_files/dun_world/dun_world.dmm
+++ b/_maps/map_files/dun_world/dun_world.dmm
@@ -9315,6 +9315,13 @@
 "cCB" = (
 /turf/closed/wall/mineral/rogue/stone/moss,
 /area/rogue/outdoors/river)
+"cCD" = (
+/obj/structure/fluff/railing/border{
+	pixel_x = 0;
+	pixel_y = 0
+	},
+/turf/open/transparent/openspace,
+/area/rogue/outdoors/mountains)
 "cCF" = (
 /obj/structure/bed/rogue/inn/wool,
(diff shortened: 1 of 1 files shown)


pinkledwrenis:
> i wouldn't say this remvoes soul or anything palisades are really bad and if anything this encourages wardens to instead rip down the palisades and make real walls with all the wood rather than spending it on useless barricades that fall in two seconds from being axe chopped by a heretic knight

Wood walls are subjectively inferior because of their climbing difficulty. With palisades, at least they have to stop for a few seconds to destroy it. With walls, they quite literally can just climb over, unless you build palisades on top of them as well.

There is no right answer, but I have to admit, any PR that reduces tedium in the long run is good. (Looking at soilson)

PR #1945, opened 2026-01-04
crimson cursed to not exist

chemistrymain3:
unfortunately relentless abuse of a fun virtue by a select few individuals such as a squire named leia who is the select individual but also everyone else who's ever run this has turned what was meant to be a fun and cute feature into a genuine detriment to the server

as the biggest detriment on the server, i'm jealous so i'm removing it

### DIFF
--- a/modular_azurepeak/virtues/combat.dm
+++ b/modular_azurepeak/virtues/combat.dm
@@ -204,7 +204,7 @@
 		else
 			recipient.mob_biotypes |= MOB_UNDEAD //Undead biotype is already applied by damned vice.
 
-/datum/virtue/combat/crimson_curse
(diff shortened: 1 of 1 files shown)


Plaguewalker:
eat shit

> unfortunately relentless abuse of a fun virtue by a select few individuals such as a squire named leia who is the select individual but also everyone else who's ever run this has turned what was meant to be a fun and cute feature into a genuine detriment to the server
> 
> as the biggest detriment on the server, i'm jealous so i'm removing it

How about ya just remove it instead of commenting it out then?

RandomGooberCoder:
In any case - this PR should not be merged. 
We do not comment out code.
We remove it completely.

PR #1953, opened 2026-01-05
Disables Ancillae gamemode. The one that spawns ~4 strong lickers instead of one VL.

RandomGooberCoder:
## About The Pull Request

Originally, the masquerade-like gamemode was intended to spawn powerful vampires in town.
You know, to give players some vampire-on-vampire action, instead of a one-sided vampire turbowar.
This was to be ensured by forcing them into different clans, giving them explicit objectives to work against each other while maintaining the said masquerade.

Alas! The gamemode is FUBAR in its unfinished state.

Neither that nor proper masquerade code was finished. And that resulted in...
- Players getting ancillae role as adventurers (Can there be masquerade if you are an irrelevant role unknown to the big world?
- Players not understanding that they won an antag roll that is half strong as a true VL.
- Rounds being **lame** due to this.
- Until the mode is unfucked - I propose we axe it outright, giving way for better modes.

## Testing Evidence

image

## Why It's Good For The Game

Justification for removal is the #About The Pull Request paragraph

Maybe Im too blackpilled but the goal was more about having fun shit like
"Normal round, but the whore, the blacksmith and the merchant are powerful vampire lords"
Incentivizing the inquisition to start looking  more into in-town rumors abt vampires, because you never know whether it is a harmless CC one or a mass-convert machine of doom
Or just outright inter-town conflicts instead of ye olde woode fraggery
I haven't witnessed any fun with ancillae. 
They either just.

Fuck off and do their own thing, which was not my goal.
Or they play as a discount VL. Which was not my goal.

### DIFF
--- a/roguetown.dme
+++ b/roguetown.dme
@@ -1482,7 +1482,6 @@
 #include "code\modules\events\antagonist\solo\_base.dm"
 #include "code\modules\events\antagonist\solo\bandits.dm"
 #include "code\modules\events\antagonist\solo\lich.dm"
-#include "code\modules\events\antagonist\solo\masquerade.dm"
 #include "code\modules\events\antagonist\solo\rebels.dm"
 #include "code\modules\events\antagonist\solo\thievesguild.dm"
 #include "code\modules\events\antagonist\solo\vampires.dm"


DatFrostys:
I understand the reasoning behind this PR but we are already lacking antagonists. I won't give suggestions or anything because I'm pretty sure you dislike when ppl do that, but aren't there other solutions?

PR #1959, opened 2026-01-05
Honsakiin: Saddleborn virtue & riding-based class changes

Ephemeralis:
## About The Pull Request

Introducing: **Saddleborn**.

- The Saddleborn virtue grants Apprentice riding and an ability that lets you choose from a selection of pre-saddled mounts while outdoors. However, **if your Saddleborn mount ever dies, you receive a hefty, completely unremovable debuff for the rest of the round (-10 stress).**
   - You can also set the name of this mount when spawning it.
   - Noble characters are able to pick from a selection of practically every horse that exists.
   - By popular demand, this also includes the riding hogs. Ogre mains rejoice, I guess (I still hate you).
- Saddleborn mounts can be **sent away** (teleported into nullspace) with an ability you receive after spawning it in. This only works outdoors.
   - If you send them away while in town, the bandit camp, the zurch, or the VL manor, your mount will receive a slow 5 minute healing effect that will persist while stored away and even when you call them back next time.
   - While sent away, saddleborn mounts are essentially in stasis and completely inaccessible to everybody else on the map, including you. They functionally don't exist as far as the game world is concerned. This is gamey, but intentional.
- Saddleborn mounts can be **whistled back** to you with a second ability that you receive after spawning in. This only works outdoors, and your mount must also be outdoors.
  - This will call your mount from anywhere on the map, as long as both you and it are outdoors.
     - This is how you steal a Saddleborn's mount for hostage purposes: put it in a house, a cave, or a stable somewhere.
  - This starts a fairly long channel (3-30 seconds depending) that gets longer the more dangerous the area you are in is.
     - Calling your mount in dangerous areas risks triggering an ambush as it returns to you. Being especially lucky might help mitigate this risk. Anywhere that counts as a warden area is dangerous, as is the Underdark, and Mount Decapitation.
     - Conversely, calling your mount in towns is quicker.
        - Being a noble and calling your mount from the keep is the quickest of all.
   - This makes a pretty loud whistling noise upon channel START that can be heard from almost two screens away.
 
More changes:
- Any class that previously had a static mount/horse awarded to it at spawn now has been given the Saddleborn spawning ability instead, as described above.

Bugfixes that accompany this:
- Healing miracles now properly reduce simple_mob bleeding.
- Golems and constructs no longer benefit from the wound heal associated with healing miracles, as this was applied when the status_effect was added to the character instead of on the tick that actually handles healing.

## Testing Evidence

image

image

image

## Why It's Good For The Game

Most people's experience with riding is looking at the sea of abandoned saigas at spawns and going "okay wow nobody bothers with this shit it must suck ass". There's a few reasons for this, but generally being able to safely store your mounts away and retrieve them was one of the big ones.

While this is kind of a gamified solution, it emerges from the fact that Scarlet Reach is a 100+ pop server at practically all times, and if even 10 people in the round need mount storage, there just categorically is not enough mapped stable space for all of them. This introduces a solution to that problem while also giving mount-centric builds a palpable tool that they can actually somewhat "tuck away" for use when they need it, instead of forgetting their roundstart shit at spawn because it's too much damn effort to lug the thing around.

It also provides some incentive for people to care about their mounts, and potential gameplay in people stealing people's prized horses. A 10 stress hit is pretty significant - the crankbox is 20, for example.

### DIFF
--- a/code/datums/status_effects/rogue/roguebuff.dm
+++ b/code/datums/status_effects/rogue/roguebuff.dm
@@ -539,7 +539,7 @@
 	if(wCount.len > 0)
 		owner.heal_wounds(healing_on_tick)
 		owner.update_damage_overlays()
-	if (HAS_TRAIT(owner, TRAIT_SIMPLE_WOUNDS))
+	if(HAS_TRAIT(owner, TRAIT_SIMPLE_WOUNDS))
 		owner.simple_bleeding = max(0, owner.simple_bleeding-(healing_on_tick/2))
 	owner.adjustBruteLoss(-healing_on_tick, 0)
 	owner.adjustFireLoss(-healing_on_tick, 0)
--- a/modular_scarletreach/virtues/saddleborn.dm
+++ b/modular_scarletreach/virtues/saddleborn.dm
@@ -12,8 +12,7 @@
 
 /mob/living/simple_animal/hostile/retaliate/rogue/goatmale/tame/saddled/Initialize()
 	. = ..()
-	var/obj/item/natural/saddle/S = new(src)
-	ssaddle = S
+	ssaddle = new /obj/item/natural/saddle(src)
 	update_icon()
 
 /mob/living/simple_animal/hostile/retaliate/rogue/goat/tame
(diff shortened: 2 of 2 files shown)


ChickyRooster:
Feedback: Make sending the mount away faster, as the mount will often just move out of range and enable the cooldown which is coal

BoblinTheGoblin2:
Great idea, however, whilst taking saddleborn as a class that gets the summon ability-you get two mounts.

Seashine:
Oh my god is lancer viable now?
Awesome PR though, I always love more character specialization stuff.

